PR #101, opened 2013-11-04
adding ability for any pre/post hooks to be run in same virtualenv as th...

dinopetrone:
adding ability for pre/post hooks to be run in same virtual env if the suffix of the hook file is .py


michaeljoseph:
Implemented by #152.

Issue #121, opened 2013-11-18
git clone over ssh not supported anymore


BrainGrylls:
So far I was using cookiecutter with a template that I store in a git repository, which I have to access over ssh. 
With version 0.6.4 of cookiecutter I was able to checkout that template without problem by using an scp-like syntax (e.g. cookiecutter user@host:path-to-repo.git).
With the changes made in version 0.7.0 to detect if it’s a repo URL or a local file path, ssh repositories are no longer supported.

In order to solve the issue I see two possible solutions:
1)	Instead of verifying if the given path follows a certain URL pattern, one could check if the given path exists on the file system (e.g. with os.path.exists(input_dir) )

2)	The if-clause could be extended to check if the path starts with “ssh://”. The users would have to adapt their call to: cookiecutter ssh://user@host/path-to-repo.git

pydanny:
I believe this was fixed several releases ago. Reopen if that is not the case.

cosmo-ethz:
It seems that with cookiecutter `1.0.0` this is still a problem. 
According to #145 others also encounter the same issue

hackebrot:
Please see #490

michaeljoseph:
Fixed by #490.

PR #122, opened 2013-11-20
Fix for Issue #121: Extended url detection

BrainGrylls:
Extended the URL detection in order to support git repositories accessed over ssh. 
A call would the look something like this:
cookiecutter ssh://user@host/path-to-repo.git


michaeljoseph:
Superseded by #490.

Issue #127, opened 2013-11-27
Conditional files / directories


mgaitan:
This is a complement issue for #126. Sometimes I want to create a file only under certain condition. 

I propose that the function `generate_file()` ignores the creation of file if its filename is empty. So, for example, I could template a filename like this :

``` jinja
/{% if cookiecutter.some_var %}{{ cookiecutter.another }}.txt{% endif %}
```
It seems this logic actually works for directories, but not for files.

nichochar:
I would love this. I hadn't found this ticket so posted it on SO: http://stackoverflow.com/questions/35005098/can-i-make-a-file-optional-based-on-a-variables-value-in-cookiecutter-json/35020626#35020626

glenfant:
Got a use case where I need a `__main__.py` if there's a console_script entry point in `setup.py`. Otherwise (pure lib) I don't need such file.

pydanny:
This can be done in post-generate hooks as seen in https://github.com/pydanny/cookiecutter-django/blob/master/hooks/post_gen_project.py#L163-L165. This makes this issue more an education issue then a technical solution. This is good because making logic work in directory/file names on Windows is impossible (and we support Windows).

Therefore, I believe this is better done as a recipe described in the [advanced_usage.rst](https://github.com/audreyr/cookiecutter/blob/master/docs/advanced_usage.rst) document.

PR #144, opened 2014-01-27
implemented: https://github.com/audreyr/cookiecutter/issues/127

bogensberger:
generate_file() ignores the creation of file if its filename is empty.

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -88,6 +88,10 @@ def generate_file(project_dir, infile, context, env):
     # Render the path to the output file (not including the root project dir)
     outfile_tmpl = Template(infile)
     outfile = os.path.join(project_dir, outfile_tmpl.render(**context))
+    # If the filename is empty after rendering the filename template
+    # stop here and don't create the file
+    if os.path.isdir(outfile):
+        return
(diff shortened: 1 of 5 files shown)


jstasiak:
+1, this is really useful

myyk:
+1, This would be super useful.

thomaslee:
+1, jinja misinterprets code `gradlew` scripts in Gradle projects & breaks hard. Might be possible to work around the issue with hooks, but doing so seems pretty hacky.

hackebrot:
@thomaslee how about [copy-without-render](http://cookiecutter.readthedocs.org/en/latest/advanced_usage.html#copy-without-render)? :confused:

thomaslee:
@hackebrot my bad -- looks like exactly what I'm after!

pydanny:
This can be done in post-generate hooks as seen in https://github.com/pydanny/cookiecutter-django/blob/master/hooks/post_gen_project.py#L163-L165. This makes this issue more an education issue then a technical solution. This is good because making logic work in directory/file names on Windows is impossible (and we support Windows).

Therefore, I believe this is better done as a recipe described in the [advanced_usage.rst](https://github.com/audreyr/cookiecutter/blob/master/docs/advanced_usage.rst) document.

PR #160, opened 2014-03-07
explicitely name unfound hooks in debug logging

raphigaziano:
Hi.

Just thought it'd be nice to have a slightly more explicit log message when a hook script is not found.
This is pretty minor, but I don't know, it's bugged me for some reason and for once I could actually find some time to do a little something =).


hackebrot:
Thank you for the PR, @raphigaziano! 😃

PR #162, opened 2014-03-15
Fixed --no-input inconsistency when the repo exists

jairot:
When you try to use cookie cutter from a repo that has been already cloned with the no-input flag. An input is shown. 

I fixed it and made a test for the issue.


michaeljoseph:
Superceded by #285.

PR #182, opened 2014-06-07
Set jinja options from a config file in a template

ghost:
As mentioned in my comment in #135, I've worked on a patch that allows jinja environment options to be specified in a config file in a cookiecutter template.

I put these options in a separate config file, so that older versions of cookiecutter don't ask for this as an option. I think this is going to break a lot of old versions of cookiecutter anyway if the cookiecutter templates start using these options :( I guess this requires some thought...

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -8,6 +8,7 @@ cookiecutter.generate
 Functions for generating a project from a project template.
 """
 from __future__ import unicode_literals
+import inspect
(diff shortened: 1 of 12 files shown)


hackebrot:
Hey! 👋 

I'm closing this in favor of 6cc1831dcde961c8995e0ea4eb36cafd8ead89f9 of #848.

Thank you for your contribution!

PR #186, opened 2014-06-23
WIP export context

EnTeQuAk:
This is more a work-in-progress PR as I wanted to get some opinion about the actual implementation.

Usecase was the following: I needed to write a hook that required access to already set config values. So I'm now forwarding those via environment variables.

I'm not sure if this is the way to go. E.g CasperJS allows context-variables to be passed via command line.

Any concerns?

Unittests etc will be written once we have a concept :)


EnTeQuAk:
Closing in favor of https://github.com/audreyr/cookiecutter/pull/288

PR #258, opened 2014-08-06
Get the config file location from $COOKIECUTTER_CONFIG, if set

pfmoore:
Apart from being useful for users, this will allow us to test more easily, without having to save/restore the user's config as much as we currently do.

### DIFF
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -54,7 +54,8 @@ def get_user_config():
     """
 
     # TODO: test on windows...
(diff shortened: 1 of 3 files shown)


msabramo:
The core change looks great but somehow a bunch of changes of double quotes to single quotes stowed away here. They make the PR longer than it needs to be and are likely the reason it won't merge cleanly now. 

If you strip out the extra stuff, I'd be :+1

jhermann:
I could do this on top of my PR, it is a one line change and 1or 2 tests.

PR #269, opened 2014-08-21
Support a new YAML format for template settings

pfmoore:
This is the initial version of support for a YAML format settings file, to supersede ```cookiecutter.json```. Before committing, I would like to refactor this further, to make the "internal" format completely encapsulated in ```settings.py``` so that client code need have no knowledge of how the settings are stored. But the PR is sufficiently complete that it's ready for initial review.

At the moment, it's only the context data that is in the settings, so the refactoring won't actually impact much, but it establishes the principle that following changes should follow. The main change at this stage will be in the tests, which currently know far too much about how the settings/context is stored.

### DIFF
--- a/cookiecutter/exceptions.py
+++ b/cookiecutter/exceptions.py
@@ -51,3 +51,13 @@ class UnknownRepoType(CookiecutterException):
     """
     Raised if a repo's type cannot be determined.
     """
+
+class MissingTemplateSettingsException(CookiecutterException):
+    """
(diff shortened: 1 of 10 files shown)


msabramo:
:+1: for YAML config

Let me know if I can help move this along.

pydanny:
I'm sad to close this PR, because @pfmoore deserves so much respect. However, #743 moves us away from YAML. From there we can explore other formats and techniques for configuration.

PR #280, opened 2014-09-08
Ensure file handles are closed using with statement

svisser:
This change ensures the file handles are closed properly in setup.py.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -12,8 +12,11 @@ if sys.argv[-1] == 'publish':
     os.system('python setup.py sdist upload')
     sys.exit()
 
-readme = open('README.rst').read()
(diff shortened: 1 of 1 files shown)


msabramo:
I don't see the benefit of using `with` in `setup.py` since it is short-running and files get closed on exit. 

One line is nicer to read.

svisser:
I think standard practise also applies to `setup.py`, it shouldn't be treated differently just because the process is short-lived.

pydanny:
I think demonstrating good practices overrides the desire for just a few less lines of code.

FWIW, Audrey says the original reason why `setup.py` was written this way was because she thought Python 2.6 didn't have the `with` statement.

svisser:
Thank you!

PR #327, opened 2014-11-15
Absolute imports.

michaeljoseph:


msabramo:
Doesn't merge cleanly so probably should be closed since @pydanny was -1

PR #334, opened 2014-11-17
Hackebrot convert test cookiecutter local with input

hackebrot:
Convert `test_main.TestCookiecutterLocalWithInput` using `monkeypatch` of `py.test`.

According to the name of the original name and the doc string of `test_cookiecutter_input_extra_context` the `no_input` argument needs to be `False`, so I changed the former version.

Please let me know if this is correct.

### DIFF
--- a/tests/test_cookiecutter_local_with_input.py
+++ b/tests/test_cookiecutter_local_with_input.py
@@ -2,8 +2,8 @@
 # -*- coding: utf-8 -*-
 
 """
-test_cookiecutter_local_no_input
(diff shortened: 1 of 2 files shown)


pydanny:
@hackebrot I want to merge this in for the 1.0 release this weekend (#410) but the file has evolved a bit. Would you be up to rebasing and resubmitting?

(If not I completely understand)

hackebrot:
@pydanny @michaeljoseph :bell:

PR #335, opened 2014-11-17
Leave it to py.test discovery

### LINKED ISSUE #333: Tests are collected twice with tox
Please see https://ci.appveyor.com/project/audreyr/cookiecutter/build/1.0.270/job/r93sespqrpv7w216 for reference. May be important to fix before #314

On a side note: `make coverage` prints the report twice. Once correct, once with coverage 0%. At least on my setup. Didn't do that before latest changes on master.

michaeljoseph:
Fixes #333.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 envlist = py26, py27, py33, py34, pypy, flake8
 
 [testenv]
-commands = py.test --cov cookiecutter tests {posargs}
+commands = py.test --cov cookiecutter {posargs}
 deps = pytest
        pytest-cov
 
--- a/tox.ini
+++ b/tox.ini
@@ -2,7 +2,7 @@
 envlist = py26, py27, py33, py34, pypy, flake8
 
 [testenv]
-commands = py.test --cov cookiecutter {posargs}
+commands = py.test --cov cookiecutter {posargs:tests}
 deps = pytest
        pytest-cov
 


tox.ini
```diff
@@ -2,7 +2,7 @@
 envlist = py26, py27, py33, py34, pypy, flake8
 
 [testenv]
-commands = py.test --cov cookiecutter tests {posargs}
+commands = py.test --cov cookiecutter {posargs}
```
msabramo: I usually use `{posargs:tests}`. This way by default, it will only look in the `tests` directory but the user can override and specify specific tests and things won't be run twice.

msabramo:
One minor nit, but generally I'm :+1:

michaeljoseph:
@msabramo @pydanny :bell:

michaeljoseph:
Actually, given #347, I think we should convert tox to using either the Makefile or `python setup.py test`?
This lets us remove the test dependency duplication also.

hackebrot:
@michaeljoseph I agree. It also solves the problem of running different commands for the coverage:

### CI
`tox.ini` https://github.com/audreyr/cookiecutter/blob/master/tox.ini#L5:
* `py.test --cov cookiecutter tests {posargs}`

### Local
`Makefile` https://github.com/audreyr/cookiecutter/blob/master/Makefile#L37:
* `coverage run --source cookiecutter setup.py test`

PR #347, opened 2014-11-26
Hackebrot integrate pytest in setup

hackebrot:
I implemented a `PyTest` command class in `setup.py` similar to http://pytest.org/latest/goodpractises.html#integration-with-setuptools-test-commands

Currently `python setup.py test` still uses `unittest` which means some of the re-implementations are not run at all. By changing this the documentation changes to `CONTRIBUTING.rst` of #310 are correct again.

Additionally I removed `pytest.ini` to address #333. It stopped us from running only specific tests as for instance `py.test tests/test_generate_file.py`. More importantly the suite is not executed twice any longer and the coverage is not calculated when running tests only.

Your feedback is highly appreciated! :kissing_closed_eyes:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,14 +27,14 @@ lint:
 	flake8 cookiecutter tests
 
 test:
-	py.test
+	python setup.py test
 
 test-all:
 	tox
 
 
 coverage:
-	coverage run --source cookiecutter -m py.test
+	coverage run --source cookiecutter setup.py test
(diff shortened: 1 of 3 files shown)


msabramo:
Conflicts.

hackebrot:
Resolved :wink:

msabramo:
LGTM.

:+1:

Issue #355, opened 2014-12-22
cookiecutter.json Decoding Error: Include File Path in Exception Message


rsyring:
When the cookiecutter.json file is malformed, a traceback like the following is given:

```
rsyring@loftex:~/tmp$ cookiecutter ~/projects/cookiecutters/pyproject/ 
config_path is /home/rsyring/.cookiecutterrc
Traceback (most recent call last):
  File "/home/rsyring/.local/bin/cookiecutter", line 9, in <module>
    load_entry_point('cookiecutter==0.8.0', 'console_scripts', 'cookiecutter')()
  File "/home/rsyring/.local/lib/python2.7/site-packages/cookiecutter/main.py", line 171, in main
    cookiecutter(args.input_dir, args.checkout, args.no_input)
  File "/home/rsyring/.local/lib/python2.7/site-packages/cookiecutter/main.py", line 95, in cookiecutter
    extra_context=extra_context,
  File "/home/rsyring/.local/lib/python2.7/site-packages/cookiecutter/generate.py", line 43, in generate_context
    obj = json.load(file_handle, object_pairs_hook=OrderedDict)
  File "/usr/lib/python2.7/json/__init__.py", line 290, in load
    **kw)
  File "/usr/lib/python2.7/json/__init__.py", line 351, in loads
    return cls(encoding=encoding, **kw).decode(s)
  File "/usr/lib/python2.7/json/decoder.py", line 366, in decode
    obj, end = self.raw_decode(s, idx=_w(s, 0).end())
  File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
    obj, end = self.scan_once(s, idx)
ValueError: Expecting property name: line 4 column 1 (char 115)
```

It would be really nice if this Exception message was extended to include the path of the file in question.  For a new template developer, the added context would save some googling.

rsyring:
new PR to replace this one: #388

rsyring:
PR #388 replaces #358

hackebrot:
#388 :white_check_mark:

PR #358, opened 2014-12-29
issue #355 - add better exceptions when encountering invalid json

rsyring:
addresses issue #355

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -10,6 +10,7 @@ Functions for generating a project from a project template.
 from __future__ import unicode_literals
 import logging
 import os
+import os.path as osp
 import io
 import shutil
 
@@ -19,7 +20,7 @@ from jinja2.exceptions import TemplateSyntaxError
 from binaryornot.check import is_binary
 
 from .compat import json, OrderedDict
-from .exceptions import NonTemplatedInputDirException
+from .exceptions import NonTemplatedInputDirException, ContextDecodingException
 from .find import find_template
 from .utils import make_sure_path_exists, work_in
 from .hooks import run_hook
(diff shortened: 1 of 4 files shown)


cookiecutter/generate.py
```diff
@@ -10,6 +10,7 @@
 from __future__ import unicode_literals
 import logging
 import os
+import os.path as osp
```
pydanny: Is there any reason not to do `import os` or `from os import path`?

cookiecutter/generate.py
```diff
@@ -40,7 +41,16 @@ def generate_context(context_file='cookiecutter.json', default_context=None,
     context = {}
 
     file_handle = open(context_file)
-    obj = json.load(file_handle, object_pairs_hook=OrderedDict)
+    try:
+        obj = json.load(file_handle, object_pairs_hook=OrderedDict)
+    except ValueError, e:
```
pydanny: This will fail on Python 3. Just switch to `except ValueError as e` and it will work fine. :-)

cookiecutter/generate.py
rsyring: Sorry, just an idiom I use frequently.  I'm lazy and os.path.... feels to long.  Just using "path" can be probelematic because I've found people like to use path as a variable name and I've encountered a few weird bugs in the past due to that naming collision.

cookiecutter/generate.py
rsyring: understood, thanks

cookiecutter/generate.py
pydanny: Gotcha. I'm lazy too. :smile: 

That said, cookiecutter's pattern is to write out the entire statement and not use aliases. Especially for the `os` library, which is used in nearly every module of the project.

rsyring:
replaced by #388

Issue #249, opened 2014-07-31
New config file format


pfmoore:
In the course of discussions on a number of issues (see, for example, #248, #10, #30 and #227), the topic of extensions to the `cookiecutter.json` format has come up. Rather than have the discussion scattered across multiple threads, let's come up with a next-generation config format that is flexible enough to cater for all the various use cases.

Some thoughts:
- We need to continue parsing the current format as "version 1.0" for backward compatibility.
- Do we stick with JSON, or should we use YAML? (An advantage of using YAML is that the config version is detectable from the filename, an advantage of sticking with JSON is that we don't have 2 parsers in the code).
- Maybe include a config version, so future enhancements can be versioned more easily?

msabramo:
I just started using cookiecutter but I'll say that YAML config would be welcome. I don't care for JSON as a config file format, because it's too strict about having commas but not on the last element which makes it error-prone when editing.

GrahamDumpleton:
This is a brain dump exploring one suggestion on how to move forward with this. It may not coherent or even sensible. :-)

My suggestion is that in the first instance you not try and extend cookiecutter to add support for some alternate or extended file format, but to enhance cookiecutter to call a user supplied hook function to generate the dictionary of final key/value pairs. This hook function could be placed in a code file called ``hooks/prompt_for_config.py`` and could be called ``prompt_items_for_config()``. The function would be called in place of calling ``cookiecutter.prompt.prompt_items_for_config()`` from ``cookiecutter.prompt.prompt_for_config()``.

Adding this capability means that it would allow people to implement their own alternate systems for how the contents of the ``cookiecutter.json`` file is interpreted if they need a more complex mechanism. Alternatively, they could use ``cookiecutter.json`` just for setting meta options such as ``_copy_without_render``. All other key/value pairs could then be supplied dynamically by the code implementing the users ``prompt_items_for_config()``, or pulled in or generated by virtue of some new file format for describing what is to be prompted.

One issue though is that as cookiecutter stands now, all the code to implement any custom prompting mechanism would need to be as part of the cookiecutter template package and it wouldn't be easy to make use of third party packages in that process.

For example, one could write a custom prompting mechanism based around ``prompt_toolkit``, manually coding everything required, but ``prompt_toolkit`` may not be installed.

One option to get around the lack of the ``prompt_toolkit`` would be for the custom prompter itself to fail with an error and simply tell people it needs to be installed. This is not ideal and could confuse less knowledgeable users.

[...]

PR #359, opened 2014-12-30
Support yaml context files

dnephin:
Related to #249

I tried to use #269 but it's diverged quite a bit from master. I resolved the conflicts and started trying to get it working again, but it ended up being a lot more work than just starting from scratch.

This was the smallest change I found that allowed me to read yaml context files, and still maintain backwards compatibility with the json files.

Feedback welcome.


pydanny:
This is a really nice pull request, but #743 moves us away from YAML. From there we can explore other formats and techniques for configuration.

PR #376, opened 2015-01-27
added the possibility of directly executing the binary without installing

vaab:
Hum, this doesn't seem to be a great deal, but it allows for very simple and straightforward usage of an
uninstalled deployed git repos. Very handy for test or using your git as installed repos.

If this doesn't make unicorn cry, please consider adding these 2 lines.

After this, a simple `ln -sf cookiecutter/cli.py /usr/local/bin/cookiecutter` will work out of the box (if of course, 'cookiecutter' module is accessible in the PYTHONPATH)

Not a very big deal, but nice to have.

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -11,10 +11,18 @@ Main `cookiecutter` CLI.
 from __future__ import unicode_literals
 
 import os
+import os.path
 import sys
 import logging
 
 import click
+
+if __name__ == "__main__":
+    # resolve symlink if any
+    current_filename = os.path.realpath(__file__)
+    # insert parent path in sys.path
+    sys.path.insert(0, os.path.join(os.path.dirname(current_filename), ".."))
+
 
 from cookiecutter import __version__
 from cookiecutter.main import cookiecutter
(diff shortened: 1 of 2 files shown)


pydanny:
I like the idea. I wonder if by using `subprocess` a test could be written for the change. That would be very nice.

vaab:
This is a little more invasive, but it's to ensure that we load the python files that are actually in the same place that the ``cli.py`` that we launch. This basically removes the constraint of having to declare this ``cookiecutter`` directory parent in the PYTHONPATH.

I added a quick test.

vaab:
As side notes:
- cookiecutter source code could provide a symlink (or a bat for windows ?) to execute this from the root of the package.
- ``cli.py`` is not using relative import. Not sure why as the other modules are doing so.

pydanny:
I don't know why `cli.py` isn't using relative imports either. I think it's an oversight. :wink: 

As for the symlink idea, I'm not enough of a POSIX expert to know if that's a good idea or not. I'll ask others. The .bat for windows is a pretty good idea.

hackebrot:
See #379

vaab:
Rebased.

vaab:
Integration suite seems broken. It breaks on ``flake8`` not found or on ``git clone`` or ``hg clone`` (see #434)

hackebrot:
What's the status on this? 🔔

theodesp:
@vaab Can you review the conflicting files if this is to be merged?

ssbarnea:

PR #387, opened 2015-02-06
README.rst: Dedent code blocks

msabramo:
so they don't display with vertical bar in left margin.

See https://github.com/msabramo/cookiecutter/blob/patch-1/README.rst

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -50,35 +50,35 @@ Did someone say features?
 
 * Simple command line usage:
 
-    .. code-block:: bash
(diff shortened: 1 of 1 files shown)


msabramo:
Hmmm, can't decide whether that's better or worse. The way you had it, the blocks are indented a bit, which looks kind of nice.

If you like it the way it is better, feel free to close.

hackebrot:
Think we'll stick with the current formatting, thank you though @msabramo! :bow:

PR #396, opened 2015-03-02
Add cookiecutter-latex-article to list of LaTeX templates.

Kreger51:
Hi!

I would be pleased if you would add my template to the list.

I'm currently a student and use this template to write my lab reports/essays and basically anything that I don't do by hand. However, it could be used for anything that makes use of the ``article`` document class. 

The template can be found at: https://github.com/Kreger51/cookiecutter-latex-article. 

Appreciate the utility by the way!

Thanks, 

Selim

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -200,6 +200,7 @@ LaTeX/XeTeX
 ~~~~~~~~~~~
 
 * `pandoc-talk`_: A cookiecutter template for giving talks with pandoc and XeTeX.
+* `cookiecutter-latex-article`_: A LaTeX template geared towards academic use.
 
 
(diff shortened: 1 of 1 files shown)


msabramo:
Looks trivial to merge. :+1:

pydanny:
Merging in.

FWIW, we actually like to examine and try the cookiecutter templates before we add them to the list.

msabramo:
Ah, good to know that you try them. That means the decision to merge isn't quite as trivial as I might've imagined.

Kreger51:
Thanks! Cool to know you try them indeed.

Issue #405, opened 2015-03-08
Files created on Windows should not use CRLF


pombredanne:
When running cookiecutter on Windows the output is different from running it on posix.
In particular even though templates use LF line endings, the results is files with CRLF line endings.
CRLF is bad in general in my book, but here converting to CRLF is not so nice.
Since Jinja uses LF by default, I wonder if this is not because Git would be trying to convert things, though my Git settings on Windows are set to make no conversions at all

pombredanne:
@audreyr this is a failing test for a start. cloning is not the culprit. The issue happens somewhere during file generation

pombredanne:
The commits in https://github.com/audreyr/cookiecutter/pull/406 shoud fix it.

themanifold:
Is anyone looking at this ticket? I've come across this and would be interested in a solution.

pombredanne:
@themanifold please go for it!

themanifold:
I came across this issue because I use cookiecutter in Windows after using
dulwich (pure python git implementation) to clone a template. Dulwich
itself has some decent logic for line endings that it borrows from git. But
for this case, I wonder if we can't do something as simple as detecting
what the original line endings are and respecting them.

franciscomoma:
@pombredanne I've fixed it in [this PR](https://github.com/audreyr/cookiecutter/pull/1159) applying that @themanifold suggest. Is it solving your problems too?

insspb:
@pombredanne @themanifold @franciscomoma Fixed in cookiecuter 2.0 by #1407 
You are welcome to do test of this fix, via using current master branch.

PR #406, opened 2015-03-10
#405 : Failing test showing that line endings are converted to CRLF

pombredanne:
This test will fail on Windows for now until #405  is fixed

### DIFF
--- a/tests/vcs/test_clone.py
+++ b/tests/vcs/test_clone.py
@@ -217,3 +217,27 @@ def test_clone_unknown_subprocess_error(mocker, clone_dir):
             clone_to_dir=clone_dir,
             no_input=True
         )
+
+
+def test_git_clone_does_not_change_line_endings_on_win():
+    repo_dir = vcs.clone(
+        'https://github.com/audreyr/cookiecutter-pypackage.git'
+    )
+    readme = open(os.path.join(repo_dir, 'README.rst'), 'rb').readlines()
+    first_line = readme[0]
+    assert first_line.endswith('\n')
+    assert not first_line.endswith('\r\n')
+    if os.path.isdir('cookiecutter-pypackage'):
+        utils.rmtree('cookiecutter-pypackage')
+
+
+def test_hg_clone_does_not_change_line_endings_on_win():
+    repo_dir = vcs.clone(
+        'https://bitbucket.org/pokoli/cookiecutter-trytonmodule'
+    )
+    readme = open(os.path.join(repo_dir, 'README.rst'), 'rb').readlines()
+    first_line = readme[0]
+    assert first_line.endswith('\n')
+    assert not first_line.endswith('\r\n')
+    if os.path.isdir('cookiecutter-pypackage'):
+        utils.rmtree('cookiecutter-pypackage')
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -173,7 +173,8 @@ def generate_file(project_dir, infile, context, env):
 
         logger.debug('Writing contents to file {}'.format(outfile))
 
-        with io.open(outfile, 'w', encoding='utf-8') as fh:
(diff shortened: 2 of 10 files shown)


pombredanne:
and this fails alright as it should

pombreda:
@audreyr  do not apply this for now. It needs to be refined first.
The line endings cannot be converted blindly to LF.... some windows files must have CRLF endings such as cmd/bat scripts.... therefore the approach should be instead to preserve the line endings of a file instead. A tad more tricky

hackebrot:
Not too sure how to proceed on #405 and #406 tbh :worried:

pydanny:
🔔  What's the status on this PR?

pombredanne:
@pydanny no idea :) let me rebase this first!

pombredanne:
This is rebased and should pass the tests.... Yet the issue is still there: CRLF should not be translated to LF. That said, I have lost a lot (or most) of my patience for these kind of Windows minutiae.

lorengordon:
This is driving me nuts as well. I've converted all files in the cookiecutter template to LF, created a .gitattributes file forcing line endings to LF (both in the template repo and in the project repo), and still the project keeps getting created with CRLF line endings! Any bright ideas?

theodesp:
@pombredanne There are some conflicts that need to be fixed for that if is to be merged 👍

pombredanne:
@theodesp I rebassed... but the vcs test patterns have evolved quite bit since last year so it may not pass...  feel free to update as needed to make this mergeable.

pombredanne:
@theodesp tests arer failing  ... make it pass if you want this in! It is not like I thoroughly enjoy keeping patches alive for 2.5 ;) so it is yours now or anyone else ;)

pombredanne:
@theodesp basically the code context has changed too much... This would need to be redone from the start

pombredanne:
@pydanny @audreyr IMHO this PR should be closed at it is and someone can restart something using it as a base if needed.

insspb:
@ssbarnea Please do not close PRs silently without any resolution. 
This PR is related to #1159 and tests from here can be used there as second recheck.

tests/vcs/test_clone.py
```diff
@@ -217,3 +217,27 @@ def test_clone_unknown_subprocess_error(mocker, clone_dir):
...
+    )
+    readme = open(os.path.join(repo_dir, 'README.rst'), 'rb').readlines()
+    first_line = readme[0]
+    assert first_line.endswith('\n')
+    assert not first_line.endswith('\r\n')
+    if os.path.isdir('cookiecutter-pypackage'):
+        utils.rmtree('cookiecutter-pypackage')
+
+
+def test_hg_clone_does_not_change_line_endings_on_win():
```
insspb, line 234: This is critical and unacceptable. External executable should not be used in tests (hg in this case)

tests/vcs/test_clone.py
```diff
@@ -217,3 +217,27 @@ def test_clone_unknown_subprocess_error(mocker, clone_dir):
             clone_to_dir=clone_dir,
             no_input=True
         )
+
+
+def test_git_clone_does_not_change_line_endings_on_win():
+    repo_dir = vcs.clone(
```
insspb, line 223: This is critical and unacceptable. External executable should not be used in tests (git in this case)

insspb:
Restored as #1407

tests/vcs/test_clone.py
insspb: Fixed in #1407

tests/vcs/test_clone.py
insspb: Fixed in #1407

pombredanne: @insspb you are correct ... and way back then, I might have been lazy to use mocks ;)
Thank you  for picking up the ball and I am glad this was of some use after all.

PR #412, opened 2015-03-15
Convert test hooks to pytest

hackebrot:
Convert `test_hooks.py` to `py.test` syntax. 

I didn't change the underlying test structure to be able to reuse the helper function `make_test_repo` without drastic changes. So it's only the setup/teardown code and the assert statements that change. :sunglasses:


michaeljoseph:
:+1:

PR #420, opened 2015-03-18
pip is preferred for Python packaging over easy_install

streeter:
https://python-packaging-user-guide.readthedocs.org/en/latest/current.html


hackebrot:
:+1:

PR #424, opened 2015-03-21
Added --rc-file option to load specific or no user config file

jhermann:
Motivation (use-cases):
 * isolate unit tests of cookiecutter templates from user home
 * Selection of different config sets in a CI server or similar


jhermann:
See also #258

PR #426, opened 2015-03-24
Enriched context for hooks

jhermann:
You can find the use-case for this [here](https://github.com/Springerle/dh-virtualenv-mold/blob/master/hooks/pre_gen_project.py). It passes information to hooks (and any template) that is either hard to get or unavailable to them right now. If you agree this should be merged, I can add documentation and tests to this PR.

Sample output (with v1.0.0):

```
*** verbose=None
*** checkout=None
*** version=None
*** version_info=()
*** repo_dir=None
*** context_file=None
*** context={u'year': u'2015', …
…
```

Sample output (with patch):

```
*** verbose=False
*** checkout=None
*** version=1.0.0
*** version_info=(1, 0, 0)
*** repo_dir=/home/jhe/src/dh-virtualenv-mold
*** context_file=/home/jhe/src/dh-virtualenv-mold/cookiecutter.json
*** context={u'year': u'2015', …}
*** context[pprint]={u'distro': u'UNRELEASED',
…
u'year': u'2015'}
*** argv=['/tmp/tmpo5TZB5.py']
*** cwd=/home/jhe/src/gh-commander/debian
*** ls=['rules', 'changelog', 'pyvenv-foobar.links', 'control', 'copyright', 'pyvenv-foobar.triggers', 'compat', 'pyvenv-foobar.postinst']
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,4 @@ htmlcov/
 
 # Local env
 .envs
+.venv
--- a/cookiecutter/__init__.py
+++ b/cookiecutter/__init__.py
@@ -9,7 +9,7 @@ Main package for Cookiecutter.
 """
 from .compat import OLD_PY2
 
-__version__ = '1.0.0'
+__version__ = '1.0.1'
 
 if OLD_PY2:
     msg = 'Python 2.6 support was removed from cookiecutter in release 1.0.0.'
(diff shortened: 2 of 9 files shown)


jhermann:
Hmm, did not actually mean to merge the other PR into THIS branch. Anyway, if you ever come around to this, we can sort out the mess, or else you saved time merging this yourself.

PR #438, opened 2015-04-11
python-future support

### LINKED ISSUE #298: Address test coverage issues for compat.py
Currently `cookiecutter.compat` gets a pass on testing via `pragma: no cover`. Yes, we have tox on all the other stuff, but it ain't right. I'm opening this ticket because I believe through creative use of `mock.patch` we can have appropriate tests and remove those nasty  `pragma: no cover` statements.

This follows on #203 and #252.

On the other hand, if we decide to add another dependency and include `six`, this becomes moot.

### LINKED ISSUE #301: Consider moving to six or future for Python 2/3 compatibility
**Should we move to one of these libraries?** I think so, as it would mean we could nearly eliminate the `compat.py` module, improving test coverage.

**Which library?** `six`, `future`, or something else?

michaeljoseph:
Fixes #301
Fixes #298

### DIFF
--- a/cookiecutter/compat.py
+++ b/cookiecutter/compat.py
@@ -9,49 +9,24 @@ from past.builtins import basestring
 PY3 = sys.version_info[0] == 3
 OLD_PY2 = sys.version_info[:2] < (2, 7)
 
+
+def read_response(prompt=''):
+    """
+    Prompt the user for a response.
+
+    Prints the given prompt (which should be a Unicode string),
+    and returns the text entered by the user as a Unicode string.
+
+    :param prompt: A Unicode string that is presented to the user.
+    """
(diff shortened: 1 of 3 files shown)


hackebrot:
I'm voting for ``click`` to handle all the CLI prompting which is currently part of ``compat.py``.

Please see #443

pydanny:
As we are using `click` for CLI prompting, what should we do with this pull request?

hackebrot:
Guess we don't really need it any longer, am I right @michaeljoseph? :no_mouth:

hackebrot:
:bell:

michaeljoseph:
@hackebrot yeah, let's close, this will need to be reworked.

PR #444, opened 2015-04-21
Added fix and tests to issue 267

iljabauer:
I added a check if the name of the file which should be generated is empty. If so, we do not generate the file and log this behaviour.


audreyr:
Sorry for the delay @iljabauer. This is great. Merging...

PR #445, opened 2015-04-22
more cross-platform make commands

brad:
The `open` command doesn't exist on Linux and other variants. This is a simple way that should work in all UNIXy environments, making the Makefile less OS X-centric

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ coverage:
 	coverage run --source cookiecutter setup.py test
 	coverage report -m
 	coverage html
-	open htmlcov/index.html
+	python -c "import webbrowser as wb; wb.open('htmlcov/index.html')"
 
 docs:
 	rm -f docs/cookiecutter.rst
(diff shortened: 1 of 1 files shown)


brad:
I guess appveyor timed out, it worked fine when I reran it on my appveyor account.

hackebrot:
For some reason this doesn't work on Mac OS X. Not sure why...apparently other ppl ran into similar issues. :worried:

audreyr:
Really sorry for the delay @brad. Addressing this is a good idea, but this doesn't work for me either. 

We did something similar in cookiecutter-pypackage, where@eliasdorneles came up with a really creative workaround that would work fully cross-platform: https://github.com/audreyr/cookiecutter-pypackage/pull/58

hackebrot:
This issue seems to be resolved as per d5f3c044670f144cb55368087e22ef6a7104fe6b.

Let me know if that doesn't work for you! 🙇

PR #446, opened 2015-04-26
tests: ensure to use an existing command to test `which`

vincentbernat:
When testing cookiecutter while not having it installed, we get this
failure:

    =================================== FAILURES ===================================
    ____________________________ test_existing_command _____________________________

        def test_existing_command():
    >       assert which('cookiecutter')
    E       assert None
    E        +  where None = which('cookiecutter')

    tests/test_compat.py:15: AssertionError
    =============== 1 failed, 74 passed, 16 skipped in 0.43 seconds ================

A command that is almost guaranteed to exist is `date` (both on Unix and
Windows systems).

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -12,7 +12,7 @@ from cookiecutter.compat import which
 
 
 def test_existing_command():
-    assert which('cookiecutter')
+    assert which('date')
 
 
 def test_non_existing_command():


hackebrot:
Hi @vincentbernat,

I am afraid I don't quite get your point. :worried: 

Given that you set up **cookiecutter** for development you should have it installed and not run into issues with the named test.

pydanny:
I think @vincentbernat is correct. This particular cookiecutter test is designed to test the `which()` function, not if cookiecutter is installed. ;)

On a side note, according to @audreyr, @vincentbernat might be building a Debian package for Cookiecutter. :dancers:

vincentbernat:
Yes, that's mostly for the Debian packaging.

I was going to say that I could also run cookiecutter without installing it with `python -m cookiecutter` or `python -m cookiecutter.cli` but the support is not here. Interested in a PR about that? I could do that with `python -c 'import cookiecutter.cli ; cookiecutter.cli.main()'` but nobody would believe it. ;-)

hackebrot:
Ok, I see.

Guess I'm :+1: then.

PR #447, opened 2015-04-26
Add more checks to test_compat.py

hackebrot:
This PR changes ``test_compat.py`` to be more specific about the expected result of ``which``.


hackebrot:
:bell: @pydanny @vincentbernat

PR #449, opened 2015-04-27
cli: allow `python -m cookieclutter.cli` invocation

vincentbernat:
Without even installing cookiecutter, a user can run it with `python -m
cookieclutter.cli`.

Another option would be to allow `python -m cookieclutter` directly.

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -73,3 +73,6 @@ def main(template, no_input, checkout, verbose):
     except OutputDirExistsException as e:
         click.echo(e)
         sys.exit(1)
+
+if __name__ == "__main__":  # pragma: no cover
+    main()


hackebrot:
Hi @vincentbernat, thank you for the contribution. Can you please amend the commit message to use `cookiecutter` instead of `cookieclutter`. Other than that we need to have a test for this, I think.

vincentbernat:
I have updated the commit message and rebased the patch. As for tests, I really don't see how to easily test that into an unittest.

vincentbernat:
I have also pushed a "pragma: no cover" to not decrease code coverage.

pydanny:
In theory one could use the `subprocess` module to test these lines of code - calling the `cli` from the command-line. However, I think what's there is good enough. :+1:

hackebrot:
I am afraid `main` cannot be called without arguments. It requires you to at least specify `template` :exclamation: 

Without installing `Cookiecutter` chances are you are missing some libraries such as `click`, `yaml` etc. :worried:

vincentbernat:
`main` is decorated by click stuff, so it works just fine. Try it yourself:

```
PYTHONPATH=$PWD python -m cookiecutter.cli --help
```

hackebrot:
My bad, I just didn't feed a template to it. Sorry :sleeping:

PR #443, opened 2015-04-18
Choices in Context

hackebrot:
Add support for choices in a context as suggested in #441.

I noticed that we are using custom methods for retrieving the user input and decided to let ``click.prompt`` handle the Python2/Python3 compatibility matters. This PR requires thorough reviews and testing before merging it as I touched some core functions and I want to be certain I didn't break anything :exclamation: 

The CI might not detect these errors as most of the ``read_response`` calls are still ``monkeypatched``. I'm afraid we need to do some manual testing to be 100% positive that everyting is working as expected :sweat: 

I implemented several tests for the new functionality. :smiley: 

Please let me know your thoughts and I'll be happy to update this PR.

Issue #441, opened 2015-04-15
Prompt list of choices if list in cookiecutter.json


Pawamoy:
Example:

    {
        "status_classifier": [
            "Development Status :: 1 - Planning",
            "Development Status :: 2 - Pre-Alpha",
            "Development Status :: 3 - Alpha",
            "Development Status :: 4 - Beta",
            "Development Status :: 5 - Production/Stable",
            "Development Status :: 6 - Mature",
            "Development Status :: 7 - Inactive"
        ],
    }

... would be prompted like

    Select status_classifier:
        1 - Development Status :: 1 - Planning
        2 - Development Status :: 2 - Pre-Alpha
        3 - Development Status :: 3 - Alpha
        4 - Development Status :: 4 - Beta
        5 - Development Status :: 5 - Production/Stable
        6 - Development Status :: 6 - Mature
        7 - Development Status :: 7 - Inactive
    Enter your choice (1, 2, 3, 4, 5, 6 or 7):

pydanny:
I like this idea! Let me bounce it off the other core devs. @audreyr @michaeljoseph @hackebrot

hackebrot:
Great idea, indeed! :+1:  Need this for my Kivy templates :grinning: 

Did a quick proof-of-concept using ``click.prompt`` with [choices](http://click.pocoo.org/4/options/#choice-options) see https://github.com/hackebrot/cookiecutter/commit/3798f0945399edcb115cb09136dbbc73bcf4aa52 :triumph:

michaeljoseph:
:sparkling_heart:

Pawamoy:
Very nice, thank you guys :yum:

luzfcb:
:sparkling_heart: :+1:

audreyr:
:+1:

luzfcb:
@hackebrot in its proof of concept, how would defined the default value of choices in cookiecutter.json?

ionelmc:
Will it support multiple selection?

hackebrot:
@luzfcb The first item in the list of choices will be used as the default
@ionelmc I'm afraid it won't for now.

Please review #455 :bow:

cjw296:
ionelmc - see https://github.com/audreyr/cookiecutter/issues/1002 for multiple choices.

PR #451, opened 2015-05-05
Use click for prompting in read_response

hackebrot:
This PR removes ``read_response`` from ``compat.py`` in favor of a re-implementation using [click.prompt](http://click.pocoo.org/4/api/#click.prompt).

I wrote a simple test to make sure that ``click.prompt`` is being called with the correct arguments so that it mimics the current prompting scheme by leaving the actual prompt generation to ``prompt_for_config``.

I will submit another request to change ``query_yes_no`` and to introduce the choices feature of #441 (and close #443). 

Please let me know your thoughts. :smile:

### DIFF
--- a/cookiecutter/compat.py
+++ b/cookiecutter/compat.py
@@ -1,8 +1,14 @@
+import click
 import os
 import sys
 
 PY3 = sys.version_info[0] == 3
 OLD_PY2 = sys.version_info[:2] < (2, 7)
+
+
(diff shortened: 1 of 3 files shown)


pydanny:
I love seeing the commits where `compat.py` get smaller and smaller. :)

hackebrot:
Hooray :smile_cat:

PR #453, opened 2015-05-19
Fix encoding for latest tox

hackebrot:
This PR solves a problem with our setup of ``tox`` and ``click`` which is explained in great detail over https://gist.github.com/hackebrot/937245251887197ef542

I implement a simple test for PY3 to make sure we keep the new setting in ``tox.ini`` that comes with this PR. If anyone ever experiences a similar problem with our setup, we should have a good starting point to investigate, I think.

Tests are passing now on Mac OS X as well as Ubuntu Trusty Tahr.

Many thanks to @audreyr @pydanny @goodwillcoding and @ionelmc who eventually pointed me into the right direction. :bow:

### DIFF
--- a/tests/test_preferred_encoding.py
+++ b/tests/test_preferred_encoding.py
@@ -6,6 +6,7 @@ import pytest
 
 from cookiecutter.compat import PY3
 
+
 @pytest.mark.skipif(not PY3, reason='Only necessary on Python3')
 def test_not_ascii():
     """Make sure that the systems preferred encoding is not `ascii`.
--- a/tests/test_preferred_encoding.py
+++ b/tests/test_preferred_encoding.py
@@ -1,0 +1,22 @@
+# -*- coding: utf-8 -*-
+
+import locale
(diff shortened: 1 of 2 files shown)


ionelmc:
@hackebrot can't hurt to make a tox bug report (as this is a tox 2.0 "regression") - surely this affects other projects

pfmoore:
I think it's a deliberate change in tox 2.0. See http://tox.readthedocs.org/en/latest/changelog.html#id2 ("Environment variable isolation")

hackebrot:
Thank you @ionelmc @pfmoore! Guess I'll ping Holger and Armin and leave it up to them to decide if this requires some more clarification. Even though it is not a bug some ppl may be confused if they use both **tox** and **click** on Python3 and run into this.

To be perfectly honest I missed that change in the latest tox version although I am subscribed to the pytest-dev mailinglist. :no_mouth:

Glad we got this sorted :smile_cat:

Issue #155, opened 2014-02-26
Inject current datetime into Jinja environment


khorn:
Add the current datetime to the Jinja2 environment so that it can be rendered in templates.

e.g.
{{ cookiecutter.current_time }}

Might also want config vars for specifying local vs. UTC, etc.

See also #154

michaeljoseph:
:clock10: 
[1.4.0](https://github.com/audreyr/cookiecutter/releases/tag/1.4.0) now includes [hackebrot/jinja2-time](https://github.com/hackebrot/jinja2-time).

Issue #154, opened 2014-02-26
Include a Jinja filter to format datetime objects.


khorn:
Because there's no good way to format dates at the moment...

michaeljoseph:
:clock10: 
[1.4.0](https://github.com/audreyr/cookiecutter/releases/tag/1.4.0) now includes [hackebrot/jinja2-time](https://github.com/hackebrot/jinja2-time).

fcarvalhopacheco:
Please see #1663

PR #463, opened 2015-06-25
Provide `cookiecutter._now` for config templates

blueyed:
This allows to use the following (in the cookiecutter-django template):

    -    "now": "2015/01/13",
    -    "year": "{{ cookiecutter.now[:4] }}",
    +    "year": "{{ cookiecutter._now.strftime('%Y-%m-%d') }}",

Using a datetime object (instead of a formatted string) makes it more
flexible for different use cases.

This addresses #154 and #155 (which I've found after doing this patch).

Using the approach from #155 might be better (providing `strftime`).
It is mentioned in the docs: https://cookiecutter.readthedocs.org/en/latest/advanced_usage.html#example-injecting-a-timestamp

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -10,6 +10,7 @@ Functions for prompting the user for project info.
 
 from __future__ import unicode_literals
 from collections import OrderedDict
+import datetime
 
 import click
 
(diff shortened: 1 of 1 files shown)


hackebrot:
Thank you @blueyed! :bow: 

I'm afraid we need to find a more sophisticated approach of dealing with *builtin* variables.

Please feel free to join our [gitter](https://gitter.im/audreyr/cookiecutter) for a related discussion.

fcarvalhopacheco:
please see #1663

PR #469, opened 2015-07-08
Add new link to README

christabor:
Adding my link here, if that's acceptable, as I think this project aligns with a few others (notably django-cookiecutter), but does the rendering and CLI stuff directly, similar to yours, with a few exceptions (notably inflection, and model type inferences).

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -300,6 +300,8 @@ Similar projects
 
 * `Cog`_ python-based code generation toolkit developed by Ned Batchelder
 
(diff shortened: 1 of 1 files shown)


pydanny:
For new items on this particular list, we insist that they link back to us in a 'Similar projects' section. Skaffold does not have that section, hence we can't link to it until it does.

christabor:
@pydanny, I added a link to both CookieCutter and Django-CookieCutter, as they're both in the same vein. Thanks for the heads up!

Issue #470, opened 2015-07-08
Ask before writing into an existing project directory


lgp171188:
I have an existing Django project that does not use the cookiecutter-django template. When running ``cookiecutter https://github.com/pydanny/cookiecutter-django``, it writes into the project directory even if it already exists without warning the user about it.

The correct behaviour would be to abort if the project directory already exists or at least warn the user and ask for confirmation before writing into the existing directory.

I have raised an issue in the cookiecutter-django project for this - https://github.com/pydanny/cookiecutter-django/issues/265

After having a quick look at the source for cookiecutter-django, it looks like this could be an issue with cookiecutter.

pydanny:
:+1: to a patch.

hackebrot:
Resolved and merged! :ok_hand:

PR #471, opened 2015-07-09
[Issue 470] Error out if project output directory exists.

lgp171188:
Fix issue #470 by erroring out if the project output directory
exists. This will prevent cookiecutter from writing into the existing directories like it does now.

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -21,7 +21,11 @@ from jinja2.environment import Environment
 from jinja2.exceptions import TemplateSyntaxError
 from binaryornot.check import is_binary
 
-from .exceptions import NonTemplatedInputDirException, ContextDecodingException
+from .exceptions import (
+    NonTemplatedInputDirException,
+    ContextDecodingException,
+    OutputDirExistsException
+)
 from .find import find_template
 from .utils import make_sure_path_exists, work_in
 from .hooks import run_hook
@@ -148,7 +152,7 @@ def generate_file(project_dir, infile, context, env):
     shutil.copymode(infile, outfile)
 
 
-def render_and_create_dir(dirname, context, output_dir):
+def render_and_create_dir(dirname, context, output_dir, fail_if_exists=False):
     """
     Renders the name of a directory, creates the directory, and
     returns its path.
@@ -163,6 +167,12 @@ def render_and_create_dir(dirname, context, output_dir):
     dir_to_create = os.path.normpath(
         os.path.join(output_dir, rendered_dirname)
     )
+
+    if fail_if_exists:
+        if os.path.exists(dir_to_create):
+            msg = 'Error: "{}" directory already exists'.format(dir_to_create)
+            raise OutputDirExistsException(msg)
+
     make_sure_path_exists(dir_to_create)
     return dir_to_create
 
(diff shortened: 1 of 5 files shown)


tests/test_cli.py
```diff
@@ -1,24 +1,55 @@
...
+    Remove the fake project directory created during the tests.
+    """
+    def fin_remove_fake_project_dir():
+        if os.path.isdir('fake-project'):
+            utils.rmtree('fake-project')
+    request.addfinalizer(fin_remove_fake_project_dir)
+
+
+@pytest.fixture(scope='function')
+def make_fake_project_dir(request):
```
hackebrot: I suggest to have both setup and teardown in a single fixture.

cookiecutter/generate.py
```diff
@@ -163,6 +167,11 @@ def render_and_create_dir(dirname, context, output_dir):
     dir_to_create = os.path.normpath(
         os.path.join(output_dir, rendered_dirname)
     )
+    if os.path.exists(dir_to_create):
+        if fail_if_exists:
```
hackebrot: Let's swap these if clauses. We do not need to check for existence of the dir for `fail_if_exists == False`.

cookiecutter/generate.py
```diff
@@ -163,6 +167,11 @@ def render_and_create_dir(dirname, context, output_dir):
...
+            msg = 'Error: "{0}" directory already exists'.format(dir_to_create)
```
hackebrot: Feel free to omit the index in the in error message. We dropped Py26 support a while ago. :smile_cat:

lgp171188:
@hackebrot Made the suggested changes to ``cookiecutter/generate.py``. Please review and let me know if everything is fine.

lgp171188:
@hackebrot I have omitted the ``scope`` argument for the ``fixture`` decorator since since ``function`` is the default

tests/test_cli.py
```diff
@@ -1,24 +1,55 @@
...
     result = runner.invoke(main, ['-V'])
     assert result.exit_code == 0
     assert result.output.startswith('Cookiecutter')
 
 
+@pytest.mark.usefixtures('make_fake_project_dir', 'remove_fake_project_dir')
+def test_cli_error_on_existing_output_directory():
+    result = runner.invoke(main, ['tests/fake-repo-pre/', '--no-input'])
+    assert result.exit_code != 0
+    assert result.output.startswith('Error:')
```
hackebrot: What's your opinion on verifying the full error message?

```python
    expected_err_msg = 'Error: "fake-project" directory already exists\n'
    assert result.output == expected_err_msg
```

lgp171188: Yes that makes sense since there is a chance that the error message could be for some other error and the test does not verify it at the moment. Will fix this now.

lgp171188:
@hackebrot Do let me know if there are more issues with the changes in this pull request that I have to fix.

hackebrot:
Think you are all set. :+1: 

@pydanny I'll accept my very first Cookiecutter PR, if you agree on these changes :smile_cat:

pydanny:
:+1: @hackebrot. Don't forget to update `history.rst` and `AUTHORS.rst`.

hackebrot:
Resolved #470

PR #480, opened 2015-07-29
Remove test command in favor of tox

### LINKED ISSUE #466: Instructions for local development are incomplete.
When executing the line:

    python setup.py test

as directed, I get the following error:

    running test
      File "setup.py", line 111, in <module>
        tests_require=test_requirements
      File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
        dist.run_commands()
      File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
        self.run_command(cmd)
      File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "setup.py", line 60, in run
        import pytest
    ImportError: No module named pytest

I'm not sure what you are supposed to do to install the 'tests_require' dependencies.

hackebrot:
This PR introduces a better way of running the test suite. Instead of setting up a `TestCommand` in `setup.py` it leaves the dependency management for testing to `tox`. As a consequence I also removed `mock` from `compat.py` and `setup.py`. From what I know, it was solely used in tests (and `grep` did not report anything else).

The steps to run the test suite are as follows:

```bash
$ git clone https://github.com/audreyr/cookiecutter.git
$ cd cookiecutter/
$ mkvirtualenv -a $(pwd) cookiecutter
$ python setup.py develop

$ pip install tox
$ tox
```

I updated the contributor docs to reflect this behavior, which fixes #466.

Please let me know your thoughts!

Best
Raphael

### DIFF
--- a/docs/contributor_setup.rst
+++ b/docs/contributor_setup.rst
@@ -22,9 +22,14 @@ Now you can make your changes locally.
 
 5. When you're done making changes, check that your changes pass the tests and flake8::
 
-    $ flake8 cookiecutter tests
     $ pip install tox
     $ tox
+
+Please note that tox runs flake8 automatically, since we have a test environment for it.
+
+If you feel like running only the flake8 environment, please use the following command::
+
+    $ tox -e flake8
 
 6. Commit your changes and push your branch to GitHub::
 
--- a/docs/contributor_setup.rst
+++ b/docs/contributor_setup.rst
@@ -23,7 +23,7 @@ Now you can make your changes locally.
 5. When you're done making changes, check that your changes pass the tests and flake8::
 
     $ flake8 cookiecutter tests
-    $ python setup.py test
+    $ pip install tox
     $ tox
 
 6. Commit your changes and push your branch to GitHub::
@@ -34,8 +34,6 @@ Now you can make your changes locally.
 
 7. Check that the test coverage hasn't dropped::
 
-    coverage run --source cookiecutter setup.py test
-    coverage report -m
-    coverage html
+    $ tox -e cov-report
 
 8. Submit a pull request through the GitHub website.
--- a/tox.ini
+++ b/tox.ini
@@ -3,7 +3,7 @@ envlist = py27, py33, py34, pypy, flake8
 
 [testenv]
 passenv = LC_ALL, LANG
-commands = py.test --cov cookiecutter {posargs:tests}
+commands = py.test --cov=cookiecutter {posargs:tests}
 deps = pytest
        pytest-cov
        pytest-mock
--- a/tox.ini
+++ b/tox.ini
@@ -9,14 +9,12 @@ deps = pytest
        pytest-mock
 
 [testenv:py27]
-deps =
-    {[testenv]deps}
-    mock
+deps = {[testenv]deps}
+       mock
 
 [testenv:pypy]
-deps =
-    {[testenv]deps}
-    mock
+deps = {[testenv]deps}
+       mock
 
 [testenv:flake8]
 deps =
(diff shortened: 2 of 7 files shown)


hackebrot:
@michaeljoseph Do you happen to have time for a review? :grinning:

tox.ini
```diff
@@ -9,14 +9,12 @@ deps = pytest
        pytest-mock
 
 [testenv:py27]
-deps =
-    {[testenv]deps}
-    mock
+deps = {[testenv]deps}
+       mock
```
michaeljoseph: Do we still need this `mock` dependency?

tox.ini
```diff
@@ -9,14 +9,12 @@ deps = pytest
...
 
 [testenv:pypy]
-deps =
-    {[testenv]deps}
-    mock
+deps = {[testenv]deps}
+       mock
```
pydanny: I agree with @michaeljoseph . Is this mock needed here as well?

Otherwise LGTM

tox.ini
```diff
@@ -24,3 +22,6 @@ deps =
     pep8==1.6.2
 commands =
     flake8 cookiecutter
+
+[testenv:cov-report]
+commands = py.test --cov=cookiecutter --cov-report=term --cov-report=html
```
michaeljoseph: Seeing the way you're using `--cov` here, makes me think that https://github.com/hackebrot/cookiecutter/blob/remove-test-command-in-favor-of-tox/tox.ini#L6 is incorrect?

docs/contributor_setup.rst
```diff
@@ -23,7 +23,7 @@ Now you can make your changes locally.
 5. When you're done making changes, check that your changes pass the tests and flake8::
 
     $ flake8 cookiecutter tests
```
michaeljoseph: We should change this to use tox too, since the `flake8` dependency won't be installed by `python setup.py develop`

hackebrot:
Thank you @michaeljoseph and @pydanny for your thorough review :bow: 

From what I know codecov reports the coverage for each of the test envs. If I get rid of `--cov` in the generic `testenv` it will stop doing so, I'm afraid. That's why I left this unchanged.

tox.ini
hackebrot: I removed `mock` and everything seems to work just fine in a fresh env.

michaeljoseph:
@hackebrot I meant, the generic `testenv` should have `--cov=cookiecutter` instead of `--cov cookiecutter`?

hackebrot:
I changed the test env command to also reflect the invocation from the official docs. https://pypi.python.org/pypi/pytest-cov

The docs now explain how to run `flake8` via `tox`. However it only checks `cookiecutter` and ignores `tests`. We can exclude the *dummy* test data directories from flake8 though (they cause errors atm).

tox.ini
```diff
@@ -3,20 +3,16 @@ envlist = py27, py33, py34, pypy, flake8
...
+commands = py.test --cov=cookiecutter {posargs:tests}
 deps = pytest
        pytest-cov
        pytest-mock
 
 [testenv:py27]
-deps =
-    {[testenv]deps}
-    mock
+deps = {[testenv]deps}
```
michaeljoseph: We could remove these sections entirely now right?

hackebrot: Yes, Sir! :ok_hand:

michaeljoseph:
:+1:

hackebrot:
:triumph: 

I'll rebase if you think we are all set here. LGTM

vincentbernat:
The previous ways of doing things was generic. Due to the way tox works, it is not a general solution as we may wish to be able to test an installed version of cookiecutter without using virtualenv. `python setup.py test` should be able to test the same environment that `python setup.py install` configures.

PR #483, opened 2015-07-30
Fix flake8 in tests dir

hackebrot:
This PR fixes all of the `flake8` issues that I discovered while working on #480.

Once we got both merged, I will enable `flake8` checks for the tests dir in addition to the `cookiecutter` package in `tox.ini`. :triumph: 

cc @michaeljoseph


michaeljoseph:
:smile_cat: :+1:

PR #484, opened 2015-07-30
Implement tests for identify_repo

hackebrot:
This PR adds tests to cover `vcs.identify_repo`.

I think we should try to come up with a better solution rather than simple string analysis to detect repo types in the long run though. Any ideas? We could at least have a regex that validates if we are dealing with a **valid** bitbucket or github link.

For instance:

```bash
cookiecutter https://gitfoo
```
causes

```bash
...
subprocess.CalledProcessError: Command '[u'git', u'clone', u'https://gitfoo']' returned non-zero exit status 128
```

:worried:


pydanny:
Not sure what to tell you. I worked on something to try and do it a couple years ago: https://github.com/pydanny/watdarepo

hackebrot:
Hi @pydanny no worries. I'll have a look at your lib, thank you for sharing! :bow:

PR #487, opened 2015-08-04
Add tests for cli invocation

hackebrot:
This PR adds very basic tests for the cli invocation for cookiecutter introduced in #449.

Albeit not strictly required, I'd prefer to have a bit of coverage for this. :no_mouth: 

Please let me know your thoughts @vincentbernat @pydanny :bow:

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -74,5 +74,5 @@ def main(template, no_input, checkout, verbose):
         click.echo(e)
         sys.exit(1)
 
-if __name__ == "__main__":  # pragma: no cover
+if __name__ == "__main__":
     main()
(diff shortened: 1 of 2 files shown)


vincentbernat:
I think you need to set `PYTHONPATH` or you may test an installed version. Also, there you should get the current Python executable or you could test another Python (I think `sys.executable` would be OK).

hackebrot:
@vincentbernat like this? 7024d3b

Been trying to skip the sdist setup in tox for specific tests but yeah...not too much success with it up to now. :worried:

vincentbernat:
Yes, I think it's OK.

hackebrot:
Any objections, @pydanny?

pydanny:
LGTM!

PR #489, opened 2015-08-06
Leverage click version option

hackebrot:
This PR makes use of **clicks** feature of printing the package version. It will not break the current format though.

```
Cookiecutter 1.0.0 from /Users/raphael/hackebrot/cookiecutter (Python 2.7)
```

If you agree on the following **click** default message, I'll get rid of ``version_msg`` altogether:

```
cookiecutter, version 1.0.0
```

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -35,6 +35,7 @@ def print_version(context, param, value):
 
 
 @click.command()
+@click.version_option()
 @click.argument('template')
 @click.option(
     '--no-input', is_flag=True,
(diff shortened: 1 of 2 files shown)


michaeljoseph:
:+1: to the default click version message.

hackebrot:
@michaeljoseph  There we go :smile_cat:

michaeljoseph:
:cake: lgtm

pydanny:
While this in theory cleans up the output of our version reporting, it potentially causes issues for those who used it to find the location of their cookiecutter command. While this isn't a big issue for POSIX systems with the `which` command, it removes a feature that Windows users may or may not be using.

Until we get feedback from Windows users I'm hesitant to include this one.

hackebrot:
Many thanks @pydanny :bow:

I'll see if I can play around with it on my Win box...If I'm not mistaken Windows users should still be able to run ``where cookiecutter``.

audreyr:
I actually prefer the more verbose error message, since a lot of Cookiecutter users are beginners who don't know how to do `which` or `where`.

hackebrot:
:point_right: #506

PR #490, opened 2015-08-07
Handle all types of repo, including internal

charlax:


hackebrot:
Thank you for your contribution @charlax! :bow:

Can you possibly change the regex to use inline comments for the sake of comprehensibility?

charlax:
For sure! Do you really think that's necessary though? I fear this will make the regex look more complicated than it actually is. Let me know and I'll do it.

hackebrot:
Yeah, I feel it helps other contributors who may not be as familiar with regex and even yourself in a few months from now :grin: 

I feel something like this is way easier on the eyes.
http://www.diveintopython.net/regular_expressions/verbose.html

hackebrot:
Related to #484

hackebrot:
Please note that we also support bitbucket/hg :no_mouth:

charlax:
Added the comments and made the regex more generic!

charlax:
@hackebrot do you mind reviewing?

cookiecutter/main.py
```diff
@@ -27,6 +28,20 @@
...
+ |                   # or
+ (\w+@[\w\.]+)       # something like user@...
+)
+.*
+"""
+
+
+def is_repo_url(value):
+    """Return True if value is a repository URL."""
+    return bool(re.match(REPO_REGEX, value, re.X))
```
hackebrot: Let's change this to ``re.VERBOSE``. :smile_cat:

hackebrot:
I'm still at work. Expect a thorough review later that night. Brb!

tests/test_main.py
```diff
@@ -0,0 +1,11 @@
...
+
+
+def test_is_repo_url():
+    """Verify is_repo_url works."""
+    assert is_repo_url('gitolite@server:team/repo') is True
+    assert is_repo_url('git@github.com:audreyr/cookiecutter.git') is True
+    assert is_repo_url('https://github.com/audreyr/cookiecutter.git') is True
+
+    assert is_repo_url('/audreyr/cookiecutter.git') is False
+    assert is_repo_url('/home/audreyr/cookiecutter') is False
```
hackebrot: Can you please add a test for mercurial?

charlax: For sure.

charlax:
Added a test for Mercurial.

tests/test_main.py
```diff
@@ -0,0 +1,12 @@
+from cookiecutter.main import is_repo_url
+
+
+def test_is_repo_url():
+    """Verify is_repo_url works."""
+    assert is_repo_url('gitolite@server:team/repo') is True
+    assert is_repo_url('git@github.com:audreyr/cookiecutter.git') is True
+    assert is_repo_url('https://github.com/audreyr/cookiecutter.git') is True
+    assert is_repo_url('gh:audreyr/cookiecutter-pypackage') is True
```
hackebrot: I am afraid this is not a test for ``mercurial``, but for the abbreviations.

I was thinking of testing something like ``https://bitbucket.org/pokoli/cookiecutter-trytonmodule.hg`` :no_mouth:

charlax: I understand the thoroughness concern, but isn't that overkill since there's no semantic difference with the test testing `https://github...`?

hackebrot: Someone may change this very method to only work for *github* and we wouldn't notice a problem testing-wise, so IMHO it makes sense to have a simple test for it.

hackebrot:
LGTM :+1:

pydanny:
LGTM :+1:

Issue #493, opened 2015-08-08
Allow overriding existing files


ionelmc:
I've tried to use the cookiecutter from 75590ed66ca7ad14df62c4c43f53dd35253f606c but it give me an `Error: "python-nameless" directory already exists` if I try to regenerate a project.

Can this new behavior be reverted? What's the reason for disallowing updates?

ionelmc:
Seems it was introduced in https://github.com/audreyr/cookiecutter/commit/5c22467db2ecb8d7fa36a16f8a0162be4fb3207e 

But there should be a way to switch this off. The original issue (#470) was about a prompt :-)

hackebrot:
Please have a look at #475. Not too sure how to proceed on this.

Do you think overwriting an existing directory is the better option?

ionelmc:
Override projects all the time. It works quite well as all my projects are under version control - so I can just revert if some file need not be changed.

hackebrot:
Wouldn't it also remove your ``.git`` folder? :no_mouth:

ionelmc:
Cookiecutter doesn't copy `.git` over. Not sure what's the reason.

ionelmc:
This is the reason :) https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/find.py#L30

That directory will not have any `.git` in it. Template authors don't store a second `.git` dir in their templates AFAIK

Issue #475, opened 2015-07-15
fail_if_exists not used


schacki:
Based on the discussion in #471  and #470  and based on the code, I understand that this should be optional and the behavior should be controlled by the boolean variable "fail_if_exists". However, "fail_if_exists" is currently hardcoded to True in the call to render_and_create_dir. At the same time, the default value in render_and_create_dir is set to False. This seems contradictory and not intended.
In my view the behavior should be controllable by the user, e.g. via command line argument.

hackebrot:
cc @lgp171188

lgp171188:
@schacki @hackebrot The reason why I chose the default for ``fail_if_exists`` as ``False`` and overrode in one particular place is to keep other code which depend on the previous behaviour of render_and_create_dir from breaking. It looks to be a bad choice.

One way to fix it is to toggle the default and explicitly make ``fail_if_exists`` to be ``False`` wherever required.

As for the command-line argument, we can have ``fail_if_exists=True`` as the default behaviour (since it is never good to automatically overwrite some thing that already exists) and we can expose a command-line flag, say ``--overwrite-if-exists`` which will allow the user to override it.

Please let me know your thoughts on this.

hackebrot:
cc @pydanny @michaeljoseph

hackebrot:
:bell:

PR #495, opened 2015-08-10
Add a flag to overwrite existing output directory

lgp171188:
Fix issue #493 and allow the user to alter the default behaviour
behaviour of not overwriting the contents of the output directory
if it already exists. Fix issue #475 by removing the conflicting
use of 'fail_if_exists' arguments and refactoring it.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -84,11 +84,10 @@ def cookiecutter(
     :param: overwrite_if_exists: Overwrite the contents of output directory
         if it exists
     """
-    if replay and ((no_input is not False) or (extra_context is not None) or
-                   (overwrite_if_exists is not False)):
+    if replay and ((no_input is not False) or (extra_context is not None)):
         err_msg = (
-            "You can not use replay with no_input, extra_context"
-            " or overwrite_if_exists"
+            "You can not use both replay and no_input or extra_context "
+            "at the same time."
         )
         raise InvalidModeException(err_msg)
 
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -70,8 +70,8 @@ def expand_abbreviations(template, config_dict):
 
 
 def cookiecutter(
-        template,
-        checkout=None, no_input=False, extra_context=None, replay=False):
+        template, checkout=None, no_input=False, extra_context=None,
+        replay=False, overwrite_if_exists=False):
     """
     API equivalent to using Cookiecutter at the command line.
 
@@ -81,11 +81,14 @@ def cookiecutter(
     :param no_input: Prompt the user at command line for manual configuration?
     :param extra_context: A dictionary of context that overrides default
         and user configuration.
+    :param: overwrite_if_exists: Overwrite the contents of output directory
+        if it exists
     """
-    if replay and ((no_input is not False) or (extra_context is not None)):
+    if replay and ((no_input is not False) or (extra_context is not None) or
+                   (overwrite_if_exists is not False)):
         err_msg = (
-            "You can not use both replay and no_input or extra_context "
-            "at the same time."
+            "You can not use replay with no_input, extra_context"
+            " or overwrite_if_exists"
         )
         raise InvalidModeException(err_msg)
 
@@ -130,5 +133,6 @@ def cookiecutter(
     # Create project from local context and project template.
     generate_files(
         repo_dir=repo_dir,
-        context=context
+        context=context,
+        overwrite_if_exists=overwrite_if_exists
     )
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -76,7 +76,8 @@ def test_cli_replay(mocker):
         template_path,
         None,
         False,
-        replay=True
+        replay=True,
+        overwrite_if_exists=False
     )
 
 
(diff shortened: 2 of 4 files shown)


lgp171188:
cc @hackebrot @schacki @ionelmc

cookiecutter/main.py
```diff
@@ -52,7 +52,8 @@ def expand_abbreviations(template, config_dict):
     return template
 
 
-def cookiecutter(template, checkout=None, no_input=False, extra_context=None):
+def cookiecutter(template, checkout=None, no_input=False,
```
hackebrot: I am afraid this breaks backwards compatibility :worried: 

We should append the new kwarg at the very end.

lgp171188: The reason I inserted it in the middle rather than at the end is that in ``cookiecutter/cli.py``, which is the only place this function is called, this function is invoked in this way - ``cookiecutter(template, checkout, no_input)``. This implies ``extra_context`` is an optional argument. But the flag ``overwrite_if_exists`` is always passed to all the functions that require it and putting it at the end will require code change like one of these 2 ways, which don't look that good imho.
    
    # first way
    extra_context = None
    cookiecutter(template, checkout, no_input, extra_context, overwrite_if_exists)

    # second way
    cookiecutter(template, checkout, no_input, overwrite_if_exists=overwrite_if_exists)

Please suggest better ways to deal with this, if any.

ionelmc: There might still be external code depending on the position of arguments. Quite unfortunate as Python 2.x doesn't have keyword-only arguments.

lgp171188: @ionelmc @hackebrot Okay, I will wait for all the other review comments, if any, and fix this along with the other issues pointed out.

hackebrot: I would most certainly go for kwargs as the default value for ``extra_context`` *could* change in the future (I have doubts it wil :wink:).

ionelmc:
Anyone still working on this?

hackebrot:
Nope, I am not. I'm still in favor of using kwargs :smirk:

pydanny:
:-1:  on breaking backwards compatibility. 

FWIW cookiecutter is used in a lot of places and packages, including as an imported package. If we accept this PR as it stands the issues board will be filled with help requests.

lgp171188:
@pydanny I will fix the backward compatibility issues pointed out by @hackebrot. I was just waiting for all the review comments which would require a change so that I could do all of them in one shot. Will make the suggested changes and get back.

lgp171188:
@hackebrot I have updated the pull request not to break the backward compatibility and also rebased on top of the latest changes in the master branch. Please review. /cc @ionelmc 

I also see a new replay feature added to the codebase. Not sure how ``--overwrite-if-exists`` should work with it. At the moment, I have made cookiecutter throw an error if it ``-f`` or ``--overwrite-if-exists`` is used along with replay since I am not sure about it.

What about defaulting to ``--overwrite-if-exists`` when using replay?

lgp171188:
Ping @hackebrot :)

hackebrot:
Hi @lgp171188! Apologies for my late reply. :no_mouth: 

I think it should be possible to **replay** and **overwrite** at the same time.

cookiecutter/main.py
```diff
@@ -70,8 +70,8 @@ def expand_abbreviations(template, config_dict):
 
 
 def cookiecutter(
-        template,
-        checkout=None, no_input=False, extra_context=None, replay=False):
+        template, checkout=None, no_input=False, extra_context=None,
+        replay=False, **kwargs):
```
hackebrot: I'd prefer to use ``overwrite_if_exists=False`` instead of ``**kwargs``.

pydanny: I agree. Explicit over implicit!

pydanny:
@lgp171188 This looks great. If you could just change the use of `kwargs` to an explicit argument, then I think this is ready to merge in.

Thank you for all your hard work!

cookiecutter/main.py
lgp171188: @pydanny @hackebrot It looks like misinterpreted the previous comments on using kwargs and thought that I should use the ``**kwargs`` structure. My bad. Will fix that :)

lgp171188:
Details of the most recent changes to this PR:
* Make ``overwrite_if_exists`` as an explicit argument instead of using it via the implicit ``**kwargs`` argument
* Make the ``overwrite-if-exists`` flag independent of the replay feature as suggested by @hackebrot

tests/test_cli.py
```diff
@@ -101,11 +102,56 @@ def test_cli_exit_on_noinput_and_replay(mocker):
...
         template_path,
         None,
         True,
-        replay=True
+        replay=True,
+        overwrite_if_exists=False
     )
+
+@pytest.mark.usefixtures('remove_fake_project_dir')
+def test_cli_exit_on_overwrite_if_exists_and_replay(mocker):
```
hackebrot: This test is not valid anymore as we do support using both options simultaneously.

tests/test_cli.py
```diff
@@ -101,11 +102,56 @@ def test_cli_exit_on_noinput_and_replay(mocker):
...
+        None,
+        False,
+        replay=True,
+        overwrite_if_exists=True
+    )
+
+
+@pytest.mark.usefixtures('remove_fake_project_dir')
+def test_cli_overwrite_if_exists_when_output_dir_does_not_exist():
+    if os.path.isdir('fake-project'):
```
hackebrot: I think we should move this code into a fixture.

tests/test_cli.py
hackebrot: Or we should at least rename it :smile_cat:

lgp171188: @hackebrot I left this test intact because I felt that it was good to test both the options together and verify that the command exits normally.

tests/test_cli.py
lgp171188: I did not move it into a fixture because there is no other test as of now which uses similar code. Will move it to a fixture if you strongly feel so.

There is already a ``fin_remove_fake_project_dir`` fixture function but it is a nested function and set as a finalizer. So there is no scope to reuse it.

tests/test_cli.py
hackebrot: I see. :no_mouth: 

The name is a bit misleading, I'm afraid. I thought it is supposed to test that **replay** and **overwrite** cause a the cli to exit without running **cookiecutter** (see ``test_cli_exit_on_noinput_and_replay`` which asserts for an error exit code).

lgp171188: Then is it okay to rename the test to ``test_cli_successful_exit_on_overwrite_if_exists_and_replay`` or maybe ``test_cli_does_not_exit_with_failure_on_overwrite_if_exists_and_replay``? :-)

hackebrot:
Many thanks for your work @lgp171188! :bow:

pydanny:
Great work @lgp171188! :+1:

PR #498, opened 2015-08-12
Raise exception for undefined variables.

### LINKED ISSUE #111: Unknown placeholders are silently ignored (replaced with empty string)
VERSION: cookiecutter-0.7.0

Unknown placeholders in dirname/filename or Jinja templates are silently ignored and replaced with empty string. This is BAD because neither the template provider nor the template user detects these problems easily.

```python
# FILE: cookiecutter/generate.py
# FUNCTION: generate_files()
    # -- CURRENTLY:
    env = Environment()
    # -- SHOULD-BE:
    env = Environment(undefined=StrictUndefined)
```

ojongerius:
Fixes #111.

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -16,7 +16,7 @@ import logging
 import os
 import shutil
 
-from jinja2 import FileSystemLoader, Template
+from jinja2 import FileSystemLoader, Template, StrictUndefined
 from jinja2.environment import Environment
 from jinja2.exceptions import TemplateSyntaxError
 from binaryornot.check import is_binary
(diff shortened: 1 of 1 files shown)


hackebrot:
Hi @ojongerius, thank you for your contribution.

I'm afraid this does not solve #111, as @audreyr and the others agreed on catching the exception and printing an according warning and aborting the process. :worried:

ojongerius:
Hey @hackebrot . Raising an exception solves my issue -I want generation to fail, and know why. But I also noticed some tests seem to rely on the current behaviour :unamused:  We could easily change the raise into a warning and abort.

ojongerius:
@hackebrot thanks for you answer, I've put a comment in #111 to see if we can agree on an implementation strategy.

I'll keep this PR open for now, and will rework if reach an agreement.

ojongerius:
TTL expired. Closing this one.

PR #499, opened 2015-08-14
Added requirement.txt file to easily add  and install dependencies

sp1rs:


hackebrot:
Hi @sp1rs, thank you for submitting a PR.

Could you please elaborate on why this change is necessary? I am afraid I don't get quite why this is an improvement over having all the dependencies in ``setup.py``.

As this also requires the user to have ``pip`` installed, I am :-1: for now.

sp1rs:
@hackebrot  This what inspires me to make this changes .. https://pip.readthedocs.org/en/1.1/requirements.html
this link will give you the brief explanation of why i think we  require requirement.txt

pydanny:
There doesn't seem to be any existing problem that having a `requirements.txt` file would solve. Besides an obscure edge case https://pip.readthedocs.org/en/1.1/requirements.html doesn't provide any reason why we need this file in cookiecutter.

pydanny:
Closing as unnecessary. I'm willing to listen to explanations, so am willing to reopen in the future.

PR #506, opened 2015-08-28
Use click version

hackebrot:
Make use of ``click.version_option`` without changing the current prompt.

This PR also adds a test for the ``--version`` cli option.

Based on the feedback to #489.

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -37,6 +37,7 @@ def print_version(context, param, value):
 
 
 @click.command()
+@click.version_option()
 @click.argument('template')
 @click.option(
     '--no-input', is_flag=True,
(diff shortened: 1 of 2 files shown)


michaeljoseph:
While I'm not averse to this over #506, I'm a little confused about how someone could invoke `cookiecutter --version` to figure out the location of `cookiecutter` without knowing it already?

hackebrot:
I don't really have an answer to this @michaeljoseph. I feel few tools actually provide this feature, but I am not opposed to sticking to the feedback from @pydanny and @audreyr.

One of the use cases I could think of would be to figure out if you are running ``cookiecutter`` from your global Python installation or in a virtual environment. :no_mouth:

michaeljoseph:
Yeah, this isn't really an issue, we can preserve the existing behaviour.
:+1:

PR #507, opened 2015-08-30
Add a flag to overwrite existing output directory

ionelmc:
I grew a bit impatient and rebased #495.

It should have the positional argument also fixed.


hackebrot:
Thanks @ionelmc!

Think we can close this guy though. Guess we'll merge #495 any time soon.

PR #508, opened 2015-09-02
Export tox environment variables to codecov

maiksensi:
This will export tox environment variables such as py27, py33, ... to codecov. Codecov will then show these variables when an individual travis build run is selected. This is very handy to see what environment is selected, when a individual environment is selected for coverage results instead of the unified results. Here is an example how this looks in the UI for this PR: https://codecov.io/github/audreyr/cookiecutter?ref=533143d96a2a879b69689ca30b83593fcd7cb7fb&build=1051.2

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,9 +13,9 @@ env:
 script: tox -e $TOX_ENV
 
 install:
-    - pip install tox
+    - pip install tox codecov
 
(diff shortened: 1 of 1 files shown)


hackebrot:
Thank you @maiksensi! :+1:

PR #513, opened 2015-09-17
Add Option class for customising the prompt, validating user input

timheap:
The Option class allows for customising the prompt displayed when asking for user input, as well as opening up future possibilities of more complicated features, such as type conversion and validation.

The tests and documentation have not yet been updated. I would like to discuss this proposal further before investing the time to change the tests.

The new JSON syntax is backwards compatible with the old syntax, but adds some new features, most of which are covered in the sample `cookiecutter.json` below:

``` json
{
    "old_style_string": "with default",
    "old_style_choice": ["choice", "list"],
    "project_name": {
        "prompt": "Project name"
    },
    "repo_name": {
        "prompt": "The name of the git repository on Github, Bitbucket, etc",
        "default": "{{ cookiecutter.project_name|lower|replace(' ', '-') }}"
    },
    "pkg_name": {
        "prompt": "The name of the Python package",
        "default": "{{ cookiecutter.repo_name|replace('-', '') }}"
    },
    "port": {
        "prompt": "Default port number",
        "default": 8000,
        "type": "int",
        "min": 1024,
        "max": 65530
    },
    "license": {
        "prompt": "License",
        "choices": ["BSD", "GPL", "MIT", "Apache", "CC"],
        "type": "choice"
    }
}
```

At this point, I have only committed a basic `StringOption` and `ChoiceOption`, to cover the old functionality. I have made an `IntOption` as well though, for testing. It converts its input to an integer, and can validate that it is within a certain range. It is used in the example above, and its implementation is below:

``` python
@option_types.register
class IntOption(Option):
    type_name = 'int'

    def __init__(self, min=None, max=None, **kwargs):
        self.min = min
        self.max = max
        super(IntOption, self).__init__(**kwargs)

    def clean(self, value):
        value = super(IntOption, self).clean(value)

        try:
            value = int(value)
        except ValueError:
            raise ValidationException('"{0}" is not a valid integer'.format(value))

        if self.min is not None and value < self.min:
            raise ValidationException('The value must be greater than {0}'.format(
                self.min))

        if self.max is not None and value > self.max:
            raise ValidationException('The value must be less than than {0}'.format(
                self.max))

        return value
```

What do you think?

### DIFF
--- a/cookiecutter/compat.py
+++ b/cookiecutter/compat.py
@@ -8,6 +8,7 @@ if PY3:  # pragma: no cover
     input_str = 'builtins.input'
     iteritems = lambda d: iter(d.items())
     from io import StringIO
+    text_type = str
 
 
 else:  # pragma: no cover
@@ -16,6 +17,7 @@ else:  # pragma: no cover
     input_str = '__builtin__.raw_input'
     iteritems = lambda d: d.iteritems()
     from cStringIO import StringIO
+    text_type = unicode
 
 
 def is_string(obj):
--- a/cookiecutter/exceptions.py
+++ b/cookiecutter/exceptions.py
@@ -81,3 +81,9 @@ class InvalidModeException(CookiecutterException):
     Raised when cookiecutter is called with both `no_input==True` and
     `replay==True` at the same time.
     """
+
+
+class ValidationException(CookiecutterException):
+    """
+    Raised when some input to a prompt does not pass validation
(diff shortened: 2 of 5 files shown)


hackebrot:
Thank you for this PR @timheap and apologies for the very late response! 😿 

I'm closing this in favor of #848. Your feedback on the new context is very much appreciated.

PR #520, opened 2015-09-19
Adding `cookiecutter-django-rest` to list of Python / Django templates.

agconti:


audreyr:
Nice work, thanks @agconti.

agconti:
@audreyr thanks! thank you for merging.

PR #522, opened 2015-09-20
Enable python 35

hackebrot:
Python 3.5 is out and we don't seem to have any incompatibilities :rocket:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ pip-log.txt
 # Unit test / coverage reports
 .coverage
 .tox
+.cache
 nosetests.xml
 
 # Translations
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,6 @@ env:
     - TOX_ENV=py27
     - TOX_ENV=py33
     - TOX_ENV=py34
-    - TOX_ENV=py35
     - TOX_ENV=pypy
     - TOX_ENV=flake8
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,7 @@ env:
     - TOX_ENV=py27
     - TOX_ENV=py33
     - TOX_ENV=py34
+    - TOX_ENV=py35
     - TOX_ENV=pypy
     - TOX_ENV=flake8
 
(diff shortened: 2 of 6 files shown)


hackebrot:
Hmmm.. https://github.com/mitsuhiko/jinja2/pull/495/files

maiksensi:
How should we proceed? I suggest we integrate the changes from #524. 
To get this working (almost) merge my changes from PR #523 to your branch. Then reapply all your changes, but change the ```appveyor.yml``` accordingly to the changes from PR #523 . 
To get Travis-CI working integrate my changes, which change the base python install (and thus downloads it): https://github.com/maiksensi/cookiecutter/blob/feat/support-for-py35/.travis.yml
This will result in one last failing check (flake8). Which I didn't have a look at yet.

pydanny:
Err... I just merged in #523. Maybe that was premature, but in hindsight this PR (#522) should be built off of it.

hackebrot:
Rebase :smile_cat:

maiksensi:
I am not sure, but I think a newer '''tox.ini''' (with py35 in the matrix) slipped through, which will not work in travis currently.

hackebrot:
Yes, see 8e69a492c.

As it is working completely fine locally, we can possibly deactivate it on travis only.

What's your opinion on this?

audreyr:
I think deactivating 3.5 on Travis only for now would be alright.

zoidbergwill:
What about configuring tox in Travis using a matrix like [pip does](https://github.com/pypa/pip/blob/develop/.travis.yml#L4-L33)?

maiksensi:
@zoidbergwill 
> What about configuring tox in Travis using a matrix like pip does? 

Together with Tox, this would move build logic from Tox to Travis which I think is confusing. We would also need to change the tox config, as the pip like configuration would create a matrix with many redundant tox tests. I would also just wait till travis-ci adds py35 as an alias, as all tests run fine locally. There's also appveyor as "backup" CI which already works fine with py35.

hackebrot:
The scope of this PR slightly changed. :no_mouth: 

It adds Python 3.5 to the docs and appveyor. That's pretty much it.

Please let me know your feedback!

maiksensi:
:+1:

PR #529, opened 2015-09-24
Use future instead of custom compat

### LINKED ISSUE #298: Address test coverage issues for compat.py

### LINKED ISSUE #301: Consider moving to six or future for Python 2/3 compatibility

hackebrot:
Remove ``compat.py`` and use [future](https://pypi.python.org/pypi/future) for ``basestring`` and ``iteritems``.

Resolve #298 with regards to #438. Also resolve #301 cc @michaeljoseph @pydanny

### DIFF
--- a/cookiecutter/__init__.py
+++ b/cookiecutter/__init__.py
@@ -7,10 +7,10 @@ cookiecutter
 
 Main package for Cookiecutter.
 """
-from .compat import OLD_PY2
+import sys
 
 __version__ = '1.0.0'
 
-if OLD_PY2:
+if sys.version_info[:2] < (2, 7):
     msg = 'Python 2.6 support was removed from cookiecutter in release 1.0.0.'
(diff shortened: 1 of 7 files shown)


audreyr:
Just reviewed it in detail, :+1:

pydanny:
Lovely! :+1:

PR #542, opened 2015-09-29
Choice with default in user config

### LINKED ISSUE #536: Choice prompts are disabled if there's a ~/.cookiecutterrc
Seems I don't get the nice prompts from https://github.com/audreyr/cookiecutter/pull/455 if I have a `~/.cookiecutterrc`

hackebrot:
This PR makes sure to factor in choice variables in templates when using a user config or extra context.

Instead of replacing the list, it will set the *overwrite* value as the new default for the choice variable.

Please note that 8f08298 slightly changes the current behavior as it ignores variables which are not used in the template as opposed to ``dict.update({})``.

Resolve #536

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -51,6 +51,7 @@ def copy_without_render(path, context):
 
 
 def apply_overwrites_to_context(context, overwrite_context):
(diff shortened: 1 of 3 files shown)


hackebrot:
Thank you @pydanny :bow:

PR #548, opened 2015-10-06
Bool config

### LINKED ISSUE #126: Allow boolean config variables
`cookiecutter.json` allows a key-value configuration, where value is a string. I would like to ask for a boolean (yes/no) variable to condition a block like this

``` jinja

    {% if use_that_magic %}
         ....
    {% endif %}
```

The easiest way should be autodetect the variable type from the default value. For example 

``` json

{ "full_name": "Martín Gaitán",
  "use_that_magic": true,  
   ...
}
```

When prompt for `use_that_magic` the valid answer should be yes or no. 

Also, could be an attribute `type` when implement #30

sbuss:
If the default value of an entry in `cookiecutter.json` is a boolean then preserve further user input as boolean using read_user_yes_no.

For example, the json:

``` json
{
    "is_foo": true
}
```

would result in a prompt like

``` sh
is_foo [True]:
```

Then if a user input "false", "False", "n", etc it would result in a False value in python.

Fixes #126

### DIFF
--- a/tests/test-generate-context/choices_template.json
+++ b/tests/test-generate-context/choices_template.json
@@ -3,6 +3,5 @@
     "github_username": "hackebrot",
     "project_name": "Kivy Project",
     "repo_name": "{{cookiecutter.project_name|lower}}",
-    "orientation": ["all", "landscape", "portrait"],
-    "is_alive": true
+    "orientation": ["all", "landscape", "portrait"]
 }
--- a/tests/test-generate-context/choices_template.json
+++ b/tests/test-generate-context/choices_template.json
@@ -3,5 +3,6 @@
     "github_username": "hackebrot",
     "project_name": "Kivy Project",
     "repo_name": "{{cookiecutter.project_name|lower}}",
-    "orientation": ["all", "landscape", "portrait"]
+    "orientation": ["all", "landscape", "portrait"],
+    "is_alive": true
 }
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -212,7 +212,16 @@ class TestPromptChoiceForConfig(object):
     def context(self, choices):
         return {
             'cookiecutter': {
-                'orientation': choices
+                'orientation': choices,
+                'is_alive': True,
+            }
+        }
+
+    @pytest.fixture
+    def bool_context(self, choices):
+        return {
+            'cookiecutter': {
+                'is_alive': True,
             }
         }
 
@@ -247,3 +256,32 @@ class TestPromptChoiceForConfig(object):
         )
         read_choice.assert_called_once_with('orientation', choices)
         assert expected_choice == actual_choice
+
+    def test_should_read_yes_no(self, mocker, bool_context):
+        key = 'is_alive'
+        expected_bool = False
+
+        read_bool = mocker.patch('cookiecutter.prompt.read_user_yes_no')
+        read_bool.return_value = expected_bool
+
+        config = prompt.prompt_for_config(
+            bool_context,
+            False  # Ask the user for input
+        )
+        read_bool.assert_called_once_with(key, 'True')
+        assert config[key] == expected_bool
+
+    def test_should_read_yes_no_false(self, mocker, bool_context):
+        key = 'is_alive'
+        expected_bool = True
+
+        read_bool = mocker.patch('cookiecutter.prompt.read_user_yes_no')
+        read_bool.return_value = expected_bool
+
+        bool_context['cookiecutter'][key] = False
+        config = prompt.prompt_for_config(
+            bool_context,
+            False  # Ask the user for input
+        )
+        read_bool.assert_called_once_with(key, 'False')
+        assert config[key] == expected_bool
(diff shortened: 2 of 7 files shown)


myyk:
Nice this looks great!

tests/test-generate-context/choices_template.json
```diff
@@ -3,5 +3,6 @@
     "github_username": "hackebrot",
     "project_name": "Kivy Project",
     "repo_name": "{{cookiecutter.project_name|lower}}",
-    "orientation": ["all", "landscape", "portrait"]
+    "orientation": ["all", "landscape", "portrait"],
+    "is_alive": true
```
hackebrot: I'd prefer to have a separate test file for booleans variables.

We might have a hard time debugging test failures. Can you please set up a dedicated test for booleans?

hackebrot:
Many thanks for submitting this @sbuss :bow:

I am :+1: as to whether we should have boolean variables, yet I would like to see separate tests for the feature rather than updating choices tests.

sbuss:
@hackebrot I've added separate bool tests in 55c2517, but I don't think it added much value, since the values in the context json are not parsed, so a `'True'` string would still be treated as a string. Do you want that behavior to change? Currently those strings are only parsed when reading from a prompt, since that was where bools would get overwritten with strings.

sbuss:
@hackebrot once you're happy with the changes I'll rebase to get rid of the intermediate commits.

tests/test_prompt.py
```diff
@@ -212,7 +212,16 @@ def choices(self):
     def context(self, choices):
         return {
             'cookiecutter': {
-                'orientation': choices
+                'orientation': choices,
+                'is_alive': True,
```
hackebrot: ``context`` does not need to change, if I am not mistaken.

tests/test_prompt.py
```diff
@@ -247,3 +256,32 @@ def test_should_read_userchoice(self, mocker, choices, context):
...
+        read_bool.return_value = expected_bool
+
+        config = prompt.prompt_for_config(
+            bool_context,
+            False  # Ask the user for input
+        )
+        read_bool.assert_called_once_with(key, 'True')
+        assert config[key] == expected_bool
+
+    def test_should_read_yes_no_false(self, mocker, bool_context):
```
hackebrot: What are you trying to achieve with this test? :confused:

tests/test_prompt.py
```diff
@@ -212,7 +212,16 @@ def choices(self):
...
         return {
             'cookiecutter': {
-                'orientation': choices
+                'orientation': choices,
+                'is_alive': True,
+            }
+        }
+
+    @pytest.fixture
+    def bool_context(self, choices):
```
hackebrot: Please remove ``choices`` from this fixture.

tests/test_prompt.py
```diff
@@ -247,3 +256,32 @@ def test_should_read_userchoice(self, mocker, choices, context):
...
+        assert config[key] == expected_bool
+
+    def test_should_read_yes_no_false(self, mocker, bool_context):
+        key = 'is_alive'
+        expected_bool = True
+
+        read_bool = mocker.patch('cookiecutter.prompt.read_user_yes_no')
+        read_bool.return_value = expected_bool
+
+        bool_context['cookiecutter'][key] = False
```
hackebrot: In case you are trying to test both ``True`` and ``False``, you could use a parametrized fixture.

```python
@pytest.fixture(params=[True, False])
def bool_context(request):
    return {
        'cookiecutter': {
            'is_alive': request.param
        }
    }
```

tests/test_prompt.py
```diff
@@ -247,3 +256,32 @@ def test_should_read_userchoice(self, mocker, choices, context):
         )
         read_choice.assert_called_once_with('orientation', choices)
         assert expected_choice == actual_choice
+
+    def test_should_read_yes_no(self, mocker, bool_context):
+        key = 'is_alive'
+        expected_bool = False
+
+        read_bool = mocker.patch('cookiecutter.prompt.read_user_yes_no')
```
hackebrot: It would be great to also test that neither ``prompt_choice_for_config`` nor ``read_user_variable`` are called for **bool** variables. :smile:

timmartin19:
What's the status on this?  I would love to be able to use this...

sbuss:
@timmartin19 Sorry, this dropped off my TODO list accidentally. I'll address the remaining comments in the next couple days.

michaeljoseph:
:bell: @sbuss

sbuss:
Huh. I totally forgot about this. Is this still wanted or has it been implemented in another way?

michaeljoseph:
Hey @sbuss: yes, we'd still like this feature, thanks 😄

hackebrot:
Please have a look at #848

hackebrot:
Hey @sbuss @michaeljoseph, how do we want to proceed with this? 🤔

hackebrot:
Hey @sbuss! 👋 

Thanks again for your contribution. We'd still like to add support for boolean variables, but I think it makes sense to land it as part of #848 or a version of it. I'm closing this issue now.

I hope to see another pull-request from you in the future! 🍪

Issue #464, opened 2015-06-29
Stop build if pre hook fails


thoppe:
There may be a way to do this already, but the documentation for hooks isn't complete. If we have a pre-hook file that is simply

    raise SyntaxError

it would be nice to halt the cookiecutter process here. What I envision is a pre-hook file that checks for certain external dependencies before continuing. It would be to fail out here before the template is pulled and rendered.

thoppe:
I've asked the question on SO here and I'll post any response I get:

http://stackoverflow.com/q/31128808/249341

pydanny:
I don't believe raising an exception will stop a `pre-hook` right now, as those are run as subprocesses (https://github.com/audreyr/cookiecutter/blob/256cd6f4b3ebd684887406713fb966d829f8d7d1/cookiecutter/hooks.py#L65-L69). Of course, I could be wrong -- I'm just not in a place where I can test this issue right now.

However, if this doesn't exist now, we can add that functionality in a future release.

thoppe:
I found that code block as well. Couldn't we try to catch any errors on the subprocess call and raise an exception right there? It seems like it shouldn't be hard (famous last words). I can try a hack at it if you like, though I've never done any testing suites so I would need help putting a unit test in.

hackebrot:
Hi @thoppe, we merged #549 by @eliasdorneles.

Closing this for now. Please let us know if it helps! Thanks :bow:

thoppe:
Thanks @eliasdorneles!

PR #549, opened 2015-10-08
Stop project generation if pre hook script fails

eliasdorneles:
Hey folks,

Would this be an adequate solution to address issue #464 ?

My use case is handling an invalid package name (see https://github.com/audreyr/cookiecutter-pypackage/issues/6)

This proposal will stop the project generation if the pre hook script returns with an exit status different from 0 (indicating success).

With this proposal, we'd be able to handle that case with a `pre_gen_project` hook script like:

```
import re, sys

PACKAGE_REGEX = r'^[_a-zA-Z][_a-zA-Z0-9]+$'

repo_name = '{{ cookiecutter.repo_name }}'
if not re.match(PACKAGE_REGEX, repo_name):
    print('ERROR: %s is not a valid package name!' % repo_name)
    sys.exit(1)
```

Thoughts?

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -229,7 +229,7 @@ def _run_hook_from_repo_dir(repo_dir, hook_name, project_dir, context):
     """
     with work_in(repo_dir):
         try:
-            run_hook('pre_gen_project', project_dir, context)
+            run_hook(hook_name, project_dir, context)
         except FailedHookException:
             rmtree(project_dir)
             logging.error("Stopping generation because %s"
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -28,7 +28,7 @@ from .exceptions import (
     OutputDirExistsException
 )
 from .find import find_template
-from .utils import make_sure_path_exists, work_in
+from .utils import make_sure_path_exists, work_in, rmtree
 from .hooks import run_hook
 
 
@@ -223,6 +223,20 @@ def ensure_dir_is_templated(dirname):
         raise NonTemplatedInputDirException
 
 
+def _run_hook_from_repo_dir(repo_dir, hook_name, project_dir, context):
+    """
+    Run hook from repo directory, cleaning up project directory if hook fails
+    """
+    with work_in(repo_dir):
+        try:
+            run_hook('pre_gen_project', project_dir, context)
+        except FailedHookException:
+            rmtree(project_dir)
+            logging.error("Stopping generation because %s"
+                          " hook script didn't exit sucessfully" % hook_name)
+            raise
+
+
 def generate_files(repo_dir, context=None, output_dir='.',
                    overwrite_if_exists=False):
     """
@@ -256,15 +270,7 @@ def generate_files(repo_dir, context=None, output_dir='.',
     project_dir = os.path.abspath(project_dir)
     logging.debug('project_dir is {0}'.format(project_dir))
 
-    # run pre-gen hook from repo_dir
-    with work_in(repo_dir):
-        try:
-            run_hook('pre_gen_project', project_dir, context)
-        except FailedHookException:
-            shutil.rmtree(project_dir, ignore_errors=True)
-            logging.error("Stopping generation because pre_gen_project"
-                          " hook script didn't exit sucessfully")
-            raise
+    _run_hook_from_repo_dir(repo_dir, 'pre_gen_project', project_dir, context)
 
     with work_in(template_dir):
         env = Environment(keep_trailing_newline=True)
@@ -320,8 +326,6 @@ def generate_files(repo_dir, context=None, output_dir='.',
                 logging.debug('f is {0}'.format(f))
                 generate_file(project_dir, infile, context, env)
 
-    # run post-gen hook from repo_dir
-    with work_in(repo_dir):
-        run_hook('post_gen_project', project_dir, context)
+    _run_hook_from_repo_dir(repo_dir, 'post_gen_project', project_dir, context)
 
     return project_dir
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -24,11 +24,12 @@ from binaryornot.check import is_binary
 from .exceptions import (
     NonTemplatedInputDirException,
     ContextDecodingException,
+    FailedHookException,
     OutputDirExistsException
 )
 from .find import find_template
 from .utils import make_sure_path_exists, work_in
-from .hooks import run_hook, EXIT_SUCCESS
+from .hooks import run_hook
 
 
 def copy_without_render(path, context):
@@ -257,7 +258,10 @@ def generate_files(repo_dir, context=None, output_dir='.',
 
     # run pre-gen hook from repo_dir
     with work_in(repo_dir):
-        if run_hook('pre_gen_project', project_dir, context) != EXIT_SUCCESS:
+        try:
+            run_hook('pre_gen_project', project_dir, context)
+        except FailedHookException:
+            shutil.rmtree(project_dir, ignore_errors=True)
             logging.error("Stopping generation because pre_gen_project"
                           " hook script didn't exit sucessfully")
             return
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -264,7 +264,7 @@ def generate_files(repo_dir, context=None, output_dir='.',
             shutil.rmtree(project_dir, ignore_errors=True)
             logging.error("Stopping generation because pre_gen_project"
                           " hook script didn't exit sucessfully")
-            return
+            raise
 
     with work_in(template_dir):
         env = Environment(keep_trailing_newline=True)
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -28,7 +28,7 @@ from .exceptions import (
 )
 from .find import find_template
 from .utils import make_sure_path_exists, work_in
-from .hooks import run_hook
+from .hooks import run_hook, EXIT_SUCCESS
 
 
 def copy_without_render(path, context):
@@ -257,7 +257,10 @@ def generate_files(repo_dir, context=None, output_dir='.',
 
     # run pre-gen hook from repo_dir
     with work_in(repo_dir):
-        run_hook('pre_gen_project', project_dir, context)
+        if run_hook('pre_gen_project', project_dir, context) != EXIT_SUCCESS:
+            logging.error("Stopping generation because pre_gen_project"
+                          " hook script didn't exit sucessfully")
+            return
 
     with work_in(template_dir):
         env = Environment(keep_trailing_newline=True)
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -19,11 +19,13 @@ from jinja2 import Template
 
 from cookiecutter import utils
 
+
 _HOOKS = [
     'pre_gen_project',
     'post_gen_project',
     # TODO: other hooks should be listed here
 ]
+EXIT_SUCCESS = 0
 
 
 def find_hooks():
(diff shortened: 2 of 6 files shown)


eliasdorneles:
Okay, coverage bot, I'll add tests and improve coverage once I get some consensus that this is a good-enough approach. :)

hackebrot:
Hi @eliasdorneles, thank you for your contribution! :bow: 

Tbh, I'd rather raise an error instead of returning ``0``. If I am not mistaken there is no error report outside of ``generate`` that the generation failed apart from a log entry.

eliasdorneles:
Hi @hackebrot :)
Right, my first try was raising a custom `HookFailedException`, but then I wasn't sure where it should be caught.
Also, it would raise for all hooks (not only for `pre_gen_project`), but maybe that'd be the desired behavior?

eliasdorneles:
Hey, sorry for the delay -- I updated the code to use a custom exception and added cleaning up the project dir when aborting the generation, and added a little test.
Does this look better?

eliasdorneles:
Note: tests are broken because of issue addressed in this PR: https://github.com/audreyr/cookiecutter/pull/555

eliasdorneles:
Rebased to incorporate tests fix -- would appreciate some feedback. :)

I'm particularly not sure about what should be the handling the `FailedHookException` for the `post_gen_project` case -- which is why I didn't code anything for it.

hackebrot:
I'd say we handle errors for ``post_gen_project`` in the same fashion. :no_mouth: 

More importantly, I feel we should definitely re-raise the exception after cleaning up the filesystem instead of [returning None](https://github.com/audreyr/cookiecutter/pull/549/files#diff-b4c1da8a76ea242bf99a397dcb90b2a6R267). We can catch the ``FailedHookException`` in the callers and ``sys.exit(1)``.

cc @michaeljoseph @audreyr @pydanny

tests/test_hooks.py
```diff
@@ -166,3 +167,16 @@ def test_run_hook(self):
...
+        tests_dir = os.path.join(self.repo_path, 'input{{hooks}}')
+
+        with open(hook_path, 'w') as f:
+            f.write("#!/usr/bin/env python\n")
+            f.write("import sys; sys.exit(1)\n")
+
+        with utils.work_in(self.repo_path):
+            with pytest.raises(exceptions.FailedHookException) as excinfo:
+                hooks.run_hook('pre_gen_project', tests_dir, {})
+            assert 'Hook script failed' in str(excinfo)
```
hackebrot: I think ``assert 'Hook script failed' in excinfo.value`` should work too.

eliasdorneles:
Thanks for reviewing @hackebrot -- I updated it with your suggestions. :)

Only thing, I think it's best not to clean up the filesystem for `post_gen_project` in the same fashion as for `pre_gen_project`. As it is a failure _after_ the project has been generated, the user wouldn't expect it to go away, and removing the project might make harder for users debug the hook scripts for their templates.

hackebrot:
IMHO if a ``post_gen_project`` hook fails, the generated project is likely to corrupt regardless of the actual purpose of the hook. Even if it is just debug logging messages, I expect the hook to exit without an error.

For instance, if the [rename_kv_file hook](https://github.com/hackebrot/cookiedozer/blob/master/hooks/post_gen_project.py) fails, the generated kivy app is corrupt and cannot be launched. :worried: 

So the question is why would you want to keep a broken project? I feel it is rather up to hook to use comprehensive logging messages to help users debug any error that might occur.

eliasdorneles:
Well, I don't have much experience using `post_gen_project` hooks, so if you say that when it fails it's likely the project to be broken, I trust you. :)

I think I was worrying unnecessarily, so lemme update this.

eliasdorneles:
Does this look better?

cookiecutter/generate.py
```diff
@@ -222,6 +223,20 @@ def ensure_dir_is_templated(dirname):
         raise NonTemplatedInputDirException
 
 
+def _run_hook_from_repo_dir(repo_dir, hook_name, project_dir, context):
+    """
+    Run hook from repo directory, cleaning up project directory if hook fails
+    """
+    with work_in(repo_dir):
+        try:
+            run_hook('pre_gen_project', project_dir, context)
```
hackebrot: I guess ``'pre_gen_project'`` is supposed to be ``hook_name``. :no_mouth:

eliasdorneles: ooops, lemme fix that! :D

tests/test_hooks.py
```diff
@@ -166,3 +167,16 @@ def test_run_hook(self):
 
             hooks.run_hook('post_gen_project', tests_dir, {})
             assert os.path.isfile(os.path.join(tests_dir, 'shell_post.txt'))
+
+    def test_run_failing_hook(self):
+        hook_path = os.path.join(self.hooks_path, 'pre_gen_project.py')
+        tests_dir = os.path.join(self.repo_path, 'input{{hooks}}')
+
+        with open(hook_path, 'w') as f:
```
hackebrot: Uhhm, are you changing the test files at runtime? This may break other tests unintentionally... :worried:

eliasdorneles: `self.repo_path` is destroyed in `teardown_method` and recreated in `setup_method` for each test:

https://github.com/audreyr/cookiecutter/pull/549/files#diff-392f380cd08f2c7cc3ac39b10cf80d95R94

hackebrot: You are right :sleepy:

hackebrot:
Great work @eliasdorneles :+1: 

Let's see what the others think :smile_cat:

eliasdorneles:
Cool, thanks for reviewing @hackebrot ! :)

cookiecutter/hooks.py
```diff
@@ -104,4 +110,4 @@ def run_hook(hook_name, project_dir, context):
     if script is None:
         logging.debug('No hooks found')
         return
-    return run_script_with_context(script, project_dir, context)
+    run_script_with_context(script, project_dir, context)
```
pydanny: Why remove the ability to return a value?

pydanny:
This is a really nice feature that deserves documentation ;)

hackebrot:
You are dead right @pydanny! :+1:

*DocsOrDidntHappen* :smile_cat:

cookiecutter/hooks.py
eliasdorneles: well, it was meant as a cleanup, `run_script_with_context` was never returning anything, so I removed the return to make clear this function is a [command not a query](https://en.wikipedia.org/wiki/Command%E2%80%93query_separation).

But I can revert the change, if you have other plans or I missed something. :)

eliasdorneles:
Agreed, lemme add some docs! :)

eliasdorneles:
I added a note under the hook section -- does it look good?

docs/advanced_usage.rst
```diff
@@ -34,6 +34,12 @@ hooks, as these can be run on any platform. However, if you intend for your
 template to only be run on a single platform, a shell script (or `.bat` file
 on Windows) can be a quicker alternative.
 
+.. note::
+    Make sure your hook scripts work in a robust manner.  If a hook script
+    fails (that is, if it finishes with an exit status that is not one of
+    success), the project generation will stop and the generated directory will
```
hackebrot: ...with an exit status that is not ``0``... :grey_question:

hackebrot:
I think the example that you provided in the description of the PR would be a great example for using this new feature. Let's add it to the docs :wink: 

(if @eliasdorneles and @pydanny agree :grin: )

docs/advanced_usage.rst
eliasdorneles: right, I was unsure if mentioning `0` would be more clear or more confusing.
I guess `0` is more clear. :)

hackebrot: You could refer to the [official sys.exit docs](https://docs.python.org/2/library/sys.html#sys.exit) to be very explicit.

>  If it is an integer, zero is considered “successful termination” and any nonzero value is considered “abnormal termination” by shells and the like.

eliasdorneles: Right, though the comment applies to shell script hooks too.
I'll link to it.

eliasdorneles:
I've added an example and improved the wording -- looks good now?

docs/advanced_usage.rst
```diff
@@ -34,6 +34,32 @@ hooks, as these can be run on any platform. However, if you intend for your
...
+    Make sure your hook scripts work in a robust manner. If a hook script fails
+    (that is, `if it finishes with a nonzero exit status
+    <https://docs.python.org/2/library/sys.html#sys.exit>`_), the project
+    generation will stop and the generated directory will be cleaned up.
+
+Example: Validating template variables
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Here is an example of script that validates a template variable
+before generating the project, to be used as ``hooks/pre_gen_project.py``::
```
hackebrot: I feel sorry for being this picky... :speak_no_evil: 

[PEP8 Imports](https://www.python.org/dev/peps/pep-0008/#imports) recommends the following:

```python
import re
import sys
```

Also you could enable syntax highlighting in rst via:

```
to be used as ``hooks/pre_gen_project.py``:

.. code-block:: python

    import re
```

eliasdorneles: haha, I don't mind the nitpicking, keep them coming. :)

who doesn't enjoy some bikeshedding? :bike: :circus_tent: :D

eliasdorneles: although... python highlighting works for me without that markup. :confused:

eliasdorneles: done!

PS: wow, vim highlights sample code inline with `.. code-block::  python` :open_mouth: :notbad:

docs/advanced_usage.rst
```diff
@@ -34,6 +34,35 @@ hooks, as these can be run on any platform. However, if you intend for your
 template to only be run on a single platform, a shell script (or `.bat` file
 on Windows) can be a quicker alternative.
 
+.. note::
+    Make sure your hook scripts work in a robust manner. If a hook script fails
+    (that is, `if it finishes with a nonzero exit status
+    <https://docs.python.org/2/library/sys.html#sys.exit>`_), the project
```
luzfcb: @eliasdorneles It would be nice to indicate the Python3 documentation instead of Legacy Python.

eliasdorneles: :+1:

hackebrot:
:smile_cat: :+1:

eliasdorneles:
Anything else, fellows?
Do you prefer I squash the commits?

hackebrot:
No, you're fine! :+1: 

We are just super busy atm. Apologies for the delay. :worried: 

Going to merge this guy once we get another upvote from a contributor.

eliasdorneles:
gotcha!

btw, in Scrapy we find useful to rename the PR prefixing with `[MRG+1]` for these cases (PRs that were already reviewed and are only waiting for another upvote). =)

pydanny:
LGTM! :+1:

hackebrot:
Great work! :+1: :+1:

eliasdorneles:
Thank you! =)

hackebrot:
I just realized we missed sth :cry: 

We need coverage for ``cookiecutter.generate``:

```python
def _run_hook_from_repo_dir(repo_dir, hook_name, project_dir, context):
    """
    Run hook from repo directory, cleaning up project directory if hook fails
    """
    with work_in(repo_dir):
        try:
            run_hook(hook_name, project_dir, context)
        except FailedHookException:
            rmtree(project_dir)
            logging.error("Stopping generation because %s"
                          " hook script didn't exit sucessfully" % hook_name)
            raise
```

hackebrot:
Do you mind submitting a PR @eliasdorneles?

eliasdorneles:
oh, right.
sure, but it'll take me a few days, I'm on the road to Python Brasil. :)
I'll look at it after the conference.

hackebrot:
No worries! Have fun :smile_cat: 

#580

PR #550, opened 2015-10-09
Make tests work without installing cookiecutter

vincentbernat:
When running tests without installing cookiecutter, some of them fail
because they try to invoke cookiecutter. Moreover, if cookiecutter is
installed on the system, the tests would use the installed version
instead of the one in the build tree.

To solve both issues, use `python -m cookiecutter.cli` instead of
`cookiecutter`.

I know that the official way of testing is using tox, but in some cases, tox is not usable because it requires a network access.


hackebrot:
:+1:

PR #554, opened 2015-10-17
WIP: add virtualenv util for hooks, so that templates can enable virtualenvs

cguardia:
The idea for this PR, is that cookiecutter can offer some utils that templates can use for executing some common post (or pre?) setup tasks with a minimum of work. Currently setting up virtualenvs is the only added hook utility.

The rationale for this is that templates can't require dependencies, so having code in a hook that depends on something gets much more difficult when there is no cookiecutter support. 

I can imagine having a small number of useful utilities like this to make life easier for template authors. The utils let the developer avoid repeating the same code in various templates. Unless they are needed, they sit harmlessly out of the way, except for required dependencies, which would get pulled when installing cookiecutter. This last thing could be considered a point against this approach.

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -14,6 +14,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import virtualenv
 
 from jinja2 import Template
 
@@ -24,6 +25,23 @@ _HOOKS = [
     'post_gen_project',
     # TODO: other hooks should be listed here
 ]
+
+
+def setup_virtualenv(cwd='.', edit_mode=False):
+    """
+    Provide a way for templates to set up virtualenv automatically. Called
+    with the virtualenv path, which will usually be the template output
+    directory. If the edit_mode flag is True, will also run setup.py develop
+    to leave the project in edit mode.
+    """
+    virtualenv.create_environment(cwd)
+    if edit_mode:
+        proc = subprocess.Popen(
+            ['bin/python', 'setup.py', 'develop'],
+            shell=sys.platform.startswith('win'),
+            cwd=cwd
+        )
+        proc.wait()
 
 
 def find_hooks():
--- a/setup.py
+++ b/setup.py
@@ -32,7 +32,8 @@ requirements = [
     'jinja2>=2.7',
     'PyYAML>=3.10',
     'click>=5.0',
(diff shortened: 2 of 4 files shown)


pydanny:
My concern with this pull request is that it might cause significant issues for the Cookiecutter distributions that are done outside of the pip/virtualenv toolchain. For example:

* Homebrew
* Various Linux distros
* Anaconda - Does that come with virtualenv now?

cguardia:
Thanks. I'll look into that.

cguardia:
@pydanny I can at least confirm that virtualenv is included in Anaconda. See this list:

http://docs.continuum.io/anaconda/pkg-docs

I'm looking for help with homebrew. Do you have any specific Linux distros in mind that could be a problem?

cguardia:
@pydanny homebrew users I know appear puzzled as to why you would be worried about virtualenv being a dependency. I myself do not understand why this would be different than depending on click, for example. Aside from this, any other comments?

mnazbro:
@cguardia Anaconda Python does have a virtualenv package, but it is not reliable. If you create a virtualenv from Anaconda, it will copy Python binaries from the system rather than use the ones from Anaconda.

When developing on a Mac, this doesn't usually come up because you already have the right Python binaries (2.7.10, etc.) from other Python versions installed. So the virtualenv *appears* to work. If you test this on a system which has an older Python (e.g. CentOS with Python 2.6), you'll see the error. You can get around this by setting `LD_LIBRARY_PATH`, but that is quite hacky.

The alternative to use is conda envs for anaconda Python since they will work even if you don't have the Python binaries on your system.

cguardia:
@mnazbro thanks a lot for the information.

However, note that this PR basically only makes virtualenv a required package. It's not used by cookiecutter at all, but only is made available for template authors that would like to use it. I think this makes the dependency acceptable, but I understand there are some risks.

pauleveritt:
@mnazbro Is it correct that, given the Anaconda situation, it is unlikely that CookieCutter will go in a direction of automating the virtual environment part, no matter the PR? If so, that's understandable, but still a drag.

cguardia:
Since Python 2.7 end of life is coming and recent Python 3 includes venv with pip, we feel that this feature is now not necessary. Thanks.

PR #555, opened 2015-10-17
Fix tests as per last changes in cookiecutter-pypackage

eliasdorneles:
So, this fixes the tests depending on the the default project name of cookiecutter-pypackage in master -- which changed in the last few commits.


hackebrot:
Thank you @eliasdorneles for spotting this! :+1:

PR #476, opened 2015-07-19
Add wagtail-cookiecutter-foundation to list of cookiecutters

chrisdev:

PR #566, opened 2015-10-26
Add wagtail-cookiecutter-foundation to list of Python-Django cookiecu…

Parbhat:
…tters


hackebrot:
Thank you @Parbhat and @chrisdev for fixing #476 :bow:

Issue #557, opened 2015-10-18
Consider moving off PyYAML to ruamel.yaml


pydanny:
PyYAML's current wheels do not appear to work for some Windows users. Specifically, they cause some Cookiecutter installations to fail. Also, there hasn't been a release endorsing use of Python 3.5. The PyYAML site hasn't seen an update in a long time.

On the other hand, https://pypi.python.org/pypi/ruamel.yaml is a fork/derivative of Kirill Simonov’s PyYAML. It appears to be working. We'll check on local Windows test machines to determine if it is a suitable replacement.

hackebrot:
Resolved via #558 :+1:

hackebrot:
Installation on Win 7 fails for [Cookiecutter 1.2.1](https://github.com/audreyr/cookiecutter/releases/tag/1.2.1) :worried:

hackebrot:
It tries to install ``ruamel.yaml>=0.10.12``. :confused: 

See https://ci.appveyor.com/project/hackebrot/pytest-cookies/build/1.0.9/job/0m3dw868u1a7s79w

Breaks on local Win 7 box too.

Is it because the wheel is **universal** and somehow doesn't get that it is installed on win?

pydanny:
This snippet was supposed to prevent PyYAML being installed on Windows machines installing it in a Python 2.7 environment: https://github.com/audreyr/cookiecutter/blob/master/setup.py#L38-L44

Clearly, it's working for some, but not all. Hand. Bang. Desk.

pydanny:
I think I understand now: Because Cookiecutter is installed as a `Wheel` in modern `pip`, I think it's already been compiled. So while it works great off source, it fails in formal releases. :cry:

hackebrot:
@audreyr @pydanny can you please explain why win7 requires special treatment? :bow: 

I set up **Cookiecutter v1.2.1** on another win box with x64 python and it worked just fine.

@ionelmc pointed out the according wheel is broken. I updated a issue in the [ruamel.yaml tracker](https://bitbucket.org/ruamel/yaml/issues/14/cant-install-ruamelyaml-package-under).

ionelmc:
@pydanny use extras_require instead: https://wheel.readthedocs.org/en/latest/index.html?highlight=marker#defining-conditional-dependencies

pydanny:
@ionelmc Thanks! We'll use that going forward

However, what we're facing is:

[...]

PR #567, opened 2015-10-26
Test x86 and x86-64 architectures on appveyor

maiksensi:
This PR will make sure cookiecutter is tested against x86 and x86-64 architecture interpreters on appveyor. The idea was to catch problems if a dependency fails to install on either architecture, like #557. 

This only changes testing on appveyor, as I am not sure whether this is possible on travis-ci without installing the corresponding interpreter from source and without changing ``tox.ini``. 
Maybe someone has some insight on this, I am not sure I can have a look at it anytime soon. 

Does this make sense, or is it better to leave the current testing as it is, to make sure testing is identical on both buildservers? 

Should documentation be updated?

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,19 +6,31 @@ environment:
     - PYTHON: "C:\\Python27"
       TOX_ENV: "py27"
 
+    - PYTHON: "C:\\Python27-x64"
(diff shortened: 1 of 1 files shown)


hackebrot:
LGTM :+1:

PR #568, opened 2015-10-26
Fix conditional dependencies for wheels in setup.py

hackebrot:
Related to #557

cc @ionelmc

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 import os
-import platform
 import sys
 
 try:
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -71,6 +62,14 @@
...
     include_package_data=True,
     install_requires=requirements,
+    extras_require={
+        ':sys_platform=="win32" and python_version=="2.7"': [
+            'PyYAML>=3.10'
+        ],
+        ':sys_platform!="win32" or python_version!="2.7"': [
+            'ruamel.yaml>=0.10.12'
+        ]
+    },
```
ionelmc: This doesn't install anything on *nix and py2.7. Intended?

hackebrot: Uhmhmm

```
':sys_platform!="win32" or python_version!="2.7"'
```
anything on *nix and py2.7 > True or False > True

ionelmc: Wat. You're right. :grin:

hackebrot: :stuck_out_tongue_closed_eyes:

hackebrot:
:bell:

hackebrot:
Rebased :triumph:

hackebrot:
Think we should get this in as long as we don't have a better solution to the whole YAML issue.

PR #360, opened 2014-12-30
Support better prompts

dnephin:
Resolves #30 

This change maintains backwards compatibility with the old format, so a context version isn't really necessary (although it wouldn't hurt if the context supported it).

Currently only supports one `type` of boolean, so that context parameters can be used in `if` blocks. Without this support you can't really have a default of `True`, and have the user enter `False` (because any value the user enters would be `True`.  I suspect there will  be other types that can be added here.

Issue #30, opened 2013-08-18
Support better prompts


audreyr:
Maybe we could support an enhanced `cookiecutter.json` file, like:

``` json
    [
        {
            "full_name": "Audrey Roy",
            "prompt": "Your full name?"
        },
        {
            "email": "audreyr@gmail.com",
            "prompt": "Your email address?"
        },
        {
            "github_username": "audreyr",
            "prompt": "Your username on github.com?"
        },
        {
            "project_name": "Python Boilerplate",
            "prompt": "Official name of your project, as used in your docs?"
        },
        {
            "repo_name": "boilerplate",
            "prompt": "Lowercase, spaceless, version of your repo name?",
            "help_text": "This should be a name that is importable as a Python package."
        },
        {
            "project_short_description": "Python Boilerplate contains all the boilerplate you need to create a Python package.",
            "prompt": "A 1-liner description of your project?"
        },
        {
            "release_date": "2013-08-11",
            "prompt": "Date that you plan to release your project?"
        },
        {
            "year": "2013",
            "prompt": "What year would you like to use in the copyright text?"
        },
        {
            "version": "0.1.0",
            "prompt": "Version number to start at?",
            "help_text": "We recommend using semantic versioning and starting at 0.1.0. See semver.org."
        }
    ]
```

Pawamoy:
:+1: (with khorn's format)

hackebrot:
Closing this in favor of #848

couling:
@hackebrot Would there be any chance of reconsidering that. Since it seems even in 2016 #848 was dead in the water. 

The justification there was that:
> [pydanny Nov 10, 2016](https://github.com/cookiecutter/cookiecutter/pull/848#issuecomment-259575460)
> This is a huge, HUGE change to Cookiecutter. It's a huge amount of work for the maintainers, forcing a 2.0 release. This is the easy part, but the real work is the ongoing maintenance burden.

That was a totally reasonable statement... but if #848 was too hard in 2017 and now in 2023 there's still no sponsorship for it, would a PR implementing this #30 be likely to get accepted?

I note that this could be done with a non-breaking change by making the key/value pairs of the config accept string or dict interchangeably:

```json
{
    "name": {
        "default": "Audrey Roy",
        "prompt": "Your FULL name"
    },
    "email": "audreyr@gmail.com"
}
```

PR #572, opened 2015-10-27
Allow to define custom prompts in cookiecutter.json

goerz:
This is an extremely lightweight implementation of custom prompts. It uses an extra `_prompt` key in the json data structure that collects all custom prompts. For example,

```
{
    "full_name": "Audrey Roy",
    "email": "audreyr@gmail.com",
    "project_name": "Complexity",
    "_prompt": {
        "full_name": "your full name",
        "email": "email address",
    }
}
```

Keys that do not define a custom prompt use the default (i.e., the keyname).

The implementation in this patch is an alternative solution to issue #30 than the one provided by pull requests #360 #513, or #173. Its benefit is that it requires minimal change to the code, and is fully compatible with the existing json data structure.

If there is interest in merging this request in lieu of more fundamental rewrites of the `cookiecutter.json` file format, I could add further documentation and testing.

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -17,15 +17,15 @@ from future.utils import iteritems
 from jinja2.environment import Environment
 
 
-def read_user_variable(var_name, default_value):
-    """Prompt the user for the given variable and return the entered value
+def read_user_variable(text, default_value):
+    """Prompt the user with the given text and return the entered value
     or the given default.
 
-    :param str var_name: Variable of the context to query the user
+    :param str text: Text to query the user with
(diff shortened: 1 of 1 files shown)


Klortho:
I had updated this PR, merging in the latest master -- there were merge conflicts, and opened up a new PR, #644. I thought that might save somebody some trouble. I do hope this feature makes it into the next release.

hackebrot:
Thank you for this PR @goerz! 🙇 

I think it would be really cool to be able to customize prompts, but I'm hesitant to add more *private* variables to **cookiecutter.json**. Instead I would like to focus on #848.

PR #573, opened 2015-10-30
Handle trailing slash on url. fix #546.

### LINKED ISSUE #546: Problem with --checkout reclone
The message should ask me about recloning `/Users/audreyr/.cookiecutters/cookiecutter-pypackage`, not `/Users/audreyr/.cookiecutters`.

```
$ cookiecutter https://github.com/eliasdorneles/cookiecutter-pypackage/ -c adding-travis-setup-for-pypi-deployment
You've cloned /Users/audreyr/.cookiecutters before. Is it okay to delete and re-clone it? [yes]: 
```

ramiroluz:
Quick fix to handle an extra trailing slash on the template url.

Fix issue #546.

### DIFF
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -103,6 +103,8 @@ def clone(repo_url, checkout=None, clone_to_dir=".", no_input=False):
         msg = "'{0}' is not installed.".format(repo_type)
         raise VCSNotInstalled(msg)
 
+    if repo_url.endswith('/'):
+        repo_url = repo_url[:-1]
     tail = os.path.split(repo_url)[1]
     if repo_type == 'git':
         repo_dir = os.path.normpath(os.path.join(clone_to_dir,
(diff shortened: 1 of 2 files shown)


cookiecutter/vcs.py
```diff
@@ -103,6 +103,8 @@ def clone(repo_url, checkout=None, clone_to_dir=".", no_input=False):
         msg = "'{0}' is not installed.".format(repo_type)
         raise VCSNotInstalled(msg)
 
+    if repo_url.endswith('/'):
+        repo_url = repo_url[:-1]
```
hackebrot: Think we can also use ``repo_url.strip('/')``.

However I am not sure if this is valid for all of the possible repo urls.. :worried:

ramiroluz: There are any repo_url starting with a slash? Like: repo_url = '/home/eliasdorneles/cookiecutter-pypackage/'

If so, it would be better to use rstrip(). I will try it and run the tests.

ramiroluz: Yep, rstrip passed all tests.

PR #574, opened 2015-10-31
Revert skipif markers to use correct reasons (bug fixed in pytest)

hackebrot:
Pytest bug fixed by @nicoddemus via https://github.com/pytest-dev/pytest/pull/865


hackebrot:
Rebased :triumph:

PR #575, opened 2015-11-02
Drop yaml libs use custom parser

hackebrot:
WIP :raised_hand:

Simple regex parser implementation for [user configs](https://cookiecutter.readthedocs.org/en/latest/advanced_usage.html#user-config-0-7-0) to replace **PyYAML AND ruamel.yaml**. Users on various platforms and Python versions seem to encounter various problems with both libs when installing **Cookiecutter**, which is a huge problem (see #557 :worried: ).

Please let me know your thoughts!

(Related to #568)

### DIFF
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -14,13 +14,9 @@ import logging
 import os
 import io
 
-try:
-    import ruamel.yaml as yaml
-except ImportError:
-    import yaml
-
 from .exceptions import ConfigDoesNotExistException
 from .exceptions import InvalidConfiguration
+from . import config_parser
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


hackebrot:
**Known issue:** Bad user configs do not cause an ``InvalidConfiguration`` error.

Won't update the tests until we are on the same page regarding the implementation. :point_left:

pydanny:
Hooray to saying goodbye to the overkill of heavy-duty formal YAML libraries! That said, `config_parser` is really close to being `ConfigParser`. How about something like `simple_yaml` instead? That is a more obvious name, and lends itself to being separated out into a different, pure Python YAML parser project. :smile:

hackebrot:
Note: See gitter

audreyr:
@hackebrot Have you considered making this a separate library that we then include as a dependency?

hackebrot:
Closing in favor of #621

PR #588, opened 2015-11-11
Allow setting parameters with no defaults

### LINKED ISSUE #587: No way to define options that have no defaults
Currently if you set a value in `cookiecutter.json` to `null` it becomes `None` and is then turned into the _string_ `'None'`.

johtso:
Fixes #587

### DIFF
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -41,6 +41,8 @@ def test_convert_to_str(mocker, raw_var, rendered_var):
         if not isinstance(raw_var, basestring):
             raw_var = str(raw_var)
         from_string.assert_called_once_with(raw_var)
+    else:
+        assert not from_string.called
 
 
 @pytest.fixture(autouse=True)
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -22,7 +22,8 @@ from jinja2.environment import Environment
     (1, '1'),
     (True, 'True'),
     ('foo', 'foo'),
-    ('{{cookiecutter.project}}', 'foobar')
+    ('{{cookiecutter.project}}', 'foobar'),
+    (None, None),
 ])
 def test_convert_to_str(mocker, raw_var, rendered_var):
     env = Environment()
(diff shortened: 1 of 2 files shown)


pydanny:
@johtso, thank you for your pull request. :cookie: :sunglasses: 

Could you add a test for the condition? While this is technically covered by our test coverage (https://codecov.io/github/audreyr/cookiecutter/cookiecutter/prompt.py?ref=bc3cb0ed034bde485d963ee9b50217cbd29ee069), that's a false positive.

johtso:
@pydanny will do!

hackebrot:
Hi @johtso! :wave: 

Could you please explain why you would have a variable with a default that is ``null`` in your template? An empty string should work perfectly fine if you do not want to specify a particular string value.

tests/test_prompt.py
```diff
@@ -35,10 +36,11 @@ def test_convert_to_str(mocker, raw_var, rendered_var):
     result = prompt.render_variable(env, raw_var, context)
     assert result == rendered_var
 
-    # Make sure that non str variables are conerted beforehand
-    if not isinstance(raw_var, basestring):
-        raw_var = str(raw_var)
-    from_string.assert_called_once_with(raw_var)
+    # Make sure that non None non str variables are conerted beforehand
+    if raw_var is not None:
```
hackebrot: This counteracts the extra test parameter! :worried:

johtso: Right, because the assertion doesn't apply anymore (there's still the assertion above checking the return value).

I could assert that it's not called I suppose..

johtso:
@hackebrot the idea was that sometimes there is no sensible default for a value and you want to require that the user supplies one.

Even if that's not a particularly useful use-case, it's definitely less surprising than converting `null` into `'None'`.

hackebrot:
Hey @johtso can you please rebase this guy? We sorted out the issues with AppVeyor and I'd like to see the test results for windows. Thank you :bow:

johtso:
@hackebrot done :smile_cat:

tests/test_prompt.py
```diff
@@ -35,10 +36,13 @@ def test_convert_to_str(mocker, raw_var, rendered_var):
     result = prompt.render_variable(env, raw_var, context)
     assert result == rendered_var
 
-    # Make sure that non str variables are conerted beforehand
-    if not isinstance(raw_var, basestring):
-        raw_var = str(raw_var)
-    from_string.assert_called_once_with(raw_var)
+    # Make sure that non None non str variables are conerted beforehand
```
hackebrot: This sentence seems wrong :grin:

johtso: Any suggestions welcome! Looks like I kept a typo too..

pydanny:
Grats on your first pull request here @johtso. Welcome aboard! :ship: :train:

johtso:
Cheers @pydanny

PR #595, opened 2015-11-16
Fix the repo regex to handle + in URLs

habnabit:
The documentation said that you could do git+https:// or git+ssh:// but that was not, in fact, true.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -30,19 +30,18 @@ builtin_abbreviations = {
     'bb': 'https://bitbucket.org/{0}',
 }
 
-REPO_REGEX = """
-(
-((git|ssh|https|http):(//)?)    # something like git:// ssh:// etc.
- |                              # or
- (\w+@[\w\.]+)                  # something like user@...
+REPO_REGEX = re.compile(r"""
+(?x)
+((((git|hg)\+)?(git|ssh|https?):(//)?)  # something like git:// ssh:// etc.
+ |                                      # or
+ (\w+@[\w\.]+)                          # something like user@...
 )
-.*
-"""
+""")
 
 
 def is_repo_url(value):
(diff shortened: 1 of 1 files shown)


pydanny:
@habnabit We've tweaked the @appveyor integration. Can you bump this pull request so @appveyor runs again?

habnabit:
@pydanny what is involved in a bump? I can delete and recreate the branch, but I don't know how github will take that. Or push an empty commit?

pydanny:
@habnabit - That's where you make a tiny change to get CI to run again. No worries, I finally figured out how to restart it. :smile:

pydanny:
Still breaking in @appveyor. Trying to debug...

habnabit:
Anything I can do to help out with this?

hackebrot:
Is the regex still compiled with ``VERBOSE`` on?

storax:
`re.compile` takes flags as second argument. So in this case it is not compiled with `VERBOSE` on and needs `re.VERBOSE` as second argument. See [Python Regular expression Documentation](https://docs.python.org/2/library/re.html#re.VERBOSE).

habnabit:
@hackebrot yes. I switched it to use `(?x)` instead, which is the inline flag for `VERBOSE`.

storax:
Ah learned something new :+1:

hackebrot:
Let's stick to ``re.VERBOSE`` :grimacing: Comprehensibility is key IMHO

btw Happy New Year!

habnabit:
Will update soon.

hackebrot:
:bell:

hackebrot:
:bell:

michaeljoseph:
@habnabit :bell:
Also, could you please add a test case to https://github.com/audreyr/cookiecutter/blob/master/tests/test_main.py?

Issue #571, opened 2015-10-27
extra_context stopped working for me in 1.2.1


xsteadfastx:
Im sure im missing a change but i cant find what its wrong with this snippet im using:

```
from cookiecutter.main import cookiecutter
from datetime import datetime
from tzlocal import get_localzone
import pytz


cookiecutter(
    'cookiecutter-new-post',
    overwrite_if_exists=True,
    extra_context={
        'date': datetime.utcnow().replace(tzinfo=pytz.utc).astimezone(get_localzone()).strftime('%Y-%m-%d %H:%M')})
```
It doesnt create `{{ cookiecutter.date }}`. This works fine with version 1.1.0. Maybe someone has a hint for me. And by the way: thanks for this beautiful piece of software. I use it everyday and it makes everything so much easier :smile:

hackebrot:
Hi @xsteadfastx! Thank you for your feedback :+1:

Happy to hear that you like **Cookiecutter** :smile:

You should be fine by adding ``date`` to your ``cookiecutter.json``. I *broke* this *feature* in 8f08298 in the course of #542. The rationale of this change is that template authors should declare all the used variable in the context file... :no_mouth: 

HTH

xsteadfastx:
this did the trick :) thanks alot

hackebrot:
You're welcome! :smiley_cat:

danielsamuels:
Hey,

I'm just wondering what the rationale was behind not adding values to the context if they're not defined in the JSON file? 

We're calling `cookiecutter()` using the API and passing in a dict of values that we need in our hooks & templates, but now we have to skip past a whole bunch of values in the setup script, even if we know that they will never be defined during initialisation, nor will they ever be changed.  

An example of how this applies to us is that we generate a Github project via the API before we create the project, we then pass the clone URL into the extra context so we can use it within a post_gen hook to set the generated project's remote. This is a value that clearly will never be user-defined, but is still 100% necessary to be passed into the Jinja2 context, but now doesn't work due to the aforementioned change.

Is there any way to get this working again without having to downgrade our Cookiecutter version & without making it a key in the cookiecutter.json file?

Thanks,
Daniel.

pydanny:
@danielsamuels My deepest regrets, but I'm unable to wrap my head around the double negatives in the first sentence. Since the rest of your question seems to build on it, I want to make certain I understand it before answering. Could you rephrase your question?

danielsamuels:
Hi @pydanny,

[...]

PR #597, opened 2015-11-18
Include unknown extra variables

hackebrot:
Address the **feature/bug** discussed in #571.

----

Variables which are not specified in **cookiecutter.json** are added to the context unless the according flag (``known_only=True``) is set for ``generate.apply_overwrites_to_context()``.

The rationale of 8f082981fdcb3d1138d2962de88622006070ae91 was to **not** prompt the user for variables which are not declared in **cookiecutter.json** but injected via default context, as the template most probably doesn't use them.

This is different for extra context though, as it needs to be specified on every single invocation of cookiecutter.

Feedback is greatly appreciated! Thanks :bow: 

cc @danielsamuels @pydanny

----

Imagine the following **cookiecutter.json**:

```json
{
    "github_username": "hackebrot",
    "plugin_name": "foobar"
}
```

A user config file **~/.cookiecutterrc**:

```
default_context:
    github_username: "pydanny"
    foo: "bar"
```
Now if we specify extra context using the Python API:

```python
extra_context = {
    'unknown_in_extra': 'helloworld'
}
```

The context will look as follows (before user cli overwrites):

```json
{
    "github_username": "pydanny",
    "plugin_name": "foobar",
    "unknown_in_extra": "helloworld"
}
```
Please note that the user will be prompted for ``unknown_in_extra``!

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -56,7 +56,9 @@ def apply_overwrites_to_context(context, overwrite_context):
     """Modify the given context in place based on the overwrite_context."""
     for variable, overwrite in overwrite_context.items():
         if variable not in context:
-            # Do not include variables which are not used in the template
+            # Albeit not relevant for the template, set this variable to
+            # resemble ``dict.update()``
+            context[variable] = overwrite
             continue
 
         context_value = context[variable]
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -112,9 +112,9 @@ def generate_context(context_file='cookiecutter.json', default_context=None,
     # Overwrite context variable defaults with the default context from the
     # user's global config, if available
     if default_context:
-        apply_overwrites_to_context(obj, default_context)
+        apply_overwrites_to_context(obj, default_context, known_only=True)
     if extra_context:
-        apply_overwrites_to_context(obj, extra_context)
+        apply_overwrites_to_context(obj, extra_context, known_only=False)
 
     logging.debug('Context generated is {0}'.format(context))
     return context
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -56,8 +56,8 @@ def apply_overwrites_to_context(context, overwrite_context):
     """Modify the given context in place based on the overwrite_context."""
     for variable, overwrite in overwrite_context.items():
         if variable not in context:
-            # Albeit not relevant for the template, set this variable to
-            # resemble ``dict.update()``
+            # Add a new entry to the context dict for values that are not
+            # declared in context but required in the template itself
             context[variable] = overwrite
             continue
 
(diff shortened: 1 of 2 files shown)


danielsamuels:
As mentioned in #571, one of the values we're passing through using the `cookiecutter()` call is the Git clone URL, we will never need the user to enter this value via the initial prompt so for us it wouldn't make sense for it to be prompted for. I don't know if the codebase currently supports this, but is it possible to have it in the Jinja2 context, but _not_ have it prompted for in the initial question / answer flow?

hackebrot:
Nope, it does not (which lead to 8f082981fdcb3d1138d2962de88622006070ae91) :no_mouth: This is not a regression though!

Users are prompted for the context, which is a simple dict that doesn't hold any meta info on whether the variables originates from **cookiecutter.json**, **default_context** or **extra_context**.

Please see https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/main.py#L131

danielsamuels:
Hmm, I think it would be possible, but perhaps not in a clean way. You could potentially pass `extra_context` into that `prompt_for_config` call and have it check the keys within [this loop](https://github.com/audreyr/cookiecutter/blob/6f6cd90c8114b293d38a5177f512d0e341b8c848/cookiecutter/prompt.py#L115). Would that work at all?

hackebrot:
I think you could rather try to prepend your *private* extra variables with an underscore ``_``.

hackebrot:
I think you could rather try to prepend your *private* extra variables with an underscore ``_``.

hackebrot:
:rage: appveyor fails due to git-connectivity problems again...

danielsamuels:
That's probably the best option actually. In that case, I'm :+1:

hackebrot:
Hmm I'm still not quite sure if it's any good to use template variables that are not declared in **cookiecutter.json**. At the end of the day it is some kind of sanity check they you don't accidentally forget to set a reasonable default. Cookiecutter renders empty strings for unknown variables.

danielsamuels:
It's been a while since there's been any movement on this, and a discussion doesn't seem to have occurred. Would it be possible to get this merged, or an alternative implemented? This issue is unfortunately blocking a couple of our internal projects from working, so it would great to see it resolved. Is there anything I can do to help?

hackebrot:
Rebased :smirk_cat: 

@danielsamuels have you tried to add your variables to **cookiecutter.json** with an ``_`` prefix?

danielsamuels:
I haven't. If that's a solution that already works, then great, I'll do that. Unfortunately I haven't touched the project I was working on for a little while (I've had to work on other things). I'll give it a try at some point soon.

audreyr:
I tried to review this, but the logic of when context variables override each other is too confusing for me (or @pydanny) to follow. I'm nervous that this PR will make it even more complex.

Before accepting this, we need to at least have a document explaining the latest context override order. (Is there one that I missed?)

hackebrot:
By now I think we don't need this. Happy to re-open though. :bow:

Issue #599, opened 2015-11-22
Tests that require git break on appveyor


hackebrot:
See #597, #598 :worried:

hackebrot:
cc @maiksensi

luzfcb:
@hackebrot this problem was expected?: https://ci.appveyor.com/project/audreyr/cookiecutter/build/1.0.806/job/mv3x1v5rjvi87kvp#L150

seems a bug in appveyor and mingw64 installation.

hackebrot:
Nope :worried:

maiksensi:
As I am not at home, I only had a brief look at it. You could try changing GIT's config by adding `http.sslverify false`. And try to verify whether this file exists `C:/Program Files/Git/mingw64/libexec/ssl/certs/ca-bundle.crt` (`Get-ChildItem $DIRECTORY -force`). I am confident it is something with the `PATH` to the CABundle, depending on whether a new mingw64 was installed lately. I'll try to create a minimal example tomorrow and see if I can reproduce this.

PR #603, opened 2015-11-23
Remove external deps from tests

### LINKED ISSUE #434: test relies to external repository
I'm not sure having extensive test actually needing to use web connection is a good idea. I'm currently in China, and connections towards western world are not so easy. Even with a VPN, I can't launch the test suite without having several test failing while cloning (git clone or hg clone).

I also noticed that travis as difficulties itself to clone all these (https://travis-ci.org/0k/cookiecutter/builds/56044642). (along with strange travis issues not finding ``py33`` or ``flake8``) 

And any bad push to one of these external repository will have the consequence to break tests from cookiecutter.
At last, having these external dependency raises the whole test time, which is never a good thing as it is an incitation to run tests less often.

Basically, there probably shouldn' t be even 1 test having to rely to external sites, you could test git and mercurial locally, and rely on fixed version of some reference templates.

My 2 cents.

hackebrot:
This PR rewrites any test that relies on network access and as such is likely to fail due to outdated reference templates or connectivity issues (see #599).

Please note that the new tests **DO** use mocks.

I would appreciate a thorough review and look forward to your feedback.Thanks :bow: 

Resolve #434


hackebrot:
cc @vaab @michaeljoseph

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
 
     expanded_template = expand_abbreviations(template, {})
     assert is_repo_url(expanded_template) is True
+
+
+@pytest.fixture(scope='session')
+def user_dir(tmpdir_factory):
```
audreyr: Add docstring:
`"""Fixture that simulates the user's home directory"""`

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+    return tmpdir_factory.mktemp('user_dir')
+
+
+@pytest.fixture(scope='session')
+def user_config_data(user_dir):
```
audreyr: Add docstring:
```
"""Fixture that creates 2 Cookiecutter user config dirs in the user's home directory.
* `cookiecutters`
* `replay_str`
:return: Dict with name of both user config dirs and empty default context
"""
```

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+
+    return {
+        'cookiecutters_dir': str(cookiecutters_dir),
+        'replay_dir': str(replay_dir),
+        'default_context': {}
+    }
+
+
+@pytest.fixture(scope='session')
+def user_config_file(user_dir, user_config_data):
```
audreyr: Add docstring:
```
"""Fixture that creates a config file called `config` in the user's home directory, with JSON from `user_config_data`.
:param user_dir: Simulated user's home directory
:param user_config_data: Dict of config values
:returns: String of path to config file
```

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+    config_file = user_dir.join('config')
+    config_file.write(json.dumps(user_config_data))
+    return str(config_file)
+
+
+@pytest.fixture
+def template_url():
```
audreyr: Add docstring:
```
"""URL to example Cookiecutter template on GitHub. Note: when used, git clone is mocked"""
```

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+    return 'https://github.com/pytest-dev/cookiecutter-pytest-plugin.git'
+
+
+@pytest.fixture
+def output_dir(tmpdir):
```
audreyr: Add docstring:
```
"""Given a temporary dir, create an `output` subdirectory inside it and return its path."""
```

audreyr:
@hackebrot I hope I'm not overwhelming you with the docstring comments. I'm just documenting as I go along so that I can make sure I understand each part.

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+    return tmpdir.mkdir('output')
+
+
+def test_cookiecutter_repository_url_should_clone(
```
audreyr: Add docstring:
```
"""`clone()` should be called with correct args when `cookiecutter()` is called."""
```

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+    )
+
+    mock_clone.assert_called_once_with(
+        repo_url=template_url,
+        checkout=None,
+        clone_to_dir=user_config_data['cookiecutters_dir'],
+        no_input=True
+    )
+
+    assert os.path.isdir(project_dir)
```
audreyr: This test is a great start. If we remove the integration tests, I suppose we ought to unit test the `cookiecutter()` function more thoroughly. 

To address this, I noticed that there are already other unit tests covering lines of `cookiecutter()` in cookiecutter/tests/replay/test_replay.py. Perhaps we can move them here to start, and then see what lines still need tests along the lines of the one you've written here.

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
 def test_identify_raise_on_unknown_repo(unknown_repo_type_url):
     with pytest.raises(exceptions.UnknownRepoType):
         vcs.identify_repo(unknown_repo_type_url)
+
+
+def test_prompt_should_ask_and_rm_repo_dir(mocker, tmpdir):
```
audreyr: Add docstring:
```
"""In `prompt()`, the user should be asked whether it's okay to delete and reclone the repo."""
```

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+    )
+    repo_dir = tmpdir.mkdir('repo')
+
+    vcs.prompt_and_delete_repo(str(repo_dir))
+
+    assert mock_read_user.called
+    assert not repo_dir.exists()
+
+
+def test_prompt_should_ask_and_keep_repo_dir(mocker, tmpdir):
```
audreyr: Add docstring:
```
"""In `prompt()`, if the user wants to keep their old cloned template repo, it should not be deleted."""
```

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+    repo_dir = tmpdir.mkdir('repo')
+
+    with pytest.raises(SystemExit):
+        vcs.prompt_and_delete_repo(str(repo_dir))
+
+    assert mock_read_user.called
+    assert repo_dir.exists()
+
+
+def test_prompt_should_not_ask_if_no_input_and_rm_repo_dir(mocker, tmpdir):
```
audreyr: Add docstring:
```
"""In `prompt_and_delete_repo()`, if `no_input` is True, the call to `vcs.read_user_yes_no()` should be suppressed."""

File: tests/test_vcs.py [hunk 11]
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+    mock_read_user = mocker.patch(
+        'cookiecutter.vcs.read_user_yes_no',
```
audreyr: Seems like it would be clearer to patch `cookiecutter.prompt.read_user_yes_no`, rather than the imported version of the function. (Unless this is following some pytest/patch best practice that I don't know about?)

tests/test_main.py
```diff
@@ -27,3 +33,62 @@ def test_expand_abbreviations():
...
+
+@pytest.fixture
+def output_dir(tmpdir):
+    return tmpdir.mkdir('output')
+
+
+def test_cookiecutter_repository_url_should_clone(
+        mocker, template_url, output_dir, user_config_file, user_config_data):
+    mock_clone = mocker.patch(
+        'cookiecutter.main.clone',
```
audreyr: It seems like it would be clearer to patch `cookiecutter.vcs.clone` rather than the imported copy of it here.

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+
+
+@pytest.fixture
+def clone_dir(tmpdir):
+    return str(tmpdir.mkdir('clone_dir'))
+
+
+def test_clone_should_raise_if_vcs_not_installed(mocker, clone_dir):
+    mocker.patch(
+        'cookiecutter.vcs.which',
```
audreyr: Patch `whichcraft.which` here (I think)

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+    repo_dir = tmpdir.mkdir('repo')
+
+    vcs.prompt_and_delete_repo(str(repo_dir), no_input=True)
+
+    assert not mock_read_user.called
+    assert not repo_dir.exists()
+
+
+@pytest.fixture
+def clone_dir(tmpdir):
```
audreyr: Add docstring:
"""Simulates creation of a directory called `clone_dir` inside of `tmpdir`."""

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+    return str(tmpdir.mkdir('clone_dir'))
+
+
+def test_clone_should_raise_if_vcs_not_installed(mocker, clone_dir):
```
audreyr: Add docstring:
```
"""In `clone()`, a `VCSNotInstalled` exception should be raised if no VCS is installed."""
```

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+
+    with pytest.raises(exceptions.VCSNotInstalled):
+        vcs.clone(repo_url, clone_to_dir=clone_dir)
+
+
+@pytest.mark.parametrize('repo_type, repo_url, repo_name', [
+    ('git', 'https://github.com/hello/world.git', 'world'),
+    ('hg', 'https://bitbucket.org/foo/bar', 'bar'),
+])
+def test_clone_should_invoke_git(
```
audreyr: Add docstring:
```
"""When `clone()` is called with a git/hg repo, the corresponding VCS command should be run via `subprocess.check_call()`.
This should take place:
* In the correct dir
* With the correct args."""
```

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+
+
+def test_clone_should_abort_if_user_does_not_want_to_reclone(mocker, tmpdir):
+    mocker.patch(
+        'cookiecutter.vcs.is_vcs_installed',
+        autospec=True,
+        return_value=True
+    )
+    mocker.patch(
+        'cookiecutter.vcs.read_user_yes_no',
```
audreyr: To make this more of a unitlike unit test, patch `cookiecutter.vcs.prompt_and_delete_repo` instead.

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+
+    mock_subprocess.assert_any_call(
+        [repo_type, 'clone', repo_url], cwd=clone_dir
+    )
+    mock_subprocess.assert_any_call(
+        [repo_type, 'checkout', branch], cwd=expected_repo_dir
+    )
+
+
+def test_clone_should_abort_if_user_does_not_want_to_reclone(mocker, tmpdir):
```
audreyr: """In `clone()`, if user doesn't want to reclone, Cookiecutter should exit."""

tests/test_vcs.py
```diff
@@ -143,3 +50,152 @@ def unknown_repo_type_url(request):
...
+    clone_to_dir.mkdir('cookiecutter-pytest-plugin')
+
+    repo_url = 'https://github.com/pytest-dev/cookiecutter-pytest-plugin.git'
+
+    with pytest.raises(SystemExit):
+        vcs.clone(repo_url, clone_to_dir=str(clone_to_dir))
+    assert not mock_subprocess.called
+
+
+def test_clone_should_rstrip_trailing_slash_in_repo_url(mocker, clone_dir):
```
audreyr: Add docstring:
```
"""In `clone()`, repo URL's trailing slash should be stripped if one is present."""

**Comment by audreyr** [comment]
Really great work @hackebrot. Once you add the docstrings and fixes, I'll recheck and merge. Feel free to modify the docstrings if you think something else would sound better.

**Comment by hackebrot** [comment]
Thank you for the thorough review @audreyr :+1: 

It'll take a bit for me to get back to this. I'll keep you posted! :bow:

**Comment by audreyr** [comment]
Take your time, there's no rush :sunglasses:

**Comment by hackebrot** [comment]
Doc strings added :joy_cat:

File: tests/test_vcs.py (hunk 17 above)
Comment (hackebrot) [inline tests/test_vcs.py]: Done

File: tests/test_vcs.py (hunk 13 above)
Comment (hackebrot) [inline tests/test_vcs.py]: I'm afraid we need to keep the patch as mock requires the string to match the **actual used** call, if that makes sense :speak_no_evil:

File: tests/test_main.py (hunk 12 above)
Comment (hackebrot) [inline tests/test_main.py]: See below

File: tests/test_vcs.py (hunk 11 above)
Comment (hackebrot) [inline tests/test_vcs.py]: See below

**Comment by hackebrot** [comment]
Sorry that it took me this long to get back to this @audreyr

File: tests/test_vcs.py (hunk 13 above)
Comment (michaeljoseph) [inline tests/test_vcs.py]: Because I had to look it up: http://www.voidspace.org.uk/python/mock/patch.html#where-to-patch

Comment (hackebrot) [inline tests/test_vcs.py]: thanks for the reference @michaeljoseph :bow:

**Comment by hackebrot** [comment]
Any objections @audreyr @pydanny @michaeljoseph?

**Comment by hackebrot** [comment]
appveyor seems to have problems, PR is still in queue after 2 hrs. :disappointed:

**Comment by hackebrot** [comment]
It's fine now. What do you think?

File: tests/test_vcs.py [hunk 20]
```diff
@@ -186,3 +186,17 @@ def test_prompt_should_not_ask_if_no_input_and_rm_repo_dir(mocker, tmpdir):
...
     assert not mock_read_user.called
     assert not repo_dir.exists()
+
+
+def test_clone_should_raise_if_vcs_not_installed(mocker, tmpdir):
+    mocker.patch(
+        'cookiecutter.vcs.which',
+        autospec=True,
+        return_value=''
+    )
```
michaeljoseph: I think this should be mocking `vcs.is_vcs_installed` rather than `witchcraft.which`?
And then add a `test_is_vcs_installed` that mocks `cookiecutter.vcs.which`?
This only raised my interest because `witchcraft.which` returns `None` if it can't find the executable, so we should probably change `cookiecutter.vcs.which` to test for `None` also?

michaeljoseph:
@hackebrot Looks good :smiley_cat: 
Happy for my comment to be resolved later.
:sparkles: Faster Tests! :fireworks:

tests/test_vcs.py
hackebrot: On it :smiley:

hackebrot: ``bool(None)`` is ``False`` just as ``bool("")``. I can parametrize the test the test though.

hackebrot:
Comments addressed @michaeljoseph!

Please feel free to merge, once CI is finished :bow:

hackebrot:
Looking good :grin:

hackebrot:
Thank you @michaeljoseph :bow:

PR #608, opened 2015-11-27
Use open context manager to read context_file in generate

### LINKED ISSUE #607: file_handle.close() is never called -- and it causes a bug!!
In https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/generate.py#L90, `file_handle.close()` is never called, and there's no need for it to remain open.

This is the first time in over 10 years of python programming that I've run into an actual issue with file handles left open, so I felt that the use of two exclamation points in the subject was warranted. I'm removing a temporary template after an unrelated error, and...

      File "D:\anaconda32\lib\shutil.py", line 250, in rmtree
        os.remove(fullname)
    WindowsError: [Error 32] The process cannot access the file because it is being
    used by another process: '.\\tmpp2duu1\\cookiecutter.json'

This change in generate.py:90 helps python not stumble on this particular issue:

    try:
        with open(context_file) as file_handle:
            obj = json.load(file_handle, object_pairs_hook=OrderedDict)
    except ValueError as e:
        # ...

hackebrot:
Fix #607

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -87,9 +87,9 @@ def generate_context(context_file='cookiecutter.json', default_context=None,
 
     context = {}
 
-    file_handle = open(context_file)
     try:
-        obj = json.load(file_handle, object_pairs_hook=OrderedDict)
+        with open(context_file) as file_handle:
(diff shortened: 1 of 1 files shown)


audreyr:
:+1:

PR #612, opened 2015-12-03
Fix AUTHORS typo

johtso:


hackebrot:
Sorry about that. Good catch :+1:

PR #616, opened 2015-12-10
Add cookiecutter-python to the list of templates

flyondrag:
I have done with the new template for the Cookiecutter:

It comes with the number of useful features (implemented via Makefile facility):

* Management of Python virtual environments within the project:
  - Set up which versions of Python interpreter will be used for the project
  - Set up how a virtual environments will be created
  - Prepare an environments according to requirement files

* Support of the development lifecycle:
  - Work with sources in "development mode"
  - Test the project with Pytest
  - Build the distribution package (either in wheel of tarball formats)
  - Clean the project's directory in a Makefile-flavour::

* Issuing of the project's documentation using Sphinx:
  - Render documentation as a single PDF document

* Debian packaging
  - Configure most valuable parameters for a Debian packaging system via Makefile variables
  - Build a Debian package containing an isolated virtual environment with installed project and all of the project's requirements
  - Build a Debian package containing a prefix-based installation of the
project

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -152,6 +152,9 @@ Python
 ~~~~~~
 * `cookiecutter-pypackage`_: `@audreyr`_'s ultimate Python package project
   template.
+* `cookiecutter-python`_: The boilerplate Python project that aims to create
+  facility for maintaining of the package easily. It considering tools for
+  building, testing and distribution.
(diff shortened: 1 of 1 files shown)


audreyr:
Nice work @flyondrag.

Issue #615, opened 2015-12-10
Revert appveyor architecture tests


hackebrot:
Since tox does not distinguish between different architectures by default (we figured that out just recently in a different project), I feel it makes sense to reduce our CI times... :worried: 

Tox has its own way of determining the python interpreter, which means even if you create the virtual environment with a x64 version, tox uses x86 on appveyor internally.

Let's revert #567 

cc @maiksensi

maiksensi:
Actually, I would rather add another small patch to the current config to make sure that tox uses the correct interpreter internally. The lookup mechanism (on Windows at least) is to look at the `PATH` and if there is a python interpreter, it uses that one. So in order to use the correct interpreter, we would need to set the `PATH` variable in `appveyor.yml`. 

I would like to test this before reverting the changes. 
--> todo would be: 
* look up how tox determines the interpreter
* create a test repo with appveyor support and test the interpreter with a test

hackebrot:
Cool, whatever works best :+1: 

As of right now, I'd rather not wait for appveyor to finish if the tests are redundant.

maiksensi:
then revert the changes and I'll put up a new PR when I get to it.

insspb:
We are dropping appveyor. Closing issue.

PR #620, opened 2015-12-22
set path before running tox

maiksensi:
This should fix the problem described in #615.
The default interpreter should be the one used to run Tox (https://tox.readthedocs.org/en/latest/config.html?highlight=interpreter#confval-basepython=NAME-OR-PATH) but to be sure, set PATH to contain the correct interpreter. On Windows Tox also searches there (https://github.com/msabramo/tox/blob/master/tox/interpreters.py#L121).

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -29,6 +29,7 @@ environment:
 
 
 init:
+  - set PATH=%PYTHON%;%PYTHON%\Scripts;%PATH%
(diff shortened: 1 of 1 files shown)


maiksensi:
PS: We would remove 32 Bit Interpreters. Using 32 Bit on Travis and 64 Bit on Appveyor should be sufficient for cookiecutter I guess or even completely drop 64 Bit testing to speed tests up!

hackebrot:
LGTM :+1:

Issue #569, opened 2015-10-27
pip install cookiecutter without python-dev fails


SweetAndLow:
OS: Ubuntu 14.04
Version: cookiecutter-1.2.1-py2.py3-none-any

### Problem
Simply installing the package doesn't work and required some extra effort to get it working.
It would be great if the error was easier to figure out or it handled the dependency better.

### Repro:
New system, install python-pip, pip install cookiecutter 

### Output: 
`Command /usr/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip_build_root/ruamel.yaml/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-zWVNPC-record/install-record.txt --single-version-externally-managed --compile failed with error code 1 in /tmp/pip_build_root/ruamel.yaml
Storing debug log for failure in /home/svetter/.pip/pip.log`

### Workaround
install python-dev on the ubuntu system and the errors is fixed.

hackebrot:
Hi @SweetAndLow, is there any useful information in ``~/.pip/pip.log``? :worried:

SweetAndLow:
Can't attach logs because you need write permissions to the repo for that. I'll attempt to past the important bits from .pip/pip.log. Both tracebacks look to be the same but are a little different so I'll add both. This doesn't look like an issues with cookiecutter and more like one with ruamel. 

```
copying ./loader.py -> build/lib.linux-x86_64-2.7/ruamel/yaml

copying ./LICENSE -> build/lib.linux-x86_64-2.7/ruamel/yaml

running build_ext

building '_ruamel_yaml' extension

creating build/temp.linux-x86_64-2.7

creating build/temp.linux-x86_64-2.7/ext

x86_64-linux-gnu-gcc -pthread -fno-strict-aliasing -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -fPIC -I/usr/include/python2.7 -c ext/_ruamel_yaml.c -o build/temp.linux-x86_64-2.7/ext/_ruamel_yaml.o

ext/_ruamel_yaml.c:4:20: fatal error: Python.h: No such file or directory

 #include "Python.h"

                    ^

compilation terminated.

error: command 'x86_64-linux-gnu-gcc' failed with exit status 1

----------------------------------------
Cleaning up...
[...]

PR #621, opened 2015-12-28
Use poyo for user rc files

hackebrot:
I wrote a YAML parser for simple files. :chicken: 

https://pypi.python.org/pypi/poyo

It should be fine for ``cookiecutterrc`` files and may help us overcome the troubles with **PyYAML** and **ruamel.yaml** (#557, #569). :worried:

### DIFF
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -14,10 +14,7 @@ import logging
 import os
 import io
 
-try:
-    import ruamel.yaml as yaml
-except ImportError:
-    import yaml
+import poyo
 
 from .exceptions import ConfigDoesNotExistException
 from .exceptions import InvalidConfiguration
@@ -45,13 +42,12 @@ def get_config(config_path):
     logger.debug('config_path is {0}'.format(config_path))
     with io.open(config_path, encoding='utf-8') as file_handle:
         try:
-            yaml_dict = yaml.safe_load(file_handle)
-        except yaml.scanner.ScannerError as e:
+            yaml_dict = poyo.parse_string(file_handle.read())
+        except poyo.exceptions.PoyoException as e:
             raise InvalidConfiguration(
-                '{0} is not a valid YAML file: line {1}: {2}'.format(
-                    config_path,
-                    e.problem_mark.line,
-                    e.problem))
+                'Unable to parse YAML file {}. Error: {}'
+                ''.format(config_path, e)
+            )
 
     config_dict = copy.copy(DEFAULT_CONFIG)
     config_dict.update(yaml_dict)
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,8 @@ requirements = [
     'binaryornot>=0.2.0',
     'jinja2>=2.7',
     'click>=5.0',
-    'whichcraft>=0.1.1'
+    'whichcraft>=0.1.1',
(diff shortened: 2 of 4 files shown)


pydanny:
I'm having trouble getting this PR to pass tests locally. This is an example error:

```
config_path = '/Users/danielgreenfeld/.cookiecutterrc'

    def get_config(config_path):
        """
        Retrieve the config from the specified path, returning it as a config dict.
        """
    
        if not os.path.exists(config_path):
            raise ConfigDoesNotExistException
    
        logger.debug('config_path is {0}'.format(config_path))
        with io.open(config_path, encoding='utf-8') as file_handle:
            try:
                yaml_dict = poyo.parse_string(file_handle.read())
            except poyo.exceptions.PoyoException as e:
                raise InvalidConfiguration(
                    'Unable to parse YAML file {}. Error: {}'
>                   ''.format(config_path, e)
                )
E               cookiecutter.exceptions.InvalidConfiguration: Unable to parse YAML file /Users/danielgreenfeld/.cookiecutterrc. Error: None of the known patterns match for default_context
E                   full_name:
E                   email: "firstname.lastname@gmail.com"
E                   github_username: "example"
E               cookiecutters_dir: "/home/example/some-path-to-templates"

cookiecutter/config.py:49: InvalidConfiguration
```

pydanny:
Poyo's tests pass for me, but this PR fails. I've rebuilt my `virtualenv` and `tox` environments a few times to no avail. Digging into to see where the problem is...

pydanny:
The problem was on my end with a corrupted `.cookiecutterrc` file. By corrupted, it had a borked line ending.

vincentbernat:
Would it be possible to support whatever YAML parser is already present on the system? I can't package (and maintain) all the YAML parsers on earth just to get cookiecutter in Debian.

hackebrot:
@vincentbernat I'd prefer to get rid of YAML altogether

pydanny:
The problem we have is that the other YAML parsers for Python aren't maintained (PyYAML) and fail in some places or are maintained (Raumel.YAML) and fail in others. The area of failure is over multiple operating systems and versions of python. Trying to gracefully allow local parsers of YAML is actually challenging.

Creative setup.py configurations are tricky to try and garantee. We know because we tried. And ran into many problems.

Poyo is pure python. It works everywhere. By breaking it into its own project we can better test it against multiple python distress and multiple flavors of Linux, OSX, and Windows.

vincentbernat:
In Debian, both of them work fine and the problems they have on other OS do not really concern us. With poyo, I need to introduce a new package whose description would be something like "pure Python YAML parser that doesn't really parse YAML". I will be told that such a package shouldn't be in Debian (not because it doesn't do its job but because we already have two YAML parsers for Python and it would be odd to introduce another one that only accepts a subset of YAML).

I can just patch my copy of cookiecutter to use PyYAML or ruamel.yaml if you don't want more work on this. Would this be OK?

vincentbernat:
FYI, I applied the following patch to avoid an additional dependency: http://sources.debian.net/src/cookiecutter/1.4.0-1/debian/patches/0002-Use-PyYAML-instead-of-poyo.patch/

hackebrot:
Thanks for the update @vincentbernat! 🙇

PR #631, opened 2016-02-04
Add official Tryton cookiecutter

cedk:


hackebrot:
Hey @cedk! :wave: 

Have you tried to get in touch with the author of the related template?

cedk:
Their template is based on their own standard but not the Tryton's standard.
For example in Tryton standard module, we use drone and tox for testing when they use travis and Makefile. So I doubt they will agree to change their methodology as since the beginning they did not followed the Tryton's one.

pydanny:
Hey @cedk, glad to have Tryton submitting an official Cookiecutter template! Tryton is one of those projects which makes me wish I had more time in the day to explore ways to use it for clients and various other things.

I wonder if we should have an "Official Project" section in the README for Cookiecutter templates that qualify as such?

cedk:
Hey @pydanny, we will be glade to help you to explore Tryton. We have mailing lists and IRC channels where is the community is quite active to help others.
Thanks for the merge.

hackebrot:
@pydanny IMHO we should try to find a  way to remove the templates from the readme altogether and have them somewhere else. Something that doesn't require us to maintain it at best :grin:

PR #638, opened 2016-02-24
Explicitly set encoding for reading history file.

0-wiz-0:
Fixes build in C locale. Otherwise I see:

  File "setup.py", line 24, in <module>
    history = history_file.read().replace('.. :changelog:', '')
  File "/usr/pkg/lib/python3.5/encodings/ascii.py", line 26, in decode
    return codecs.ascii_decode(input, self.errors)[0]
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe2 in position 6348: ordinal not in range(128)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ if sys.argv[-1] == 'tag':
 with open('README.rst') as readme_file:
     readme = readme_file.read()
 
-with open('HISTORY.rst') as history_file:
+with open('HISTORY.rst', encoding="utf-8") as history_file:
     history = history_file.read().replace('.. :changelog:', '')
 
 requirements = [


0-wiz-0:
I just saw that this breaks on python 2.x. I don't know what the proper solution for that is, suggestions?
(To reproduce the issue, just clear your environment of `LANG` and all `LC_*` variables and run `python3 setup.py build`.)

hackebrot:
Hi @0-wiz-0, thanks for submitting this issue. :bow: 

TBH Cookiecutter requires **LC_ALL** and **LANG** as we use click for prompts. I'll send a PR to set the encoding though.

hackebrot:
Please try #639

0-wiz-0:
I can confirm that #639 fixes the problem for me, both with python-2.7 and python-3.5.
(Having to set a different locale at runtime is fine with me.)

Thank you for fixing this so quickly!

hackebrot:
Np :smiley: 

Closing in favor of #639

PR #639, opened 2016-02-25
Use io.open with utf-8 instead of open

hackebrot:
Resolve #638


ionelmc:
👍

hackebrot:
:bow: @pydanny @ionelmc

PR #645, opened 2016-03-04
Add cookiecutter-mediawiki-extension template

JonasGroeger:


hackebrot:
Hey @JonasGroeger, thanks for submitting a PR :bow: 

Unfortunately your template raises an error during generation. The problem is [this comment in jinja2 context](https://github.com/JonasGroeger/cookiecutter-mediawiki-extension/blob/master/%7B%7B%20cookiecutter.repo_name%20%7D%7D/%7B%7B%20cookiecutter.repo_name%20%7D%7D.hooks.php#L18).

[This](http://jinja.pocoo.org/docs/dev/templates/#escaping) may help solve the problem.

Also it may be worth adding tests to your template as in [cookiecutter-django](https://github.com/pydanny/cookiecutter-django/blob/master/tests/test_cookiecutter_generation.py)

JonasGroeger:
Hey @hackebrot 

I have solved the problems. Tests are yet to come but for now its alright.

hackebrot:
thanks :bow:

PR #647, opened 2016-03-12
Docs: Fix key name / link to _copy_without_render

ptim:


hackebrot:
Good catch! Thank you @ptim :bow:

Issue #385, opened 2015-02-05
.cookiecutterrc and cookiecutters_dir not working as I expected


apapanico:
Hi,
Here's the setup:
+ I have a ~/.cookiecutterrc as talked about here: http://cookiecutter.readthedocs.org/en/latest/advanced_usage.html
+ I also have a cookiecutters_dir at ~/.cookiecutters/ with the subdirectory cookiecutter-pypackage/.

When I try to run "cookiecutter cookiecutter-pypackage/" in ~/Projects/, I get the following error

    Traceback (most recent call last):
      File "/opt/anaconda/bin/cookiecutter", line 9, in <module>
        load_entry_point('cookiecutter==0.9.0', 'console_scripts', 'cookiecutter')()
      File "/opt/anaconda/lib/python2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 610, in __call__
        return self.main(*args, **kwargs)
      File "/opt/anaconda/lib/python2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 590, in main
        rv = self.invoke(ctx)
      File "/opt/anaconda/lib/python2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 782, in invoke
        return ctx.invoke(self.callback, **ctx.params)
      File "/opt/anaconda/lib/python2.7/site-packages/click-3.3-py2.7.egg/click/core.py", line 416, in invoke
        return callback(*args, **kwargs)
      File "/Users/alexp/Projects/cookiecutter/cookiecutter/cli.py", line 70, in main
        cookiecutter(template, checkout, no_input)
      File "/Users/alexp/Projects/cookiecutter/cookiecutter/main.py", line 95, in cookiecutter
        extra_context=extra_context,
      File "/Users/alexp/Projects/cookiecutter/cookiecutter/generate.py", line 43, in generate_context
        file_handle = open(context_file)
    IOError: [Errno 2] No such file or directory: u'cookiecutter-pypackage/cookiecutter.json'

This error shows up if I either do pip install or with the git repo locally.  Naturally, it makes a bit of sense.  There is no directory ~/Projects/cookiecutter-pypackage/.  

[...]

PR #652, opened 2016-03-16
Complain if we can't locate a local repository

michaeljoseph:
Addresses the stack trace in #385.

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -23,7 +23,8 @@ from cookiecutter.exceptions import (
     InvalidModeException,
     FailedHookException,
     UndefinedVariableInTemplate,
-    UnknownExtension
+    UnknownExtension,
+    RepositoryNotFound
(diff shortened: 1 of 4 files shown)


hackebrot:
Hmm isn't test_replay.py useless now as the test will error out before the assertions are hit?

pydanny:
Hmmmm... You might be right.

hackebrot:
Currently working on the ``jinja2_time`` PR.

Can you please revert the changes in **test_replay.py** and replace ``foobar`` with ``tests/fake-repo-tmpl`` in it? Should do the trick.

hackebrot:
Wrote a tiny test :grin: Do want the coverage to drop, please review 6fed44e26539e607e60acdc39aa73a69e769ec5f on **master**

michaeljoseph:
:grimacing: thanks for the fixes @hackebrot

Issue #610, opened 2015-11-29
Release 1.4.0


hackebrot:
Let's review the following issues/PRs and release v1.4.0 :smiley_cat:

#588 
#595 
#598 
#604 
#608

hackebrot:
#611

pydanny:
#588 merged in!

pydanny:
#604 merged in by @audreyr

pydanny:
Merge in: #598, #608

hackebrot:
How do you feel about a Christmas release? :grinning:

pydanny:
Would be fun. Our time is limited though. :p

hackebrot:
Can you have a brief look at #617? I'll see if I have time to handle the rest :+1: 

Merry Christmas! :santa:

pydanny:
We'll try and get to it this evening.

hackebrot:
Update:

- [x] #588 
- [ ] #595 
- [x] #598 
- [x] #604 
- [x] #608 
- [x] #611 
- [ ] #617 
- [ ] #618 
- [ ] #620 
- [ ] #621 

Template PRs: 
- [ ] #622 
- [ ] #624 

cc @audreyr @pydanny @michaeljoseph :smiley_cat:

hackebrot:
Commencing release :runner: 

---
1.4.0 (2016-03-20) Shortbread
-----------------------------

The goal of this release is changing to a strict Jinja2 environment, paving the way to more awesome in the future, as well as adding support for Jinja2 extensions.

New Features:

-   Added support for Jinja2 extension support, thanks to [@hackebrot](https://github.com/hackebrot) (\#617).
-   Now raises an error if Cookiecutter tries to render a template that contains an undefined variable. Makes generation more robust and secure (\#586). Work done by [@hackebrot](https://github.com/hackebrot) (\#111, \#586, \#592)
-   Uses strict Jinja2 env in prompt, thanks to [@hackebrot](https://github.com/hackebrot) (\#598, \#613)
-   Switched from pyyaml/ruamel.yaml libraries that were problematic across platforms to the pure Python [poyo](https://pypi.python.org/pypi/poyo) library, thanks to [@hackebrot](https://github.com/hackebrot) (\#557, \#569, \#621)
[...]

PR #655, opened 2016-03-18
Fix repo urls

hackebrot:
Resolve #595 and add tests :smile_cat: 

Let's get this in #610


hackebrot:
@michaeljoseph :cookie:

PR #656, opened 2016-03-23
fix small typo on hook error text

luzfcb:


hackebrot:
Thank you @luzfcb :bow:

PR #665, opened 2016-04-05
Link to Hovercraft presentation projects template

jhermann:
I put this under Python because of the tooling, but a new
Documentation section might be a good idea.


jhermann:
… and this was during the GH outage, and has some strange status problems, too. Giving up till the wkend…

hackebrot:
feelsbadman :sob:

audreyr:
Went ahead and merged it in. 😄

PR #667, opened 2016-04-09
Add ratson/cookiecutter-es6-package

ratson:
Add [ratson/cookiecutter-es6-package](https://github.com/ratson/cookiecutter-es6-package)


pydanny:
Merged in. Thanks!

PR #669, opened 2016-04-10
Added a test for the ondelete callback in utils.rmtree

ewjoachim:
This might look not really useful but it allows being really sure that the callback does what we think it does. For the record, this test makes coverage goes from 99.6% to 100% which is the reason I wrote it.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -33,6 +33,44 @@ def test_rmtree():
     assert not os.path.exists('foo')
 
 
+def test_rmtree_error_uses_callback(mocker):
+    """
+    Test that on system who fails deleting files when those files
+    are read only, the workaround that sets them to writable first
+    works.
+    """
+
+    # Setting up mocks to check if callbacks were really called
+    real_force_delete = utils.force_delete
+    force_delete = mocker.patch("cookiecutter.utils.force_delete")
+    force_delete.side_effect = real_force_delete
+
+    real_unlink = os.unlink
+    # python 2.7 uses os.remove, python3.+ uses os.unlink
+    unlink = mocker.patch("os.unlink")
+    mocker.patch("os.remove", new=unlink)
+
+    def fail_if_readonly(*args, **kwargs):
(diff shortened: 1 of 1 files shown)


ewjoachim:
Coverage is positive : this project is now tested "cover to cover". Badum tssss. Sorry, had to make a bad pun here.

hackebrot:
Thank you for submitting this PR @ewjoachim 🙇 

I'm wondering why ``test_rmtree`` does not cover those lines. 😕

ewjoachim:
The original bug for which those lines were written is [tagged "windows"](https://github.com/audreyr/cookiecutter/pull/226). I think the problem is that it seems that as long as you are the owner of the file, on a Unix platform, you can delete it, even if it's readonly, On windows, you can't, you have to pass it write-only first. That's what I'm trying to reproduce in the mocks, a deletion operation that would raise an OSError when trying to remove a readonly file.

If the coverage were to run also on Windows, those line would be marked as covered. That might be another solution.

hackebrot:
Hmm right [appveyor win coverage](https://ci.appveyor.com/project/audreyr/cookiecutter/build/1.0.960/job/naveaoartha3l6qa#L128)

ewjoachim:
Well, if there's a simple way to integrate appveyor coverage information into codecov.io, then it can replace this PR altogether. I'll try to look.

michaeljoseph:
@ewjoachim I think we just need to install and call `codecov` in `appveyor.yml` (like we do for Travis).

ewjoachim:
Will it be able to merge results from different apps ? Or will one overwrite the other one ? Second problem : current build of codecov for AppVeyor is [failing](https://github.com/codecov/codecov-python#ci-providers) :(

ewjoachim:
I'm creating an alternate PR #670 to test and discuss this possibility

ewjoachim:
#670 definitely offers a better way to achieve this. Closing here.

hackebrot:
Thank you @ewjoachim 🙇

PR #677, opened 2016-04-16
Documented __init__ methods for Environment objects

pydanny:
Alas, `Cookiecutter` is forced to have classes because jinja2 requires them. In a better world we could avoid them, but such is how the cookie crumbles. In any case, this pull request documents the `__init__` methods, which is useful because the purpose of those methods (especially the mixin's `__init__`) is not immediately clear.

### DIFF
--- a/cookiecutter/environment.py
+++ b/cookiecutter/environment.py
@@ -13,6 +13,13 @@ class ExtensionLoaderMixin(object):
     the next parent class in line of the child.
     """
     def __init__(self, **kwargs):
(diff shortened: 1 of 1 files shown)


hackebrot:
We could change it so use a function to create the ``jinja2.Environment``, that's how I did it initially. However inheritance seemed like the single best solution to me. I added the mixin as there has been talking around whether we should choose the user to run either with a strict or a non strict environment, so the mixin would allow us to create a second class of let's say ``NonStrictEnvironment`` and re-use the extension loading mechanism.

PR #678, opened 2016-04-19
Documented render_variable function in prompt.py

pydanny:
Another bit of documentation. This is the first time I've ever touched this module!

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -85,6 +85,18 @@ def read_user_choice(var_name, options):
 
 
 def render_variable(env, raw, cookiecutter_dict):
+    """Inside the prompting taken from the cookiecutter.json file, this renders
+    the next variable. For example, if a project_name is "Peanut Butter Cookie", the repo_name could be be rendered with:
+
+        `{{ cookiecutter.project_name.replace(" ", "_") }}`.
+
+    This is then presented to the user as the default.
+
+    :param Environment env: A Jinja2 Environment object.
+    :param str raw: The next value to be prompted for by the user.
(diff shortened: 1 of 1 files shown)


cookiecutter/prompt.py
```diff
@@ -85,6 +85,18 @@ def read_user_choice(var_name, options):
 
 
 def render_variable(env, raw, cookiecutter_dict):
+    """Inside the prompting taken from the cookiecutter.json file, this renders
+    the next variable. For example, if a project_name is "Peanut Butter Cookie", the repo_name could be be rendered with:
```
hackebrot: This line needs to be wrapped 😃

cookiecutter/prompt.py
```diff
@@ -85,6 +85,18 @@ def read_user_choice(var_name, options):
...
+
+        `{{ cookiecutter.project_name.replace(" ", "_") }}`.
+
+    This is then presented to the user as the default.
+
+    :param Environment env: A Jinja2 Environment object.
+    :param str raw: The next value to be prompted for by the user.
+    :param dict cookiecutter_dict: The deserialized cookiecutter.json file.
```
hackebrot: Hmmm I'm afraid it's rather a Python dict that is populated in ``prompt_for_config`` based on "The deserialized cookiecutter.json file" and user input, if that makes sense. In the beginning it is just an empty dict and looks more and more like the json when looping for the key, value pairs.

michaeljoseph:
:bell: @pydanny

hackebrot:
Resolved the flake8 issue and adapted the docs for one of the parameters. 😄

PR #680, opened 2016-04-26
Add cookiecutter-data-science project

pjbull:
Thanks for the awesome cookiecutter project! We <3 it!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -505,6 +505,12 @@ Scala
 
 .. _`cookiecutter-atari2600`: https://github.com/joeyjoejoejr/cookiecutter-atari2600
 
+Data Science
+~~~~~
+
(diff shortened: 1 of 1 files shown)


michaeljoseph:
Thanks for the template @pjbull 😄 
I think this should go under the existing Python category though...?

pjbull:
Either works--I guess calling out a separate data science category made sense for the future (including, possibly, the future of this project including multiple options).

Also, didn't want to pollute the the Python projects with something that wasn't a template for using a particular Python library (e.g., Flask, Django).

I've got a slight preference to leave as is, but happy to move if that's preferred!

audreyr:
I think those are good reasons for a new data science category, and it would be nice to see more data science templates. Thanks @pjbull.

PR #696, opened 2016-05-04
Enable passing extra template paths

dajose:
Enables the usage of base templates located outside of the cookiecutter root on internal API.

This allows to have base templates that are not included on the resulting directory.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -135,6 +135,7 @@ Other Changes:
   * `cookiecutter-pyqt5`_ by `@mandeepbhutani`_ (#797)
   * `cookiecutter-xontrib`_ by `@laerus`_ (#817)
   * `cookiecutter-reproducible-science`_ by `@mkrapp`_ (#826)
+  * `cc-automated-drf-template`_ by `@TAMU-CPT`_ (#832)
 
 .. _`@keimlink`: https://github.com/keimlink
 .. _`@luzfcb`: https://github.com/luzfcb
@@ -179,6 +180,7 @@ Other Changes:
 .. _`@B3QL`: https://github.com/B3QL
 .. _`@laerus`: https://github.com/laerus
 .. _`@mkrapp`: https://github.com/mkrapp
+.. _`@TAMU-CPT`: https://github.com/TAMU-CPT
 
 .. _`cookiecutter-pyqt5`: https://github.com/mandeepbhutani/cookiecutter-pyqt5
 .. _`cookiecutter-funkload-friendly`: https://github.com/tokibito/cookiecutter-funkload-friendly
@@ -210,6 +212,7 @@ Other Changes:
 .. _`cookiecutter-simple-django-cn`: https://github.com/shenyushun/cookiecutter-simple-django-cn
 .. _`cookiecutter-xontrib`: https://github.com/laerus/cookiecutter-xontrib
 .. _`cookiecutter-reproducible-science`: https://github.com/mkrapp/cookiecutter-reproducible-science
+.. _`cc-automated-drf-template`: https://github.com/TAMU-CPT/cc-automated-drf-template
 
 
 1.4.0 (2016-03-20) Shortbread


cookiecutter/generate.py
```diff
@@ -17,6 +17,7 @@
 import shutil
 
 from jinja2 import FileSystemLoader
+from jinja2._compat import string_types
```
michaeljoseph: We're using `python-future`, so you can use [`from past.builtins import basestring`](http://python-future.org/compatible_idioms.html#basestring) instead of using the [private](https://docs.python.org/3.5/tutorial/classes.html#private-variables) jinja2 module.

cookiecutter/generate.py
```diff
@@ -283,9 +286,16 @@ def generate_files(repo_dir, context=None, output_dir='.',
     logging.debug('project_dir is {0}'.format(project_dir))
 
     _run_hook_from_repo_dir(repo_dir, 'pre_gen_project', project_dir, context)
+    templates = ['.']
```
michaeljoseph: Could you move this to below your comment on L291 please?

cookiecutter/generate.py
```diff
@@ -248,6 +249,8 @@ def generate_files(repo_dir, context=None, output_dir='.',
     :param output_dir: Where to output the generated project dir into.
     :param overwrite_if_exists: Overwrite the contents of the output directory
         if it exists
+    :param extra_templates: list of extra template directories to add to jinja2
+        searchpath.
```
michaeljoseph: > to add to _the_ jinja2 search path

tests/test_generate_files.py
```diff
@@ -291,3 +291,21 @@ def test_raise_undefined_variable_project_dir(tmpdir):
...
+        },
+        repo_dir=os.path.abspath('tests/test-generate-external-template'),
+        output_dir='tests/custom_output_dir',
+        extra_templates=os.path.abspath(
+            'tests/test-generate-external-template'
+        )
+    )
+    assert project_dir == os.path.abspath(
+        'tests/custom_output_dir/inputpizzä/'
+    )
```
michaeljoseph: Some thoughts on this test:
- I don't think we need a custom output dir (and if we do, we should be using the `tmpdir` fixture to create it)
- I think we should be verifying the contents of the `extra-template.txt` file rather than the project_dir

dajose:
thanks for your comments @michaeljoseph , I worked on them with the second commit

cookiecutter/generate.py
```diff
@@ -284,8 +287,15 @@ def generate_files(repo_dir, context=None, output_dir='.',
 
     _run_hook_from_repo_dir(repo_dir, 'pre_gen_project', project_dir, context)
 
+    # Configure jinja templates searchpath
+    templates = ['.']
+    if extra_templates is not None:
+        if isinstance(extra_templates, basestring):
```
hackebrot: The doc string for this parameter reads that it is expected to be a list, so this instance check is always going to fail. Is it the items that you want to check?

dajose: you are right, I either 
* improve the documentation (to make clear that a single extra_template is accepted as an string) or
* allow only lists.

I prefer the first option

hackebrot:
Can you please update ``main.cookiecutter``?

Technically ``generate_files`` is not part of the API and no one will be able to use this feature 😿

dajose:
@hackebrot When I planned this feature my idea was for it to be accessible programatically, not as part of the executable. I find it messy to add a list argument to the main as I am not sure on how to pass that on run. 

Maybe we can discuss the best way to expose it and add it on the future.

hackebrot:
I suspect what you mean by "executable" is the command line interface, which you can find in ``cookiecutter.cli``. The function ``main`` calls the internal API, which is exactly what you would do when running Cookiecutter programmatically.

How do you invoke cookiecutter right now?

dajose:
Best if I start with a bit of a background.

The project I'm working on is a python module that given some data it generates 2 python modules that I can, at the end, install. For this I have a directory with base jinja2 templates that processes most of the input and 2 additional directories with a cookiecutter directory templates.

That way, when I execute my project, the inputs are read from my filesystem (or calculated) and files are generated with cookiecutter*. (* I need to have my base jinja2 templates available, but I don't want them on the generated result)

I started using the cookiecutter ```main``` but I face these problems:

* I can't have dictionaries with nested dictionaries (the function ```prompt_for_config``` makes my nested dictionaries a string)
* I need to have a file ```cookiecutter.json```.

So I started using ```generate_files``` as I didn't want the extra processing that the ```main``` does

michaeljoseph:
I'm :+1: on this, and adding `main.cookiecutter` and `cli` support in a follow-up PR.

dajose:
recheck

michaeljoseph:
:bell: @hackebrot 👀 ?

hackebrot:
@michaeljoseph @dajose I'll have a look over the weekend. If you could please resolve the conflicts?

PR #702, opened 2016-05-12
Add ssh examples for git and mercurial

pokoli:


audreyr:
Thanks @pokoli. We really appreciate documentation pull requests like this 💛

hackebrot:
What @audreyr said! 👍

michaeljoseph:
@pokoli, thanks for contributing docs! ❤️ 
However, I think these examples would work better under the [Works with private repos][https://github.com/audreyr/cookiecutter/blob/master/docs/usage.rst#works-with-private-repos) section though?
Perhaps you could add another paragraph explaining that we support [`git` and `ssh` protocols](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/main.py#L35) in addition to http(s)?

PR #734, opened 2016-06-04
Update overview, tutorials, advanced usage -- change occurrences of 'repo_name'

palmerev:
Since naming the project template directory 'repo_name' is no longer required, I changed occurrences of it to 'project_slug' (matching the current `cookiecutter-pypackage`) and clarified that this is a templated value. I also updated the `cookiecutter-pypackage/cookiecutter.json` in the tutorial to match the current version.


audreyr:
Amazing work @palmerev, thanks!

PR #738, opened 2016-06-05
Debian installation

ivanlyon:
Debian package installation steps added to docs/installation.rst.

v2: Now with added note indicating the debian YAML package is a patched package rather than  development YAML.


audreyr:
Thanks @ivanlyon

Issue #732, opened 2016-06-04
Tutorials issues


BruceEckel:
Here: https://cookiecutter.readthedocs.io/en/latest/tutorials.html
Under the subhead:
### Create Your Very Own Cookiecutter
* Audrey's tutorial "Create a Cookiecutter From Scratch" stops after creating some directories and says "To be continued..." (Also says "Create a directory called {{ cookiecutter.repo_name }}" which might be out of date).
* Danny's article seems out of date and isn't really a tutorial. Should probably be unlinked.
* Hackbot's tutorials seem thorough but are too complex for beginners.

I'd like to see (and will try to work on):
1. Creating a "Hello, World"-style template: as minimal an example as possible, to get people over the initial hump.
2. Adding to the above in steps, or creating templates with intermediate complexity.
3. ... to the point where people are ready for Hackbot's tutorials.

audreyr:
Yesterday @BruceEckel wrote https://github.com/BruceEckel/HelloCookieCutter1 which I'd like to turn into the official Cookiecutter tutorial. We've needed a minimal tutorial for a long time, and I think this is perfect.

The next step is to integrate https://github.com/BruceEckel/HelloCookieCutter1 into the Cookiecutter docs. This will require some thought about the best way to do this. Some TODOs:

* Get the link in https://github.com/audreyr/cookiecutter/blob/master/docs/tutorials.rst working properly
    * Does it make sense as a link to an external repo?
    * Or should we integrate it into the main docs as a git submodule or by bringing in the files?

* Add text to https://github.com/audreyr/cookiecutter/blob/master/docs/tutorials.rst to make the tutorial obvious as a starting point for people new to Cookiecutter

dot2dotseurat:
Working on this :)

hackebrot:
Great work @BruceEckel! 👍 

@audreyr I'm 👎 for git submodules - it makes things overly complicated. Prefer the other options 😄

audreyr:
Thanks for the feedback @hackebrot!

Decided to stick with using a git submodule because it eases the burden of maintenance to have less in core, and because @BruceEckel offered to update the tutorial if any issues come up. 

@BruceEckel if you ever decide in the future that you don't want to maintain it anymore, let us know and we can bring the tutorial into core.

PR #746, opened 2016-06-07
Add a make rule to update git submodules

hackebrot:
Related to #740 #732

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,6 +22,7 @@ help:
 	@echo "docs - generate Sphinx HTML documentation, including API docs"
 	@echo "release - package and upload a release"
 	@echo "sdist - package"
+	@echo "submodules - pull and update git submodules recursively"
 
 clean: clean-tox clean-build clean-pyc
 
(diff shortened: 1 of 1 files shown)


audreyr:
Looks good @hackebrot, thanks for doing this 👍 

@BruceEckel if you make updates to https://github.com/bruceeckel/HelloCookieCutter1 and want them to be pulled upstream into http://cookiecutter.readthedocs.io/en/latest/first_steps.html, either send a pull request that does so or let us know.

hackebrot:
Always 🙇

BruceEckel:
I just made updates, so you should refresh your version.

If you do it this way, note that your version still contains info referring
to mine, including a reference to my repository. Which is fine, but you
might also choose to have local references as well.


-- Bruce Eckel
www.MindviewInc.com <http://www.mindviewinc.com/>
Blog: BruceEckel.github.io
www.WinterTechForum.com
www.AtomicScala.com
www.Reinventing-Business.com
http://www.TrustOrganizations.com <http://www.ScalaSummit.com>

Issue #731, opened 2016-06-04
Bump the copyright date at the bottom of the ReadTheDocs pages


BruceEckel:

audreyr:
Fixed in #747. Thanks for reporting this, @BruceEckel.

PR #747, opened 2016-06-07
Updated copyright on docs

zzzirk:
Updated copyright in documentation from 2013-2015 to 2013-2016 per #731.


audreyr:
Thanks @zzzirk!

Close #731.

PR #760, opened 2016-06-20
Add MetricSet generator link

ruflin:
Link to the beat generator had to be updated as the generator is now part of the beats repository.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -468,9 +468,11 @@ Common Lisp
 Golang
 ~~~~~~
 
-* `beat-generator`_: A template to create your own beat to ship data to Elasticsearch.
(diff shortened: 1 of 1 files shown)


michaeljoseph:
We kind of have the expectation that all the cookiecutters we list are able to be passed directly to the `cookiecutter` command. I think it would be confusing if we listed these cookiecutters 😞

ruflin:
@michaeljoseph Yeah, I understand the issue. I was actually hoping for some support of this in cookiecutter (https://twitter.com/ruflin/status/733314584757604352).

As we plan to add more generators in the futures, it is easier for use to maintain them all in one repository.

michaeljoseph:
@ruflin thanks for the PR!
Closing this, since it looks like you've moved away from using cookiecutter to generate beats.

Issue #742, opened 2016-06-06
Run pep257 on each file to improve docstrings


audreyr:
How to do this:
1. `pip install pep257`
2. Run it on a single file to see what needs to be improved: `pep257 cookiecutter/generate.py`
3. Fix as much as you can in the file. Whatever you can't fix, don't worry about fixing.
4. Submit a pull request (1 file per pull request for pep257 fixes, please)

michaeljoseph:
This is now [pydocstyle](https://github.com/PyCQA/pydocstyle).
Once we have clean docstrings, I'd suggest adding flake8 support for this via https://gitlab.com/pycqa/flake8-docstrings.

staranjeet:
@audreyr @michaeljoseph : Submitted a PR #779  for this

talaniz:
Hi, I'm getting ready to make my first pull request, I just made some changes to doc strings on prompt.py. I went run the tox tests as shown in the contribution section and received some errors. I was able to paste some of those errors here: http://pastebin.com/yw07RJrp. I was wondering if this was something to be concerned with or if I need to possibly change something in my environment to get them to pass? I followed the instructions on requirement installations and didn't want to haphazardly setup a pull request if everything wasn't passing...

hackebrot:
Hi @talaniz,

I'm happy to hear that you are about to open your first pull request 👍 

We use the click library for cookiecutter and for Python 3 it requires a Unicode locale to be exported, see [click docs](http://click.pocoo.org/5/python3/#python-3-surrogate-handling). Can you please try again with the environment variables set?

```
# encoding
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8
```

talaniz:
Hello @hackebrot, 

   Thanks for the follow up. I was able to get it working. For some reason, my IDS console threw the errors, but my terminal didn't have any issues running tests. I'm getting an `InterpreterNotFound` error identical to this [issue](https://github.com/audreyr/cookiecutter/issues/404), I'm assuming this is ok as the issue was closed. I'm going to review my work one more time today and submit my pull request.

mandeep:
Is there a checklist of all the files that need to be run against pydocstyle?

meahow:
Looks like these files in the cookiecutter directory still have some pydocstyle issues:
```
cookiecutter/extensions.py
cookiecutter/__init__.py
cookiecutter/log.py
cookiecutter/main.py
cookiecutter/prompt.py
[...]

PR #762, opened 2016-06-22
Fixed docstrings in exceptions.py, ran pep8 and pep257.

terryjbates:
Working on #742.


cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
-    this class.
+    """Base exception class.
+
+    All Cookiecutter-specific exceptions should subclass this class.
     """
 
 
 class NonTemplatedInputDirException(CookiecutterException):
-    """
-    Raised when a project's input dir is not templated.
```
michaeljoseph: I have a preference for the summary line of [multi-line docstrings](https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings) to be on the next line.
>  The summary line may be on the same line as the opening quotes _or on the next line_. 

Thoughts @audreyr @pydanny @hackebrot

cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
     The name of the input directory should always contain a string that is
     rendered to something else, so that input_dir != output_dir.
     """
 
 
 class UnknownTemplateDirException(CookiecutterException):
-    """
-    Raised when Cookiecutter cannot determine which directory is the project
-    template, e.g. more than one dir appears to be a template dir.
+    """Raised when directory that is project template cannot determined.
```
michaeljoseph: Raised when the project template directory cannot be determined.

cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
+
+    More than one dir appears to be a template dir.
     """
 
 
 class MissingProjectDir(CookiecutterException):
-    """
-    Raised during cleanup when remove_repo() can't find a generated project
-    directory inside of a repo.
+    """Raised during cleanup stages when project directory appears missing.
```
michaeljoseph: .. when **the** project directory **is** missing.

cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
+
+    When remove_repo() cannot locate generated project directory inside repo.
```
michaeljoseph: ...cannot locate **the** generated project directory.

cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
+    """Raised when get_config() is passed a invalid config file path."""
 
 
 class InvalidConfiguration(CookiecutterException):
-    """
-    Raised if the global configuration file is not valid YAML or is
-    badly contructed.
+    """Raised if the global configuration file is not valid.
+
+    Invalid YAML or is configuration file is badly constructed.
```
michaeljoseph: Invalid YAML or **the configuration** file is badly constructed.

cookiecutter/exceptions.py
hackebrot: I'm fine with both styles. I personally use:

```python
"""Decorator that produces DEBUG level log messages before and after
calling a parser method.

If a callback raises an IgnoredMatchException the log will show 'IGNORED'
instead to indicate that the parser will not create any objects from
the matched string.
"""
```

and 

```python
"""Helper to log an escaped version of the given message to DEBUG"""
```

terryjbates:
Thanks to all for the feedback! @michaeljoseph I'm making mods on all the changes with the exception of the first. I _think_ I savaged the summary line as such in the first change as such, since I could not craft language to make this fall under 80 chars for pep8.  Also, pep257 complains if the summary line does not end in an explicit "."  I'm not personally wedded to either style, but this may be why I angled for a summary line, with fuller detail on next line. Thoughts? I will go ahead and push the other changes to this branch momentarily.

hackebrot:
Hey @terryjbates! Your last commit's email does not seem to be recognized by GitHub.

terryjbates:
Think I know why. Did the commit from an odd computer that lacked my proper git config info. Sorry about that. Will see if I can set this locally to proper email, amend, and push.

cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
+    """Raised when template uses an undefined variable in the context."""
+
     def __init__(self, message, error, context):
+        """Initialize instance of UndefinedVariableInTemplate class."""
         self.message = message
         self.error = error
         self.context = context
 
     def __str__(self):
+        """Create informal string representation of object."""
```
michaeljoseph: I think we should be suppressing these magic method docstrings.

cookiecutter/exceptions.py
```diff
@@ -1,104 +1,84 @@
...
 
 
 class UndefinedVariableInTemplate(CookiecutterException):
-    """Raised when a template uses a variable which is not defined in the
-    context.
-    """
+    """Raised when template uses an undefined variable in the context."""
+
     def __init__(self, message, error, context):
+        """Initialize instance of UndefinedVariableInTemplate class."""
```
michaeljoseph: @terryjbates could you please add parameter documentation here?

michaeljoseph:
@terryjbates 🔔

terryjbates:
@michaeljoseph Sure thing, I can look into adding param doc.

PR #765, opened 2016-06-24
Added new cookiecutter template

shenyushun:
Cookiecutter-simple-django-cn added to the Django Cookiecutters list.


michaeljoseph:
Thanks @shenyushun!

PR #766, opened 2016-06-25
Ran pep8 and pep257 on hooks.py.

terryjbates:
Work related to issue #742.


cookiecutter/hooks.py
```diff
@@ -76,8 +70,7 @@ def run_script(script_path, cwd='.'):
 
 
 def run_script_with_context(script_path, cwd, context):
-    """
-    Executes a script after rendering with it Jinja.
+    """Execute a script after rendering with its Jinja.
```
michaeljoseph: Execute a script after rendering _it with Jinja._

PR #768, opened 2016-06-27
Ignore backup files of hooks

### LINKED ISSUE #767: Wrong hook executed when a tilde-suffixed file of the same name exists
* Cookiecutter version: 1.4.0
* Template project url: https://github.com/thorgate/django-project-template
* Python version: 3.4
* Operating System: Ubuntu 15.10 wily

### Description:

When using gedit or some other text editor that pollutes the directory with backup files ending with a tilde, cookiecutter mistakes that for the "real" hook it should run. This resulted in cookiecutter running a ridiculously outdated version of my pre-gen hook.

The obvious solution is to just remove `pre_gen_project.py~`, which works, but I believe ideally cookiecutter shouldn't be running it in the first place.

### What I've run:

```
gedit django-template/hooks/pre_gen_project.py
cookiecutter django-template
```

hackebrot:
Resolve #767

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -37,16 +37,20 @@ def find_hooks():
     Missing scripts will not be included in the returned dict.
     """
     hooks_dir = 'hooks'
-    r = {}
+    hooks = {}
     logging.debug('hooks_dir is {0}'.format(hooks_dir))
+
     if not os.path.isdir(hooks_dir):
         logging.debug('No hooks/ dir in template_dir')
-        return r
+        return hooks
+
     for f in os.listdir(hooks_dir):
-        basename = os.path.splitext(os.path.basename(f))[0]
-        if basename in _HOOKS:
-            r[basename] = os.path.abspath(os.path.join(hooks_dir, f))
-    return r
+        filename = os.path.basename(f)
+        basename = os.path.splitext(filename)[0]
+
+        if not filename.endswith('~') and basename in _HOOKS:
+            hooks[basename] = os.path.abspath(os.path.join(hooks_dir, f))
+    return hooks
 
 
 def run_script(script_path, cwd='.'):
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -48,7 +48,7 @@ def find_hooks():
         filename = os.path.basename(f)
         basename = os.path.splitext(filename)[0]
 
-        if not filename.endswith('~') and basename in _HOOKS:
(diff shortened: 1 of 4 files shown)


michaeljoseph:
Nice! 👍

cookiecutter/hooks.py
```diff
@@ -37,16 +37,20 @@ def find_hooks():
...
-        basename = os.path.splitext(os.path.basename(f))[0]
-        if basename in _HOOKS:
-            r[basename] = os.path.abspath(os.path.join(hooks_dir, f))
-    return r
+        filename = os.path.basename(f)
+        basename = os.path.splitext(filename)[0]
+
+        if not filename.endswith('~') and basename in _HOOKS:
+            hooks[basename] = os.path.abspath(os.path.join(hooks_dir, f))
+    return hooks
```
michaeljoseph: I'm not really a fan of this existing implementation- the caller knows which hook it's looking for, so it's weird that we iterate through all the hook files on pre- and post-...

hackebrot: 100% agreed!

cookiecutter/hooks.py
```diff
@@ -37,16 +37,20 @@ def find_hooks():
...
+
     for f in os.listdir(hooks_dir):
-        basename = os.path.splitext(os.path.basename(f))[0]
-        if basename in _HOOKS:
-            r[basename] = os.path.abspath(os.path.join(hooks_dir, f))
-    return r
+        filename = os.path.basename(f)
+        basename = os.path.splitext(filename)[0]
+
+        if not filename.endswith('~') and basename in _HOOKS:
```
michaeljoseph: This is a little hard to parse with the leading `not`, I think either parens for explicitness or better:
`if basename in _HOOKS and not filename.endswith('~')`?

hackebrot: Right, I like the lazy evaluation - so we don't need the dict lookup.

But you're right. I'll change that.

michaeljoseph: Ah right, I didn't consider the short-circuit. Anyways, if we subsequently change this to `find_hook`, this problem goes away...

hackebrot:
Let's refactor ``find_hooks()`` in a separate branch.

michaeljoseph:
🍰 Cool to merge post CI.

hackebrot:
Thanks for the review and the merge @michaeljoseph! 🍰

michaeljoseph:
You make it my pleasure @hackebrot 😻

PR #770, opened 2016-06-30
Increase flexibility of prompted options

mjsisley:
This adds a list of options to be bypassed so that some options can be bypassed even with no_input=False. This adds flexibility when triggering a cookiecutter project in a post_gen hook of another cookiecutter project. 

Currently, the no_input parameter controls whether the user will be prompted for all or none of the options. This PR makes it so the user can be prompted for another percentage of the options.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -68,7 +68,7 @@ def expand_abbreviations(template, config_dict):
 
 
 def cookiecutter(
-        template, checkout=None, no_input=False, extra_context=None,
+        template, checkout=None, no_input=False, bypass_options=(), extra_context=None,
         replay=False, overwrite_if_exists=False, output_dir='.',
         config_file=USER_CONFIG_PATH):
     """
@@ -131,7 +131,7 @@ def cookiecutter(
 
         # prompt the user to manually configure at the command line.
         # except when 'no-input' flag is set
-        context['cookiecutter'] = prompt_for_config(context, no_input)
+        context['cookiecutter'] = prompt_for_config(context, no_input, bypass_options)
 
         dump(config_dict['replay_dir'], template_name, context)
 
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -68,9 +68,9 @@ def expand_abbreviations(template, config_dict):
 
 
 def cookiecutter(
-        template, checkout=None, no_input=False, bypass_options=(), extra_context=None,
-        replay=False, overwrite_if_exists=False, output_dir='.',
-        config_file=USER_CONFIG_PATH):
+        template, checkout=None, no_input=False, bypass_options=(), 
+        extra_context=None, replay=False, overwrite_if_exists=False, 
+        output_dir='.', config_file=USER_CONFIG_PATH):
     """
     API equivalent to using Cookiecutter at the command line.
 
@@ -131,7 +131,8 @@ def cookiecutter(
 
         # prompt the user to manually configure at the command line.
         # except when 'no-input' flag is set
-        context['cookiecutter'] = prompt_for_config(context, no_input, bypass_options)
+        context['cookiecutter'] = prompt_for_config(context, no_input, 
+                                                    bypass_options)
 
         dump(config_dict['replay_dir'], template_name, context)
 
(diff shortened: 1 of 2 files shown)


michaeljoseph:
Hi @mjsisley, can you give us an example of how this feature will be used? At first glance, this doesn't sounds widely applicable for core inclusion.

mjsisley:
Sure, 

This helps compose multiple cookiecutter projects. In [cookiecutter-django](https://github.com/pydanny/cookiecutter-django)'s [post_gen_project.py](https://github.com/pydanny/cookiecutter-django/blob/master/hooks/post_gen_project.py), [cookiecutter-webpack](https://github.com/hzdg/cookiecutter-webpack) is added to the project if it is selected as an option.

```
def add_webpack():
    """
    Adds webpack configuration using cookiecutter to install hzdg/cookiecutter-webpack
    """
    cookiecutter(
        'https://github.com/hzdg/cookiecutter-webpack.git',
        replay=False, overwrite_if_exists=True, output_dir='../',
        checkout='pydanny-django', no_input=True, extra_context={
            'project_name': '{{ cookiecutter.project_name }}',
            'repo_name': '{{ cookiecutter.project_slug }}',
            'repo_owner': '',
            'project_dir': '{{ cookiecutter.project_slug }}',
            'static_root': '{{ cookiecutter.project_slug }}/static/{{ cookiecutter.project_slug }}',
            'production_output_path': '{{ cookiecutter.project_slug }}/static/{{ cookiecutter.project_slug }}/dist/',
            'author_name': '{{ cookiecutter.author_name }}',
            'description': '{{ cookiecutter.description }}',
            'version': '{{ cookiecutter.version }}',
            'existing_project': 'y',
            'css_extension': 'sass',
            'use_ejs': 'n',
            'open_source_license': '{{ cookiecutter.open_source_license }}'
        })
```

It is possible that some options could apply to cookiecutter-webpack and not cookiecutter-django ('use_ejs' and maybe 'css_extensions' fit this description). It makes sense to prompt the user for these options only if webpack is selected as an option in cookiecutter-django. However, it doesn't make sense to prompt the user for the `author_name` and similar. 

Without something like this PR, either all or none of the options will be prompted for toggled by the `no_input` flag. The `bypass_options` parameter gives more control when cookiecutter projects are started within another cookiecutter project.

(I'll add docs and get coverage to 100% if you want the PR)

hackebrot:
Hmmm I can see how we accept a list of ``str`` for ``no_input`` and do not query those variables.

mjsisley:
@hackebrot, to clarify... do you mean you "can see how we could accept..."? I first thought that you were saying that `no_input` already did that and that I missed something. 

Assuming that is true, it would be better to have `no_input` handle a list of strings rather than add a parameter?

hackebrot:
I'm thinking about ways to add this functionality w/o adding more optional parameters. Especially since this directly relates to ``no_input`` and **maybe** it makes sense to allow something like the following:

```python
cookiecutter(
    'https://github.com/hzdg/cookiecutter-webpack.git',
    checkout='pydanny-django',
    no_input=['author_name'],
)
```

pydanny:
Why not just have cookiecutter-django's hook implement this?

mjsisley:
@hackebrot, that makes a lot of sense... I can convert the PR to that if this is wanted.

@pydanny TL;DR I thought about this, but it seems that having `no_input` take a list of strings would be useful for some options, while implementing it in the hook makes sense for others. I could be wrong.

I think that you could use cookiecutter-django's hooks for certain items (things that apply to both cookiecutter-webpack and cookiecutter-django), but asking questions in cookiecutter-django that only apply to cookiecutter-webpack when webpack isn't selected would be annoying. 

As an alternative, triggering more prompts one by one in cookiecutter-django's hooks to then set `extra_content` seems like it could be possible (and maybe is already implemented, I have not looked), but seems to me to be more work  than optionally letting `no_input` take a list of strings. I could be wrong.

I am not really sure how this project is governed but if you guys decide that you want this added then please let me know and I will clean up the PR (I don't want to take time cleaning it up and adding docs if you guys will reject it anyways).

mjsisley:
Sorry, too much text. I think this PR allows increased flexibility when having a cookiecutter triggered in a post_gen hook, while not preventing anything from being done as it is now for those uninterested.  

Will this PR be merged if I do the following?

- `no_input` takes string, no `bypass_options` parameter added.
- add docs
- add tests/ coverage = 100%

Just give me a thumbs up and I'll get it together over the weekend.

hackebrot:
See #848

PR #772, opened 2016-07-01
Fix module summary and 'expand_abbreviations' docstring as per pep257

terryjbates:


hackebrot:
Thank you @terryjbates! 🍪

PR #774, opened 2016-07-05
Include template directory or url in the context dictionary

aroig:
This is one use case: Have a make target in the rendered project that
pulls changes from the template source, renders the up-to-date template
with the same context used to create the project, and then merges the
changes back into the rendered project.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -79,6 +79,9 @@ def cookiecutter(
         # except when 'no-input' flag is set
         context['cookiecutter'] = prompt_for_config(context, no_input)
 
+        # include template dir or url in the context dict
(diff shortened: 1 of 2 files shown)


aroig:
By the way, I'm not sure why travis tests fail. If you accept merging something like this, I can fix any test that needs fixing and add documentation or an entry in history.

aroig:
I force-pushed a fix for the tests.

hackebrot:
Hi @aroig! 👋 

Thank you for opening a PR for this.

What do you think @audreyr @pydanny @michaeljoseph?

hackebrot:
Thank you for your contribution @aroig! 🍪

PR #778, opened 2016-07-12
Friendly error messages for VCS failures

### LINKED ISSUE #776: Unhelpful error if template branch is specified incorrectly
* Cookiecutter version: 1.4.0
* Template project url: https://github.com/pybee/Python-Android-template
* Python version: 2.7
* Operating System: macOS

### Description:

If you specify a template project URL that exists, but a branch that does *not* exist, the `git clone` command fails, raising an exception.

This is entirely expected. However, the error message that is raised isn't especially helpful - it's a simple dump of the subprocess failure, rather than something pointing at the underlying problem.

### What I've run:

The problem is easy to reproduce - specify a valid template URL, but a branch name that doesn't exist. In this example, I'm using Python-Android-template, with the non-existent 2.7 branch.

    $ cookiecutter https://github.com/pybee/Python-Android-template --checkout 2.7
    You've cloned /Users/rkm/.cookiecutters/Python-Android-template before. Is it okay to delete and re-clone it? [yes]: 
    Cloning into 'Python-Android-template'...
    remote: Counting objects: 174, done.
    remote: Total 174 (delta 0), reused 0 (delta 0), pack-reused 174
    Receiving objects: 100% (174/174), 75.87 KiB | 0 bytes/s, done.
    Resolving deltas: 100% (64/64), done.
    Checking connectivity... done.
    error: pathspec '2.7' did not match any file(s) known to git.
      File "setup.py", line 56, in <module>
        test_suite='tests'
      File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/core.py", line 151, in setup
        dist.run_commands()
      File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 953, in run_commands
        self.run_command(cmd)
      File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/distutils/dist.py", line 972, in run_command
        cmd_obj.run()
      File "/Users/rkm/projects/beeware/briefcase/briefcase/app.py", line 193, in run
        self.generate_app_template()
      File "/Users/rkm/projects/beeware/briefcase/briefcase/app.py", line 110, in generate_app_template
        'month': date.today().strftime('%B'),
      File "/Users/rkm/.virtualenvs/briefcase/lib/python2.7/site-packages/cookiecutter/main.py", line 107, in cookiecutter
        no_input=no_input
      File "/Users/rkm/.virtualenvs/briefcase/lib/python2.7/site-packages/cookiecutter/vcs.py", line 122, in clone
        cwd=repo_dir)
      File "/usr/local/Cellar/python/2.7.9/Frameworks/Python.framework/Versions/2.7/lib/python2.7/subprocess.py", line 540, in check_call
        raise CalledProcessError(retcode, cmd)
    subprocess.CalledProcessError: Command '[u'git', u'checkout', '2.7']' returned non-zero exit status 1

The message `error: pathspec '2.7' did not match any file(s) known to git.` is contained higher up in the stack trace; a public facing error message that said this would be more helpful than details about non-zero exit status of the subprocess. 

I've also tried invoking this directly via the API:

    from cookiecutter.main import cookiecutter
    cookiecutter(
            'https://github.com/pybee/Python-Android-template',
            no_input=True,
            checkout='%s.%s' % (sys.version_info.major, sys.version_info.minor),
            extra_context={
                'app_name': self.distribution.get_name(),
                'formal_name': self.formal_name,
                'organization_name': self.organization_name,
                'dir_name': self.dir,
                'bundle': self.bundle,
                'year': date.today().strftime('%Y'),
                'month': date.today().strftime('%B'),
            }
        )

Again, this raises a subprocess exception, rather than something more meaningful that could be caught and reported.

michaeljoseph:
Fixes #776


cookiecutter/exceptions.py
```diff
@@ -113,3 +113,26 @@ class RepositoryNotFound(CookiecutterException):
     """
     Raised when the specified cookiecutter repository doesn't exist.
     """
+
+
+class RepositoryCloneFailed(CookiecutterException):
```
hackebrot: 👍

cookiecutter/exceptions.py
```diff
@@ -113,3 +113,26 @@ class RepositoryNotFound(CookiecutterException):
...
+    """Raised when a cookiecutter template can't be cloned."""
+
+    BRANCH_ERRORS = [
+        'error: pathspec',
+        'unknown revision',
+    ]
+
+    def __init__(self, exc, repo_url, checkout):
```
hackebrot: ``exc`` is somewhat cryptic. How do you feel about changing this to ``exception`` or ``error``?

cookiecutter/exceptions.py
```diff
@@ -113,3 +113,26 @@ class RepositoryNotFound(CookiecutterException):
...
+        if 'not found' in exc.output.lower():
+            super(CookiecutterException, self).__init__(
```
hackebrot: So this will use ``Exception`` rather than ``CookiecutterException``. I think we should use the latter instead.

cookiecutter/exceptions.py
```diff
@@ -113,3 +113,26 @@ class RepositoryNotFound(CookiecutterException):
...
+                'The repository {} could not be found, '
+                'have you made a typo?'.format(repo_url)
```
hackebrot: This should raise a ``RepositoryNotFound`` instead maybe?

cookiecutter/exceptions.py
michaeljoseph: Yep, going with `error`.

cookiecutter/exceptions.py
michaeljoseph: Whoops! Agreed...

cookiecutter/exceptions.py
michaeljoseph: Yeah, that's better semantics.

cookiecutter/exceptions.py
```diff
@@ -113,3 +113,26 @@ class RepositoryNotFound(CookiecutterException):
...
     """
+
+
+class RepositoryCloneFailed(CookiecutterException):
+    """Raised when a cookiecutter template can't be cloned."""
+
+    BRANCH_ERRORS = [
+        'error: pathspec',
+        'unknown revision',
+    ]
```
hackebrot: Is this working for both **git** and **mercurial**?

cookiecutter/vcs.py
```diff
@@ -115,9 +115,19 @@ def clone(repo_url, checkout=None, clone_to_dir=".", no_input=False):
...
-        if checkout is not None:
-            subprocess.check_call([repo_type, 'checkout', checkout],
-                                  cwd=repo_dir)
+        try:
+            subprocess.check_output(
+                [repo_type, 'clone', repo_url],
+                cwd=clone_to_dir,
+                stderr=subprocess.STDOUT,
+            )
+            if checkout is not None:
```
hackebrot: I wonder if it's better to have separate **try-except** blocks, so we can raise our exceptions right here and do not need to check the ``exception.output``.

But I'm happy to leave this as is. What do you think?

tests/vcs/test_clone.py
```diff
@@ -239,5 +147,85 @@ def test_clone_should_rstrip_trailing_slash_in_repo_url(mocker, clone_dir):
...
     mock_subprocess.assert_called_once_with(
-        ['git', 'clone', 'https://github.com/foo/bar'], cwd=clone_dir
+        ['git', 'clone', 'https://github.com/foo/bar'],
+        cwd=clone_dir,
+        stderr=subprocess.STDOUT
     )
+
+
+@pytest.mark.parametrize('error_message', [
+    "fatal: repository 'https://github.com/hackebro/cookiedozer' not found",
```
hackebrot: 😁

tests/vcs/test_clone.py
```diff
@@ -239,5 +147,85 @@ def test_clone_should_rstrip_trailing_slash_in_repo_url(mocker, clone_dir):
...
+        '{} could not found, have you made a typo?'
+    ).format(repository_url)
+
+
+def test_clone_unknown_subprocess_error(mocker, clone_dir):
+    """In `clone()`, unknown subprocess errors should be raised."""
+    mocker.patch(
+        'cookiecutter.vcs.subprocess.check_output',
+        autospec=True,
+        side_effect=[subprocess.CalledProcessError(
```
hackebrot: Why is this a list?

cookiecutter/exceptions.py
michaeljoseph: Yup, here are the actual error messages: https://github.com/audreyr/cookiecutter/pull/778/files/55f6f0215a5293b8dc6baafb978e85bc725ffb45#diff-8327dc289ba1c55ec5be92a9f27829ffR186

tests/vcs/test_clone.py
michaeljoseph: I was working around a Python 3.4 mock regression without fully understanding it.

Build failure: https://travis-ci.org/michaeljoseph/cookiecutter/jobs/144090406.
Bug: http://bugs.python.org/issue23661

A better solution is to remove `autospec`.

tests/vcs/test_clone.py
michaeljoseph: Do you even typo, bro? 😬 😉

cookiecutter/vcs.py
michaeljoseph: You make a good point, I think the code _would_ be less surprising, but at the expense of swallowing any unexpected `subprocess` failures.
😕

hackebrot: Right, we could even check output here - but only to confirm that we don't encounter such an *expected subprocess failure*.

I don't want to over complicate things... 🙊

tests/vcs/test_clone.py
hackebrot: haha 👍

tests/vcs/test_clone.py
hackebrot: Removing ``autospec`` is not ideal. What if we set ``side_effect`` on the object?

```python
foobar = mocker.patch('magic', autospec=True)
foobar.side_effect = subprocess.CalledProcessError
```

cookiecutter/vcs.py
michaeljoseph: :+1: to moving the code from `RepositoryCloneFailed` to here.
I'm not quite understanding what you have in mind with the output checks?

tests/vcs/test_clone.py
michaeljoseph: Same problem. Coming around to think that setting `side_effect` to an iterable is not the worst solution for the py34 bug. I'll add a comment.

michaeljoseph: Actually, looks like instantiating the exception before assigning it to `side_effect` works...

michaeljoseph: Ugh, spoke too soon...

michaeljoseph:
🔔 @hackebrot

cookiecutter/vcs.py
```diff
@@ -115,9 +123,29 @@ def clone(repo_url, checkout=None, clone_to_dir=".", no_input=False):
         prompt_and_delete_repo(repo_dir, no_input=no_input)
 
     if repo_type in ['git', 'hg']:
```
hackebrot: I think this is not required as we run ``identify_repo`` before that.

cookiecutter/vcs.py
hackebrot: Something like this... 😀 

```python
try:
    subprocess.check_output(
        [repo_type, 'clone', repo_url],
        cwd=clone_to_dir,
        stderr=subprocess.STDOUT,
    )
except subprocess.CalledProcessError as clone_err:
    if 'not found' in clone_err.output.lower():
        raise RepositoryCloneFailed(
            'The repository {} could not be found, '
            'have you made a typo?'.format(repo_url)
        )
    else:
        raise clone_err

if checkout is not None:
    try:
        subprocess.check_output(
            [repo_type, 'checkout', checkout],
            cwd=repo_dir,
            stderr=subprocess.STDOUT,
        )
    except subprocess.CalledProcessError as checkout_err:
        if any(error in checkout_err.output for error in BRANCH_ERRORS):
            raise RepositoryCloneFailed(
                'The {} branch of repository {} could not found, '
                'have you made a typo?'.format(checkout, repo_url)
            )
        else:
            raise checkout_err

```

michaeljoseph: The current implementation has the virtue of all the output sniffing in one location?

hackebrot: That's right, but we rely on string matching to know which operation failed. 😢

michaeljoseph: Right. It's conceptually one operation though, maybe we can compromise by switching to a single `clone -b <branch>` command in a follow-up?

hackebrot: Sure no worries. I think the command for cloning from a branch differs from mercurial and git. I'm happy either way! 👍

cookiecutter/vcs.py
michaeljoseph: Cool, will fix.

hackebrot:
Great work @michaeljoseph! 👏

pydanny:
🎉

PR #782, opened 2016-07-15
Ensure cookiecutter repositories have json context

michaeljoseph:


cookiecutter/repository.py
```diff
@@ -41,6 +41,20 @@ def expand_abbreviations(template, abbreviations):
     return template
 
 
+def valid_repository(repo_directory):
```
pydanny: While this works great for checking if the `cookiecutter.json` file exists, it doesn't confirm that the `cookiecutter.json` module is in fact valid JSON. Indeed, later on all we do is `json.load`, providing a non-descriptive error message. Perhaps this function could address that as well?

michaeljoseph: I think we do a reasonable job with the [invalid json error message](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/generate.py#L85-L93), but I can certainly move that code to this function...

michaeljoseph: I'm not really happy with the result of this move:
- `valid_repository` now raises instead of just returning a `bool` and I was planning to use it again when [allowing already checked out cookiecutters to be referenced by name](https://github.com/audreyr/cookiecutter/issues/385#issuecomment-197502702).
- i don't like having to essentially process the json twice, it makes more sense for the validation to be where it's actually going to be used.
Ideas?

pydanny: I agree. Explicit is better than implicit. However, the function name `valid_repository` isn't correct unless it tests for JSON validity? Perhaps change the name to `repository_has_cookiecutter_json`? Hmmm... that's unweildy...

hackebrot: I'd be happy with ``repository_has_cookiecutter_json``

michaeljoseph: Sold!

tests/repository/test_determine_repository_should_use_local_repo.py
```diff
@@ -12,3 +15,31 @@ def test_determine_repository_should_use_local_repo():
...
+            checkout=None,
+            no_input=True
+        )
+
+
+def test_local_repo_typo():
+    """An unknown local repository should raise a `RepositoryNotFound`
+    exception.
+    """
+    with pytest.raises(exceptions.RepositoryNotFound):
```
hackebrot: Would be awesome to extend this to also check the error message:

```python
with pytest.raises(exceptions.RepositoryNotFound) as excinfo:
    pass

expected_message = (
    'The repository tests/unknown-repo could not be located or does not contain '
    'a "cookiecutter.json" file.'
)
assert str(excinfo.value) == expected_message
```

tests/repository/test_determine_repo_dir_clones_repo.py
```diff
@@ -43,3 +43,21 @@ def test_determine_repository_url_should_clone(
...
+
+
+def test_repository_url_with_no_context_file(mocker, template_url):
+    mocker.patch(
+        'cookiecutter.repository.clone',
+        return_value='tests/fake-repo-bad',
+        autospec=True
+    )
+
+    with pytest.raises(exceptions.RepositoryNotFound):
```
hackebrot: See ``excinfo`` comment below.

hackebrot:
LGTM 👍

PR #789, opened 2016-07-30
Fix files removed via make clean

hackebrot:
We've added a couple of **hook backup files** as static files to our tests to make sure that they are not accidentally picked up as hooks in templates, see #768.

```
tests/hooks-backup-files/hooks/post_gen_project.py~
tests/hooks-backup-files/hooks/pre_gen_project.py~
```

The problem is that ``make clean`` removes such files and I have to ``git reset --hard`` to restore them as of now. 😢 

This PR creates the hook files in a **pytest.yield_fixture** and runs the test against them.

### DIFF
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -211,7 +211,8 @@ def hook_backup_dir(tmpdir):
 
     yield str(hooks_dir)
 
-    hooks_dir.remove()
+    pre_gen_hook_file.remove()
+    post_gen_hook_file.remove()
 
 
 def test_ignore_hook_backup_files(monkeypatch, hook_backup_dir):
--- a/tests/test_hooks.py
+++ b/tests/test_hooks.py
@@ -11,6 +11,7 @@ import os
 import pytest
 import stat
 import sys
+import textwrap
 
 from cookiecutter import hooks, utils, exceptions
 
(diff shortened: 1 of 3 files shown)


hackebrot:
Anyone? 😁

tests/test_hooks.py
```diff
@@ -181,21 +182,39 @@ def test_run_failing_hook(self):
...
+    hooks_dir = tmpdir.mkdir('hook-files')
+
+    pre_hook_content = textwrap.dedent(
+        u"""
+        #!/usr/bin/env python
+        # -*- coding: utf-8 -*-
+        print('pre_gen_project.py~')
+        """
+    )
+    pre_gen_hook_file = hooks_dir / 'pre_gen_project.py~'
```
michaeljoseph: These files should be in a `hooks` directory?

hackebrot: Good catch! 

In the long run we probably want to change the code to accept a path rather than using the current working directory + ``hooks``.

michaeljoseph: Yeah 👍

hackebrot:
Should be fixed now. 😄

michaeljoseph:
:+1:

hackebrot:
Thank you for the review @michaeljoseph! 🙇

Issue #784, opened 2016-07-20
How to update a project with changes from its cookiecutter?


aroig:
Hi,

I maintain several personal cookiecutters, and I found that I spend a lot of time keeping projects in sync with updates on the cookiecutter.

I started trying a workflow to deal with this, as follows:
1. Every project I render is assumed to be versioned by git. In the project repo, there is a branch `template`, that contains the result of cookiecutter rendering, without any other changes.
2. Every cookiecutter renders a file `.cookiecutter.json` at the root of the project containing the original context used to render that project, in json format.
3. Every project has a `Makefile` containing at least an `update-template` target. This make target does the following:
   - Create a temporary checkout (via `git worktree`) of the `template` branch.
   - Use the stored context in `.cookiecutter.json` to regenerate the project on top of the temporary checkout, overriding previous content.
   - Commit the newly rendered project in the template branch.

This way, the `template` branch tracks the changes from the cookiecutter, and I can use git-merges into master to update the current project.

The key thing is that by using git merges, I can update template files that got havily modified in the project. Most of the time the merge is clean, but from time to time a conflict will arise, which is easily resolved.

I've used it for a while and works pretty well. There are at least a couple of pain points though, that could benefit from changes in cookiecutter upstream, which may be of independent interest.
1. The template I use to generate a `.cookiecutter.json` is a bit hacky. I'd like to do something like this:
   
   ```
   {{ cookiecutter | jsonify }}
   ```
   
   Would you accept distributing this `jsonify` as a default extension? I can provide the code in an other PR.
2. In order to get updates from the cookiecutter, the project stores its url in a `_template` field on its context. Right now, I hard-code this url in the cookiecutter itself. I would love to free my cookiecutters from hard-coding their url, That could be achieved if cookiecutter itself would inject a `_template` field into the context with the url used at the time of rendering. I submitted #774 to that effect.

Additionally, maybe a functionality like the update-template I described above could be moved into cookiecutter upstream. That would require some thinking, I guess...

Of course, any suggestions, or other ways in which a workflow like this could be better supported by cookiecutter would be very welcome!

hackebrot:
Hi @aroig! 👋  Thank you for opening this *issue*.

Sorry for the late response, we are currently very short on time. 😢 

Can you please provide a link to such this ``jsonify`` jinja2 extension, given it already exists?

hackebrot:
[...]

PR #791, opened 2016-07-31
add a jsonify filter to jinja default extensions

aroig:
This enables a template like this

```
{{ cookiecutter | jsonify }}
```

that renders the a python object as json, as suggested in #784

### DIFF
--- a/cookiecutter/environment.py
+++ b/cookiecutter/environment.py
@@ -25,6 +25,7 @@ class ExtensionLoaderMixin(object):
 
         default_extensions = [
             'jinja2_time.TimeExtension',
+            'cookiecutter.extensions.JsonifyExtension'
         ]
         extensions = default_extensions + self._read_extensions(context)
 
--- a/cookiecutter/extensions.py
+++ b/cookiecutter/extensions.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+"""Jinja2 extensions."""
+
+from jinja2.ext import Extension
+import json
+
+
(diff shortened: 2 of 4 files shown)


hackebrot:
Hey @aroig, thank you for submitting this PR so quickly 😄 

I'd be okay with adding this extension. Let's see what the others think.

That being said, before we can accept this change, we would need to have tests for it. Can you please add a couple of tests? I'd like to see at least one example for how a template would use this new extension in an automated test that we can leverage for identifying regressions.

Thanks!

hackebrot:
@aroig can you please add tests for this?

aroig:
@hackebrot Sorry for the delay. I'm quite busy at the moment, but I'll try to find the time.

hackebrot:
@aroig ok, cool. Thank you 🙇

aroig:
@hackebrot I pushed a test for this. Let me know if you want me to write any kind of documentation.

hackebrot:
The test fails 😿

aroig:
Sorry about that. The test passed locally, but obviously was too fragile. It turns out python versions differ on how they render spaces in json...

I force pushed a second attempt.

aroig:
I see this is still tagged as "needs tests". I pushed them some time ago and they pass. Is there anything else to be done?

hackebrot:
Hey @audreyr @pydanny @michaeljoseph! 👋 

What's your opinion on this?

hackebrot:
Thank you @aroig! 🍪

PR #792, opened 2016-07-31
Improve logging

### LINKED ISSUE #633: A "log what was done" option would be nice
Perhaps an option that writes a log file in the generated tree containing just the list of files created?

hackebrot:
This PR adds sophisticated logging to cookiecutter! 😃 

``$ cookiecutter --debug-file generate-plugin.log cookiecutter-pytest-plugin``

Rather than using ``logging.basicConfig`` to set up the root logger and a handler, a ``cookiecutter`` logger with one or two handlers:

- ``StreamHandler`` logging on ``DEBUG`` or ``INFO`` based on ``--verbose``
- ``FileHandler`` logging **always** on ``DEBUG``, but only added if a file is provided via ``--debug-file=foo.log``

The idea is to allow users to optionally output information about files/directories generated w/o polluting ``stdout``. The format for ``DEBUG`` logging now also mentions the template as specified by the user to the log messages, which will come in handy especially when using the debug file .

Resolves #633 

---

While working on [cibopath](https://github.com/hackebrot/cibopath) I noticed side effects of using ``logging.basicConfig`` which luckily never caused problems in Cookiecutter. This PR solves this problem by explicitly setting up the ``cookiecutter`` logger.

```python
# -*- coding: utf-8 -*-

import logging

# This causes problems when run before basicConfig
# Comment out to see different stdout logging
# or to ``logging.getLogger('foo').info('Before basicConfig')``
logging.info('Before basicConfig')

logging.basicConfig(
    format=u'%(levelname)s: %(message)s',
    level=logging.DEBUG
)

logging.info('After basicConfig')
logging.debug('After basicConfig')
```


cookiecutter/log.py
```diff
@@ -0,0 +1,65 @@
...
+        super(ContextFilter, self).__init__()
+        self.template = template
+
+    def filter(self, record):
+        """Add 'template' as extra information on a logging record."""
+        record.template = self.template
+        return record
+
+
+def create_logger(template, stream_level='DEBUG', debug_file=None):
```
ionelmc: Shouldn't this be named `configure_logging`? I expected something totally different given the current name (like a function that just returns a half-configured logger object).

hackebrot: I agree! Updated to ``configure_logger``. Thanks 😄

hackebrot:
What do you think @audreyr @michaeljoseph @pydanny? 😺

cookiecutter/cli.py
```diff
@@ -89,28 +87,31 @@ def validate_extra_context(ctx, param, value):
...
-        if template == u'help':
-            click.echo(click.get_current_context().get_help())
-            sys.exit(0)
+    # If you _need_ to support a local template in a directory
+    # called 'help', use a qualified path to the directory.
+    if template == u'help':
+        click.echo(click.get_current_context().get_help())
+        sys.exit(0)
+
+    template_name = os.path.basename(os.path.abspath(template))
```
michaeljoseph: Does this do the right thing for template URLs?

hackebrot: Hmm looks like it does. I use the same code as we do in [main.py](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/main.py#L64), however I didn't take into account that it's cloned first and hence refers to a directory.

hackebrot: Any ideas? At this stage we don't really know much about the template that is being passed in.

hackebrot: Actually that's not correct, the code in main uses the same parameter.

(it's then used for the ``--replay`` feature)

hackebrot:
I've been trying to solve this issue, however I'm not entirely happy with what I've got so far. I might as well remove the template information from the log records.

What do you think?

hackebrot:
Removed the template info from the log records.

Please have a look and let me know if you're happy to merge this. Thanks 🙇

hackebrot:
Hey @audreyr @pydanny @michaeljoseph! 👋 

If you have the time, can you please review this PR? I'd like to have this merged and cut a release soon-ish.

Thanks 🙇

cookiecutter/find.py
```diff
@@ -7,14 +7,16 @@
...
+
 
 def find_template(repo_dir):
     """Determine which child directory of `repo_dir` is the project template.
 
     :param repo_dir: Local directory of newly cloned repo.
     :returns project_template: Relative path to project template.
     """
-    logging.debug('Searching {0} for the project template.'.format(repo_dir))
+    logger.debug('Searching {0} for the project template.'.format(repo_dir))
```
michaeljoseph: Rm positional index.

cookiecutter/find.py
```diff
@@ -26,7 +28,7 @@ def find_template(repo_dir):
 
     if project_template:
         project_template = os.path.join(repo_dir, project_template)
-        logging.debug(
+        logger.debug(
             'The project template appears to be {0}'.format(project_template)
```
michaeljoseph: Rm positional index.

cookiecutter/generate.py
```diff
@@ -215,8 +223,10 @@ def _run_hook_from_repo_dir(repo_dir, hook_name, project_dir, context):
             run_hook(hook_name, project_dir, context)
         except FailedHookException:
             rmtree(project_dir)
-            logging.error("Stopping generation because %s"
-                          " hook script didn't exit successfully" % hook_name)
+            logger.error(
+                "Stopping generation because {} hook "
+                "script didn't exit successfully".format(hook_name)
```
michaeljoseph: Let's move to single quotes here?

cookiecutter/generate.py
```diff
@@ -260,7 +270,7 @@ def generate_files(repo_dir, context=None, output_dir='.',
     # absolute path for the target folder (project_dir)
 
     project_dir = os.path.abspath(project_dir)
-    logging.debug('project_dir is {0}'.format(project_dir))
+    logger.debug('project_dir is {}'.format(project_dir))
```
michaeljoseph: `Project directory...` ?

cookiecutter/log.py
```diff
@@ -0,0 +1,47 @@
...
+import logging
+import sys
+
+LOG_LEVELS = {
+    'DEBUG': logging.DEBUG,
+    'INFO': logging.INFO,
+    'WARNING': logging.WARNING,
+    'ERROR': logging.ERROR,
+    'CRITICAL': logging.CRITICAL,
+}
```
michaeljoseph: Thoughts on using https://docs.python.org/3/library/logging.html#logging.getLevelName instead?

cookiecutter/main.py
```diff
@@ -67,7 +67,7 @@ def cookiecutter(
         context = load(config_dict['replay_dir'], template_name)
     else:
         context_file = os.path.join(repo_dir, 'cookiecutter.json')
-        logging.debug('context_file is {0}'.format(context_file))
+        logger.debug('context_file is {0}'.format(context_file))
```
michaeljoseph: Positional index.

tests/test_log.py
```diff
@@ -0,0 +1,133 @@
...
+        'Welcome to Cookiecutter',
+
+        'DEBUG cookiecutter: '
+        'Generating project from pytest-plugin',
+
+        'INFO cookiecutter.foo: '
+        'Loading user config from home dir',
+
+        "DEBUG cookiecutter.foo.bar: "
+        "I don't know.",
```
michaeljoseph: Single quotes here.

michaeljoseph:
:+1: on this in general, just noticed a few consistency cleanups.

cookiecutter/generate.py
hackebrot: ``didn't`` 😿

cookiecutter/log.py
hackebrot: If I understand correctly ``getLevelName`` returns a string for a int. I'd like to go the other way so we don't need to pass around ints.

tests/test_log.py
hackebrot: ``don't`` 😿 

I think double quotes are better than escaping the single quote.

cookiecutter/log.py
michaeljoseph: It works both ways:
```
>>> logging.getLevelName('DEBUG') == logging.DEBUG
True
```

Probably best to leave your code as is though.
> In Python versions earlier than 3.4, this function could also be passed a text level, and would return the corresponding numeric value of the level. This undocumented behaviour was considered a mistake, and was removed in Python 3.4, but reinstated in 3.4.2 due to retain backward compatibility.

michaeljoseph:
:shipit:

hackebrot:
Thanks a ton @michaeljoseph! 😂 🍪

PR #800, opened 2016-08-19
Make the build reproducible by avoiding nondeterministic kwargs

lamby:
Whilst working on the Reproducible Builds effort [0], I noticed
that cookiecutter could not be built reproducibly.

As USER_CONFIG_PATH is pased on expanding ~ this varies between different
systems, meaning that building the documentation on a different machine (or
with a different $HOME) ends up with a different result:
- [..] config_file=u'/nonexistent/first-build/.cookiecutterrc' [..]
- [..] config_file=u'/nonexistent/second-build/.cookiecutterrc' [..]

I would normally just change this to `config_file=None` and set it at
runtime, but we need to be clever with NOT_PROVIDED to tell the difference
between the case where we _actually_ pass ``None`..

 [0] https://reproducible-builds.org/

Signed-off-by: Chris Lamb chris@chris-lamb.co.uk

### DIFF
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -21,6 +21,7 @@ from .exceptions import InvalidConfiguration
 
 logger = logging.getLogger(__name__)
 
+NOT_PROVIDED = object()
 USER_CONFIG_PATH = os.path.expanduser('~/.cookiecutterrc')
 
 BUILTIN_ABBREVIATIONS = {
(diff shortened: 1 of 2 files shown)


hackebrot:
Hi @lamby! 👋 

We do support this via ``--default-config``, but can't do this via the Python API right now.

I've submitted a PR that shifts the check to ``main.cookiecutter``.

Can you please have a look at #801 and let me know if that would work for you?

hackebrot:
How's the ``object()`` showing up as in the docs?

lamby:
On my system, it appears as:

```
cookiecutter.config.get_user_config(config_file=<object object>)
```

(I may be using a newer Sphinx that you that strips memory addresses.)

lamby:
Ping on this? :)

lamby:
Another friendly ping on this? :)

hackebrot:
Closing in favor of #861 😄

PR #801, opened 2016-08-20
Allow default config from main

hackebrot:
Please let me know your thoughts! 😄

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -111,15 +111,14 @@ def main(template, extra_context, no_input, checkout, verbose, replay,
             click.echo(click.get_current_context().get_help())
             sys.exit(0)
 
-        user_config = None if default_config else config_file
-
         cookiecutter(
             template, checkout, no_input,
             extra_context=extra_context,
             replay=replay,
             overwrite_if_exists=overwrite_if_exists,
             output_dir=output_dir,
-            config_file=user_config
+            config_file=config_file,
+            default_config=default_config,
         )
     except (OutputDirExistsException,
             InvalidModeException,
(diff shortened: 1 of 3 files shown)


lamby:
Thank you for your quick reply. I am unclear about the goal of this alternative PR given that it leaves the offending kwarg intact and thus the output would still be unreproducible. Perhaps I did not explain the issue well enough in #800..?

lamby:
Thank you for your quick reply. I am unclear about the goal of this alternative PR given that it leaves the offending kwarg intact and thus the output would still be unreproducible. Perhaps I did not explain the issue well enough in #800..?

hackebrot:
Hey, I've now read the site that you've linked in #800.

If I understand correctly, this should then be changed all across the code base?
(``cli.py`` uses the same default)

This PR would allow to explicitly go with the default config when using the Python API to invoke Cookiecutter as opposed to running the CLI app.

lamby:
> If I understand correctly, this should then be changed all across the code base?

Can you clarify what "this" refers to? But regardless, only the changes in #800 are required to ensure that the generated HTML documentation is reproducible.

I'm afraid my initial PR (#800) has still not been fully understood correctly; I am not sure why we are being distracted by changing the behaviour of default configs, etc. #800 does not intend to change any behaviour *except* the HTML output written by Sphinx.

hackebrot:
Ah the documentation.. I see

hackebrot:
https://cookiecutter.readthedocs.io/en/latest/cookiecutter.html#cookiecutter.config.get_user_config

lamby:
Right, the documentation: http://i.imgur.com/0IAsIOK.jpg. Notice how it includes "/home/docs"?

Shall we close this PR and move to #800?

PR #802, opened 2016-08-20
Rename copy_without_render() check

### LINKED ISSUE #741: Rename copy_without_render()
Since `copy_without_render()` no longer performs any actual copying, it should be renamed:

```
def copy_without_render(path, context):
    """
    Returns True if `path` matches some pattern in the
    `_copy_without_render` context setting.

    :param path: A file-system path referring to a file or dir that
        should be rendered or just copied.
    :param context: cookiecutter context.
    """
    try:
        for dont_render in context['cookiecutter']['_copy_without_render']:
            if fnmatch.fnmatch(path, dont_render):
                return True
    except KeyError:
        return False

    return False
```

hackebrot:
Resolve #741 😸 

cc @audreyr

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -28,7 +28,11 @@ from .hooks import run_hook
 
 
 def is_copy_only_path(path, context):
-    """Return True if `path` matches context dict pattern.
+    """Check whether the given `path` should only be copied as opposed to being
+    rendered.
+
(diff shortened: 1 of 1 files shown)


hackebrot:
One of the appveyor environments couldn't connect to GitHub 😠

michaeljoseph:
👍

hackebrot:
Thank you @michaeljoseph! 🍰

PR #806, opened 2016-08-25
Add support for --context-file flag

gonvaled:
http://stackoverflow.com/q/39143409/647991

We have the file `cookiecutter.json`, which defines the default context for a template. I would like to specify, via a command flag, something like:

```
cookiecutter --no-input --context-file cookiecutter.json <cookiecutter-template>
```

So that the same template can be used to generate different projects, without having to enter the data manually on the input prompts. There is a workaround to achieve this:
1. clone the template repo locally
2. modify the `cookiecutter.json` in the template repo
3. specify as template the local clone, and not the github clone

This is less than ideal, because it requires modifying a repository, does not allow for independent storage of the context files, and does not allow to use the same template to easily create different projects.

Is there a way to specify the context to cookiecutter, on the command line?

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -89,8 +89,13 @@ def validate_extra_context(ctx, param, value):
     u'--default-config', is_flag=True,
     help=u'Do not load a config file. Use the defaults instead'
 )
+@click.option(
+    u'--context-file',
+    help=u'Override cookiecutter.json with a user defined json file',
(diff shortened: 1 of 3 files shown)


witold-gren:
Hi everyone 😀 it is possible to accept this simple PR to `cookiecutter`? It is very useful for recreating existing project.

ssbarnea:

PR #815, opened 2016-08-31
Fixed #479 - Added support for dictionary data in configurations.

### LINKED ISSUE #479: passing list to template
I would like to be able to pass a list to the templates. See the two code blocks at the end of this post for an example. Right now it appears that if you pass a list in the config object, it's read as a list of options for that key.

I know that you can use `str.split()` in the jinja2 template, but that's not a real solution, since it's impossible to "escape" the character that's used as the delimiter. What's the best solution here? I would prefer to be able to pass a list in the json object and call it a day, but obviously that doesn't work for the user input prompts.

* `cookiecutter.json`:
```json
{
  "build_steps": [
     "do_something",
     "do_something_else"
  ]
}
```

* `Dockerfile`:
```jinja2
FROM something

{% for step in cookiecutter.build_steps %}
RUN {{ step }}
{% endfor %}
```

freakboy3742:
This adds the ability to specify arbitrarily deep data structures in your cookiecutter JSON file.

### DIFF
--- a/docs/advanced/dict_variables.rst
+++ b/docs/advanced/dict_variables.rst
@@ -1,0 +1,63 @@
+.. _dict-variables:
+
+Dictionary Variables (1.1+)
+---------------------
+
+Dictionary variables provide a way to define deep structured information when
+rendering a template.
+
+Basic Usage
+~~~~~~~~~~~
+
+Dictionary variables are, as the name suggests, dictionaries of key-value
+pairs. The dictionary values can, themselves, be other dictionaries and lists
+- the data structure can be as deep as you need.
+
+For example, you could provide the following dictionary variable in your
+``cookiecutter.json``::
+
+    {
+        "file_types": {
+            "png": {
+                "name": "Portable Network Graphic",
+                "library": "libpng"
+                "apps": [
+                    "GIMP",
+                ]
+            },
+            "bmp": {
+                "name": "Bitmap",
+                "library": "libbmp",
+                "apps": [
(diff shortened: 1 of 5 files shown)


docs/advanced/dict_variables.rst
```diff
@@ -0,0 +1,63 @@
+.. _dict-variables:
+
+Dictionary Variables (1.1+)
```
hackebrot: This feature would be part of **cookiecutter v1.5**.

hackebrot:
Hi @freakboy3742! 👋 

Thank you for submitting this PR. I had a brief glance at the changes and they look good to me. I left one comment for the docs addition.

Hope I get to review this more thoroughly soon.

Thanks again 🙇

freakboy3742:
@hackebrot Thanks for that - I've just updated the documentation as requested.

pydanny:
I like this particular API change, it's very close to what I've thought would be a great way to extend the API, albeit in a way that will force people to upgrade. The code is very clean too.

That said, does it make sense to possibly flag some keys for special behavior? Not in this particular pull request, but in a new one (or branch). For example, adding a dictionary that allows for better help specifications on the command-line (FWIW, this was a Guido Van Rossum request).

Anyway, this PR gets my 👍

hackebrot:
Great work @freakboy3742! Thank you 🙇 

I fixed a few minor issues on the docs example with the JSON context and the jinja2 template.

PR #820, opened 2016-09-07
Update expired link in readme

B3QL:


hackebrot:
Thanks for fixing this @B3QL! 👍

PR #826, opened 2016-09-20
Add cookiecutter-reproducible-science template

mkrapp:
Here's a new template for any transparent and reproducible science project.


hackebrot:
Nice one @mkrapp! 🍪

PR #831, opened 2016-10-07
Implement pep257 on vcs.py.

terryjbates:


hackebrot:
Great work @terryjbates! 🍪

PR #833, opened 2016-10-07
Implement pep257 on repository.py

terryjbates:


hackebrot:
Thank you @terryjbates! 🍪

PR #839, opened 2016-10-23
Added: Path templates will be rendered when copy_without_render used

### LINKED ISSUE #456: Allow for copy_without_render to render output directory name
I find it strange that copy_without_render doesn't render the directory name.

An example use case is wanting to copy a templates directory in your project that contains jinja, the directory is copied however it is placed into a folder like '{{cookiecutter.repo_name}}/templates' as opposed to 'someproject/templates'

I understand some people may in fact want this, but maybe an option to to toggle this behavior would welcome.

noirbizarre:
This pull-request ensures paths variables are rendered when using `_copy_without_render`.

It should fix #456.

It does not keep the existing behavior as it was inconsistent, different for files and directories (matched files had their path rendered but not the matched directories).
If this behavior should be kept, I'll update this pull-request with a boolean option like `_copy_without_render_paths`.

### DIFF
--- a/docs/advanced/copy_without_render.rst
+++ b/docs/advanced/copy_without_render.rst
@@ -15,3 +15,14 @@ To avoid rendering directories and files of a cookiecutter, the `_copy_without_r
             "rendered_dir/not_rendered_file.ini"
         ]
     }
+
+**Note**: Only the content of the files will be copied without being rendered. The paths are subject to rendering. This allows you to write::
+
+    {
+        "project_slug": "sample",
+        "_copy_without_render": [
+            "{{cookiecutter.repo_name}}/templates/*.html",
+        ]
+    }
+
+In this example, `{{cookiecutter.repo_name}}` will be rendered as expected but the html file content will be copied without rendering.
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -323,6 +323,7 @@ def generate_files(
             for copy_dir in copy_dirs:
                 indir = os.path.normpath(os.path.join(root, copy_dir))
                 outdir = os.path.normpath(os.path.join(project_dir, indir))
+                outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
                 shutil.copytree(indir, outdir)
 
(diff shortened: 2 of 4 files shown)


noirbizarre:
Any chance to see that one merged ? It would be great !!!
Should I change something ? ...
Any comment, any review, I'll update so it can be merge 👍

noirbizarre:
Hi !
@pydanny @audreyr @michaeljoseph @pfmoore @hackebrot 👋
I never had any feedback on this. As this a this a really simple fix to a really annoying issue (to me), I just want to know:
- is there something I need to change ?
- will this be merged ?
- If not, why ?

Any review, any comment, I'll make the changes ASAP !

theodesp:
I think it looks good @noirbizarre

michaeljoseph:
@noirbizarre could we please have a small update to the [docs](https://github.com/audreyr/cookiecutter/blob/master/docs/advanced/copy_without_render.rst) to include a note about the path vs file content behaviour of this functionality.
Perhaps with an explanatory example?

noirbizarre:
Done.
Sorry for the late response, I just noticed yours

docs/advanced/copy_without_render.rst
```diff
@@ -15,3 +15,14 @@ To avoid rendering directories and files of a cookiecutter, the `_copy_without_r
...
+**Note**: Only the content of the files will be copied without being rendered. The paths are subject to rendering. This allows you to write::
+
+    {
+        "project_slug": "sample",
+        "_copy_without_render": [
+            "{{cookiecutter.repo_name}}/templates/*.html",
+        ]
+    }
+
+In this example, `{{cookiecutter.repo_name}}` willbe rendered as expected but the html file content will be copied without rendering.
```
hackebrot: ``is rendered``

hackebrot:
I wonder if we should treat this as a breaking change... 🤔 

@audreyr @pydanny @michaeljoseph

noirbizarre:
I fixed the typo.

My opinion: I don't think this should be considered as a breaking change but as a fix because the previous behavior was inconsistent (different behavior for files and directories) and unless you are making a meta cookiecutter (ie. a cookiecutter template rendering cookiecutter templates), I don't think there is any reason to render brackets into path (which was the behavior for directories before this fix because path variables were not rendered).

noirbizarre:
@audreyr @pydanny @michaeljoseph @theodesp @hackebrot 
Could this be merged before the 1.7 is released ?

insspb:
@noirbizarre Thank you for contribution. Nothing lost. Merged after 4 year in queue :)

newlog:
Thanks for this change.

PR #860, opened 2016-11-28
load custom extensions in pre-hook

### LINKED ISSUE #850: pre_gen hook doesn't appear to be loading custom extensions
Latest version of cookiecutter from master. Last commit [ceafa1dc1b5eb70835ffba5dc7457df269d77ba8](https://github.com/audreyr/cookiecutter/commit/ceafa1dc1b5eb70835ffba5dc7457df269d77ba8)

In my pre_gen_project hook, I have this code: `{{cookiecutter.should_be_list | to_sequence}}` which turns a string into a list but cookiecutter raises a TemplateAssertionError `jinja2.exceptions.TemplateAssertionError: no filter named 'to_sequence'`. I know that my to_sequence filter isn't the issue because when I use it in other parts of my template and remove it from the pre_gen hook, everything executes successfully.

Are the custom extensions intended to only be loaded after the pre_gen hook is finished executing?

cheungnj:
Fixes #850 

StrictEnvironment is used instead of Template when processing the pre-hook. I changed the hooks for the tests because strings weren't getting returned.

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -9,8 +9,6 @@ import subprocess
 import sys
 import tempfile
 
-from jinja2 import Template
-
 from cookiecutter import utils
 from .exceptions import FailedHookException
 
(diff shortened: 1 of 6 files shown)


cheungnj:
I have no idea where force_delete was getting called. Should I add a test for it?

hackebrot:
Hi @cheungnj! 👋 

Thank you for resolving this issue. Great work!

Don't worry too much about the code coverage. I don't really see a reason why your changes would cause ``force_delete`` to be missed.

I'm happy to merge this. Let's see what the others think.

@audreyr @pydanny @michaeljoseph

cookiecutter/hooks.py
```diff
@@ -106,23 +105,25 @@ def run_script_with_context(script_path, cwd, context):
         mode='wb',
         suffix=extension
     ) as temp:
-        output = Template(contents).render(**context)
+        template = env.from_string(contents)
```
michaeljoseph: Could we just instantiate `StrictEnvironment` here, instead of passing the env through all these function calls?

hackebrot:
Thank you @cheungnj! 🍪

PR #861, opened 2016-12-01
Reproducible build with default config

hackebrot:
This is an alternative implementation for #800 to make kwargs deterministic.

I personally prefer a ``default_config=False`` flag over ``config_file=object()`` 🙊 

Please let me know your thoughts, @lamby. I should also mention that you will get credit for your contribution to cookiecutter regardless of whether we decide to use this PR instead of #800!

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -14,7 +14,6 @@ import json
 import click
 
 from cookiecutter import __version__
-from cookiecutter.config import USER_CONFIG_PATH
 from cookiecutter.log import configure_logger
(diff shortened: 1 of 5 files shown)


lamby:
An elaboration (or better: a comment) on how you differentiating between being passed ``None`` and not being passed anything would seem to be a good idea.

hackebrot:
@lamby I've updated the docstring of ``config.get_user_config()``

lamby:
In terms of which to merge, I'm easy. :)

Issue #865, opened 2016-12-07
Symlinks become normal directories when cookiecutter is run


pjbull:
* Cookiecutter version: master
* Template project url: 
* Python version: 3.5
* Operating System: OSX

### Description:

Symlinks within cookiecutter templates are copied as directories rather than as symlinks.

### What I've run:

Here's a cookiecutter template with just the `name` variable in `cookiecutter.json`:
```
12:42 $ tree symlink-cc/
symlink-cc/
├── cookiecutter.json
└── {{\ cookiecutter.name\ }}
    ├── original
    │   └── afile.txt
    └── symlink -> original
```

I can successfully run cookiecutter.
```
12:38 $ cookiecutter symlink-cc
name [pjb]: pjb
```

In the output, the symlink turns into a standard directory.
```
12:42 $ tree pjb/
pjb/
├── original
│   └── afile.txt
└── symlink
```

Working on a PR that should fix this.

michaeljoseph:
Nice catch @pjbull! We look forward to your PR 😃

davidwtbuxton:
What I would like is support for copying all symlinks as-is, not just symlinks to existing directories.

My use case is I have symlinks in the cooiecutter template that use relative paths to point to directories that do not exist (specifically, to directories that will be created by `npm install` in a new project).

In that situation, symlinks are added to os.walk's files list (not dirs list), and cookiecutter stops with an IOError exception when going to read the symlink's target.

```
Traceback (most recent call last):
  File "/Users/david/venv/bin/cookiecutter", line 11, in <module>
    load_entry_point('cookiecutter==1.5.1', 'console_scripts', 'cookiecutter')()
  File "/Users/david/venv/lib/python2.7/site-packages/click/core.py", line 722, in __call__
    return self.main(*args, **kwargs)
  File "/Users/david/venv/lib/python2.7/site-packages/click/core.py", line 697, in main
    rv = self.invoke(ctx)
  File "/Users/david/venv/lib/python2.7/site-packages/click/core.py", line 895, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/david/venv/lib/python2.7/site-packages/click/core.py", line 535, in invoke
    return callback(*args, **kwargs)
  File "/Users/david/venv/lib/python2.7/site-packages/cookiecutter/cli.py", line 123, in main
    default_config=default_config,
[...]

PR #868, opened 2016-12-07
Add support for symlinks in cookiecutter templates

pjbull:
- Support copying symlinks as symlinks for both rendered and
non-rendered directories
 - Support symlinks that link to a rendered directory
 - Support symlinks that get rendered

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -110,6 +110,8 @@ Other Changes:
 * Remove pycache directories in make clean-pyc, thanks to `@hackebrot`_ (#849)
 * Refactor hook system to only find the requested hook, thanks to
   `@michaeljoseph`_ (#834)
+* Add tests for custom extensions in ``pre_gen_project`` and
+  ``post_gen_project`` hooks, thanks to `@hackebrot`_ (#856)
 * Added more cookiecutter templates to the mix:
 
   * `cookiecutter-funkload-friendly`_ by `@tokibito`_ (#657)


pjbull:
This PR aims to tackle #865

pjbull:
Sorry for the noise here--don't have a Windows+Python 2.7 environment, so was letting AppVeyor test things for me.

This fix generally works, but there is a complication: Windows+Python2.7 doesn't support symlinks out of the box. To make this work, I've added the `win_utils.py` file which pulls together a bunch of kernel32 functions to enable symlinks. It looks to me like there's one package on PyPI that would enable this similar functions [jaraco.windows](https://github.com/jaraco/jaraco.windows) if the preference is to take an external dependency on this pacakge in the Windows+Python2.7 case.

I have a little work left to re-implement rmtree in a way that supports symlinks in Windows in the tests. Also, letting flake8 fail on the win_utils.py until we settle whether to use those versions of take the `jaraco.windows` dependency.

pjbull:
Finally got a Windows VM to develop against so could get symlinks working there. Separated out the posix and the Windows changes into separate commits so we can reason about them separately.

For background, most concise summary of symlinks in Windows and Python <3.2 is here:
 - http://stackoverflow.com/a/15259028/1692709
 - I have a [separate branch taking a dependency on `jaraco.windows` instead of rolling our own functions](https://github.com/pjbull/cookiecutter/tree/win-symlink-jaraco), but still have some failing tests when running `rmtree` on a dir with symbolic links.

bittner:
Can we merge this PR?

I understand it's a bit large, mainly due to adding symlink support for Windows. If this is a problem can we get the PR merged earlier if we move the Windows-related changes to a separate PR?

hackebrot:
@pjbull @bittner I'm not really confident merging and maintaining this. It adds a lot of complexity and it's hard for me to test as I don't usually use windows.

I don't object merging this, but we need someone with windows expertise to review this.

bittner:
Frankly, symlinks on Windows feels rather artificial to me. I wouldn't know how to create them using the file manager. Do we really have a issue with symbolic links on Windows? If we have cookiecutters (templates) that have symbolic links (e.g. created on Unix OSes) under version control, do they have a meaning under Windows?

@hackebrot If the complexity that Windows introduces were removed in this PR would you be willing to merge this PR right away?

bittner:
@hackebrot Should the complexity of Windows be removed from this PR to make the decision of merging easier?

pjbull:
Added just the PR for posix systems and newer Python, which is cleaner. I'll create an issue for Windows with Python < 3.2, which we may end up won't fixing given the complexity it introduces.

insspb:
@pjbull, unfortunately, it was not reviewed in time. Currently, we are making cookiecooter issues and prs cleanup after some time without project maintenance. If you can adapt this code to currentl project code state it will be reviewed much faster again.

insspb:
@ssbarnea Please stop closing pull requests without any resolution. At least for problematic functions like symlinks. 

It is hard to find them again. And this problem not yet solved.

PR #871, opened 2016-12-18
Remove HISTORY.rst from long_description in setup.py

hackebrot:
Currently we get errors when generating the ``long_description`` for PyPI as there are duplicate explicit target names for **README.rst** and **HISTORY.rst** for a number of template repositories.


hackebrot:
Thank you @pydanny! 👍

PR #873, opened 2016-12-21
cookiecutter-golang in pantry

lacion:
added a new project to the pantry in the readme file


michaeljoseph:
@lacion I already merged in #872, but your description is nicer.

Thanks :)

Issue #869, opened 2016-12-14
Docs: remove [sudo] from Installation, add pre-reqs and packaging tools reference


stevepiercy:
* Cookiecutter version: master
* Template project url: n/a
* Python version: n/a
* Operating System: n/a

### Description:

In the [installation instructions](https://cookiecutter.readthedocs.io/en/latest/installation.html), there are an optional `[sudo]`s, which implies installing cookiecutter into a system Python instead of a virtual environment.

In Pyramid, we do more user hand-holding, starting from installing Python, then satisfying [Requirements for Installing Packages](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html#requirements-for-installing-packages), and finally creating a virtual environment into which we install Pyramid for either [UNIX](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html#installing-pyramid-on-a-unix-system) or [Windows](http://docs.pylonsproject.org/projects/pyramid/en/latest/narr/install.html#installing-pyramid-on-a-windows-system).

Would cookiecutter maintainers be interested in an approach similar to Pyramid's? I would submit the PR, if amenable.

### What I've run:

n/a

hackebrot:
Hey @stevepiercy! 👋 

I think that's a great idea. Would love to review a PR.

mmerickel:
FWIW ``pip install --user cookiecutter`` and a note about ensuring that the relevant bin folder ``$HOME/.local/bin`` (not sure what it is on windows) is on your path is much closer to a best practice these days than anything else involving pip installing a system-wide tool.

stevepiercy:
From pip's help:

>  --user                      Install to the Python user install directory for your platform. Typically ~/.local/, or %APPDATA%\Python on Windows. (See the Python documentation for [site.USER_BASE](https://docs.python.org/3/library/site.html#site.USER_BASE) for full details.)

@hackebrot I have some docs to finish up in Pyramid first, then I'll return to this issue.

stevepiercy:
@hackebrot see #880

stevepiercy:
Closed by merging #880

PR #880, opened 2016-12-27
Update installation.rst instructions

stevepiercy:
ref: #869
- Add Prerequisites section
- remove sudo
- use --user install option
- use --upgrade for consistency

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -2,17 +2,69 @@
 Installation
 ============
 
+Prerequisites
+-------------
+
+* Python interpreter
+* Adjust your path
+* Packaging tools
+
+Python interpreter
+^^^^^^^^^^^^^^^^^^
+
+Install Python for your operating system. Consult the official `Python documentation <https://docs.python.org/3/using/index.html>`_ for details.
+
+You can install the Python binaries from `python.org <https://www.python.org/downloads/mac-osx/>`_. Alternatively on macOS, you can use the `homebrew <http://brew.sh/>`_ package manager.
+
+.. code-block:: bash
+
+    # for python 3.x
+    $ brew install python3
+
+
+Adjust your path
+^^^^^^^^^^^^^^^^
+
+Ensure that your ``bin`` folder is on your path for your platform. Typically ``~/.local/`` for UNIX, or ``%APPDATA%\Python`` on Windows. (See the Python documentation for `site.USER_BASE <https://docs.python.org/3/library/site.html#site.USER_BASE>`_ for full details.)
+
+On UNIX and macOS, for bash shells, add the following to your ``.bash_profile`` (adjust for other shells):
+
+.. code-block:: bash
+
+    # Add ~/.local/ to PATH
+    export PATH=$HOME/.local/bin:$PATH
+
+Remember to load changes with ``source ~/.bash_profile`` or open a new shell session.
+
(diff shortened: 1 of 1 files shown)


michaeljoseph:
@stevepiercy While you're here, I'm not convinced that the verbosity of the [Debian installation instructions](https://github.com/audreyr/cookiecutter/blame/master/docs/installation.rst#L46) are justified.  I _think_ we should just be recommending `apt-get install cookiecutter`?

@vincentbernat do you have any thoughts on this? Is there an easier way to install the latest/unstable cookiecutter on a Debian system?

vincentbernat:
Yep, manually installing packages is calling for trouble. You can break other parts of the distribution and it is far possible that you get a non-functional cookiecutter at the end. Also, documentation needs to be kept up-to-date with the list of packages.

For Ubuntu Xenial, we have 1.3. For the future Debian Stretch (to be released in February or March), we will have a 1.5. For Debian, I can also do backports to let people have 1.5 for Debian Jessie. For Ubuntu, I can maintain a PPA. However, this is a bit of work (because dependencies will have to be backported too), notably for Ubuntu. Is there an audience for this?

If the version of cookiecutter in Debian/Ubuntu is not recent enough, people should use pip as a fallback.

michaeljoseph:
Thanks for the insight @vincentbernat- I think those default versions are fine, we should just cater for easily supported installation scenarios.

@stevepiercy would you be up to revising the Debian/Ubuntu installation section to just be:
```
$ sudo apt-get install cookiecutter
```
?

stevepiercy:
Done in https://github.com/audreyr/cookiecutter/pull/880/commits/271c515a472b054cd97b5aa8cbe84b3b39c43aae I hope I understood what you meant. If not, I can fix tonight.

docs/installation.rst
```diff
@@ -45,28 +97,12 @@ Alternate installations
 
 **Debian (Most recentpackaged cookiecutter version):**
```
michaeljoseph: Can we make this heading: `Debian / Ubuntu` please?

docs/installation.rst
```diff
@@ -45,28 +97,12 @@ Alternate installations
...
-    $ sudo dpkg -i python-jinja2-time_0.1.0-1_all.deb
-    $ sudo dpkg -i python-arrow_0.7.0-1_all.deb
-    $ sudo dpkg -i python-whichcraft_0.1.1-1_all.deb
-    $ sudo dpkg -i python-ruamel.yaml_0.11.11-1_amd64.deb # architecture specific
-    $ sudo dpkg -i python-ruamel.ordereddict_0.4.9-1_amd64.deb # architecture specific
-    $ sudo dpkg --configure -a # configures all packages installed above
-    $ cookiecutter --version # Test installation of cookiecutter
-
-.. note:: The debian cookiecutter package is patched version of cookiecutter using a different YAML parser from other installations.
+    $ sudo apt-get install cookiecutter
```
michaeljoseph: 👍

stevepiercy:
Here it is: https://github.com/audreyr/cookiecutter/pull/880/commits/8383d010cc635271b1f0840a7b19e75550648972

michaeljoseph:
Awesome stuff @stevepiercy, thanks :smile_cat:

Issue #35, opened 2013-08-19
Support single-file templates


vfaronov:
This may be out of scope for you, but personally I like the ability to instantiate single files—such as scripts—without any containing directories. As far as I understand, this is currently not possible with Cookiecutter. Could be a magic option in `cookiecutter.json`, like `"__single_file__": true`.

retr0h:
Could someone explain how this can be accomplished?  I have looked through the linked issues/PRs and do not see a reference on how to support single file templates.  It appears this is no longer an issue, but see no good examples.

/cc @michaeljoseph

michaeljoseph:
@retr0h, yeah looks like there was some discussion in #10, but nothing's come out of it.
Re-opening.

talwrii:
Hey, this tag has the duplicate but isn't a duplicate. I would remove the tag but I'm not allowed (only maintainers).

talwrii:
I've added a pull request for a first cut of this feature.

talwrii:
Oops, accidentally self-hooray-emoji'd my own messages while trying to work out what the emoji meant "Hmm, what does that pizza emoji mean?". Please ignore...

machinekoder:
@talwrii Can you please reference the PR.

machinekoder:
Found it: https://github.com/audreyr/cookiecutter/pull/885

machinekoder:
@talwrii I tried to rebase this branch on master: https://github.com/machinekoder/cookiecutter/tree/talwrii--freestanding-files--issue-35--2017-01-06

talwrii:
Cool cool. Thanks for the heads up.

Hmm... I'm not sure what I think about this at the moment. I'm a little unhopeful about it getting it merged. I made the pull request more than a year ago for an outstanding issues with some discussion and tried quite hard to jump through the checklists and be a responsible citizen.

Also I'm rather disinclined to push too hard myself because of behaviour like this from [the project maintainers](https://github.com/audreyr/cookiecutter/issues/743):

"@vaab, this is your first and last warning about Code of Conduct violations against other contributors to Cookiecutter. The item of the Code of Conduct you are specifically violating is: [...]

Any future violations will result in an immediate banning from this project. While we encourage debate, we refuse to condone this kind of ongoing abusive behavior. We wouldn't allow this kind of treatment as paid employees, and we certainly won't take it as volunteers for an unpaid project."

(though perhaps this is a rather extreme example).

[...]

PR #885, opened 2017-01-06
First cut at issue 35: single file templates

talwrii:
Pull request for issue #35 

I imagine people might have quite a lot of opinions about how this should be implemented. But I really want this feature, so I'll throw up a strawman for people to have opinions on.

The approach here is to add a `_target` setting to `cookiecutter.json` which gets rendered to a directory into which files get copied.

Demonstrating I've read the contributing file: 

* Can someone add a please-help tag to this request I don't have the permissions?
* It is definitely possible to implement this outside of cookiecutter... but it's slightly perverse to have a separate utility for expanding a specific type of template.
* I've written tests and run them with tox they pass for py27 and py35. If someone can tell me how to easily install 7 versions of python that would be useful.

Some thorny issues that people might have opinions on:

* How file overwriting should work. I overwrite nothign unless the `-f` flag is given
* Are we actually happy with relative paths
* It's a bit perverse that you have to use a directory name like '{{cookiecutter.source}}' in the template which then gets ignored.

I've written some documentation but was unable to get make the documentation run, `make servedocs` hangs with

```
python -c "$BROWSER_PYSCRIPT" docs/_build/html/index.html
watchmedo shell-command -p '*.rst' -c 'make -C docs html' -R -D .
``` 
and no ports are listened on.

### DIFF
--- a/cookiecutter/exceptions.py
+++ b/cookiecutter/exceptions.py
@@ -75,6 +75,14 @@ class OutputDirExistsException(CookiecutterException):
     """
 
 
+class OutputFileExistsException(CookiecutterException):
+    """
+    Raised when a template file exists already. For example when using _target.
+    """
+    def __init__(self, name):
+        self.name = name
+
+
 class InvalidModeException(CookiecutterException):
     """
     Raised when cookiecutter is called with both `no_input==True` and
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -221,7 +221,7 @@ def render_and_create_dir(dirname, context, output_dir, environment,
         if output_dir_exists:
             msg = 'Error: "{}" directory already exists'.format(dir_to_create)
             raise OutputDirExistsException(msg)
-    else: #pragma: no cover
+    else:  # pragma: no cover
         # This code is unreachable
         raise ValueError(exists_action)
 
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -20,13 +20,19 @@ from .exceptions import (
     ContextDecodingException,
     FailedHookException,
     OutputDirExistsException,
-    UndefinedVariableInTemplate
+    UndefinedVariableInTemplate,
+    OutputFileExistsException,
 )
 from .find import find_template
 from .utils import make_sure_path_exists, work_in, rmtree
 from .hooks import run_hook
 
 logger = logging.getLogger(__name__)
+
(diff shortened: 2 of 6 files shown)


talwrii:
Fixing coverage at the moment.

talwrii:
Okay that should fix the coverage.

Cue argument about "#pragma: no cover" in switch statement...

talwrii:
Okay, I've actually rendered the documentation and looked at it now.

At this point I'm tempted to do a rebase but imagine this might annoy people...

talwrii:
Some thoughts on actually using this anger:

* We have to be very careful about some of the deletion logic here (whoops I deleted your entire repository by mistake). This is more of an issue because we are coexisting with code that people have written
* I found myself wanting to overwrite some code by not others, in my case I created a setup.py (which I wanted to overwrite), but I also created a module in a package directory which I didn't. This is a side effect of the whole "co-existing with real code" thing.

talwrii:
Another potential-complexity / creeping-feature from using this in practice:

* When using different templates in the same project I kept on wanting to have a shared context / replay between the different templates. The use case goes like this:

"Right, lets set up the base project with this cookie cutter"
"Let's add testing like this - hmm I need to re-enter all these details agauin"
"Let's add testing with nose rather py.test - hmm I need add the details again"
"Let's add a license as well"

I don't know what I think about the argument that these should all be in the same template...

sergray:
@talwrii,

1. overwriting only if explicitly requested with a flag should be good enough
1. not sure about ``{{cookiecutter.source}}``, imagine someone would like to have such variable. What about using something like ``{{cookicutter.__tree__}}``?

talwrii:
Thanks for the feedback:

> overwriting only if explicitly requested with a flag should be good enough
This is the current behaviour. 

> not sure about {{cookiecutter.source}}

I have no strong opinions about this. The directory can be called *anything*:{{cookiecutter.source}}, {{cookiecutter.foo}} or {{cookiecutter.__tree__}} all work fine. Using the value for something else has no effect. This implementation kind of naturally falls out of how the code is currently structured. I can throw in an assertion about the value though if constraining this prevents confusion.

turing4ever:
It seems all checks have passed but the code hasn’t been merged?

mmynsted:
Is anybody working on resolving the conflicts?

kurtmckee:

PR #887, opened 2017-01-15
Fix Windows installation instructions

stevepiercy:
- break apart platforms for better context while reading narrative
- Previous instructions were not adequately tested on Windows, but now they are, using a VM.


pydanny:
Sorry for the delay, but thanks for your contribution!

Issue #824, opened 2016-09-14
Cannot use local imports in hooks


henrahmagix:
* Cookiecutter version: `1.4.0`
* Template project url: https://github.com/henrahmagix/cookiecutter-hook-imports
* Python version: `2.7` and `3.4.3` (installed via [`pyenv`](https://github.com/yyuu/pyenv))
* Operating System: OS X 10.11.6 (El Capitan)

### Description:

In `hooks/pre_gen_project.py`, I am trying to import `hooks/utils.py`, but it errors as below.

Admittedly, the amount of code in `hooks/utils.py` isn't very much, and it's not being used in `hooks/post_gen_project.py` (though at first I thought it would need to be), so there's no current advantage to having a separate file. However, it would be nice if local imports were able to be supported, or the documentation explained that they cannot be used.

### What I've run:

```
cookiecutter --no-input gh:henrahmagix/cookiecutter-hook-imports
```
```
Traceback (most recent call last):
  File "/var/folders/fg/d0knzlw97kq5_b7fnfzy59dr0000gn/T/tmp0t3gz8pc.py", line 4, in <module>
    from utils import which
ImportError: No module named 'utils'
ERROR: Stopping generation because pre_gen_project hook script didn't exit sucessfully
Hook script failed (exit status: 1)
```

bittner:
I stumbled upon the same issue.

The reason why local imports don't work is that the "hook" script you write is meant to be executed as a standalone script, because it's actually a _template_ that is [processed by Jinja2](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/hooks.py#L94), [written to a temporary file](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/hooks.py#L95) (with an [arbitrary name](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/hooks.py#L89)) and [then executed](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/hooks.py#L97).

## How Can We Fix This?

To solve this situation, all files from the `hooks` directory would have to be generated (i.e. processed by Jinja2) and executed in the same (temporary) directory.

A positive "side-effect" of this would be that we would retain the filename of the hooks (that can then still be accessed via `__file__`, e.g. for initializing a Python logger). At the moment this is a temporary file name.

michaeljoseph:
We could add the `hooks` directory to the `PYTHONPATH` envvar of the subprocess call?

[...]

PR #892, opened 2017-01-23
Allow Jinja extensions to be defined/bundled in the template repository

ramnes:
This commit permits users to define custom Jinja extensions to be used in the `cookiecutter.json` file.

For example, with something like this in `template/helpers.py`: 
```python
from jinja2.ext import Extension


def get_app_slug(app_name):
    return app_name.lower()


class HelpersExtension(Extension):

        super(Extension, self).__init__()
        environment.filters["get_app_slug"] = get_app_slug

__all__ = ["HelpersExtension"]
```

It allows to use that Jinja filter directly:
```json
{
    "app_name": "ExampleProject",
    "app_slug": "{{cookiecutter.app_name|get_app_slug}}",
    "_extensions": ["helpers.HelpersExtension"]
}
``` 

This is really just a dumb example, but I'm pretty confident that it could be a powerful feature.

I'm kinda new to Cookiecutter, so please point me to the right direction if there's already another way to achieve something like this that I'm not aware of.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -13,6 +13,7 @@ library rather than a script.
 from __future__ import unicode_literals
 import logging
 import os
+import sys
 
 from .config import get_user_config
 from .generate import generate_context, generate_files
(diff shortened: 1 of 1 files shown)


ramnes:
This PR also fixes #824.

ramnes:
I'm closing the PR but will reopen it from another head branch, sorry for the noise.

ramnes:
Now lives in #944

PR #893, opened 2017-01-24
Remove beat-generator from README

ruflin:
For our newest version of the beat generator we use an in house built python script and not cookiecutter anymore.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -504,10 +504,8 @@ Elm
 Golang
 ~~~~~~
 
-* `beat-generator`_: A template to create your own beat to ship data to Elasticsearch.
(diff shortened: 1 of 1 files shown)


hackebrot:
Hi @ruflin! 👋 

Thank you for the update. Sorry to hear you're no longer using cookiecutter.

ruflin:
@hackebrot Cookiecutter was really valuable for us to get where we are now and to understand in detail what we need. Thanks a lot for creating and maintaining this awesome project.

hackebrot:
Thank you for saying that! Means a lot 😊

PR #895, opened 2017-01-29
Group minimal Python projects on top and add the new one

alexkey:
``cookiecutter-lux-python`` cookiecutter was added.


michaeljoseph:
Thanks @alexkey!

PR #896, opened 2017-01-31
Add repo dir to context

### LINKED ISSUE #706: Hooks don't have access to the original template directory
I'm trying to include a configuration summary with the generated projects. The summary would be a JSON file, dumped by `post_gen_project` hook, which would include the cookiecutter context that was used to create the project (same as the files used by `--replay` functionality).

But I'd also like the config summary to include the revision (commit hash) of the used cookiecutter template. This is useful for enabling easy upgrading of the used template, i.e. incorporating the template changes into a project that had been generated from an older version of the template.

Problem is that ATM hooks don't seem to have any way of accessing the template source dir (correct me if I'm wrong). I'd propose to modify `generate_files()` to add some special keys to the context, including (but perhaps not limited to) the original template directory. The new values could either go into a new toplevel dict in the context (next to the `cookiecutter` dict), or maybe into the existing `cookiecutter` object, prefixed with underscores.

Let me know if this is something that would fit into cookiecutter. If yes and we agree on the implementation, I should be able to create a PR.

natetarrh:
Useful for e.g. getting a handle on the template being used in a pre_gen_project hook.

Fixes #706

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -72,18 +72,19 @@ def apply_overwrites_to_context(context, overwrite_context):
             context[variable] = overwrite
 
 
-def generate_context(context_file='cookiecutter.json', default_context=None,
-                     extra_context=None):
+def generate_context(repo_dir, context_file='cookiecutter.json',
+                     default_context=None, extra_context=None):
     """Generate the context for a Cookiecutter project template.
 
     Loads the JSON file as a Python object, with key being the JSON filename.
 
+    :param repo_dir: Project template input directory.
     :param context_file: JSON file containing key/value pairs for populating
         the cookiecutter's variables.
     :param default_context: Dictionary containing config to take into account.
     :param extra_context: Dictionary containing configuration overrides
     """
-    context = {}
+    context = {'_repo_dir': os.path.abspath(repo_dir)}
 
     try:
         with open(context_file) as file_handle:
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -72,6 +72,7 @@ def cookiecutter(
         logger.debug('context_file is {}'.format(context_file))
 
         context = generate_context(
+            repo_dir=repo_dir,
             context_file=context_file,
             default_context=config_dict['default_context'],
             extra_context=extra_context,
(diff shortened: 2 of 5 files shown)


natetarrh:
Here's my desired use case:

```python
#!/usr/bin/python
import os
import subprocess
import sys

cwd = os.getcwd()
os.chdir("{{ _repo_dir }}")
subprocess.call(["git", "fetch"])
local_sha = subprocess.check_output(["git", "rev-parse", "@"])
remote_sha = subprocess.check_output(["git", "rev-parse", "origin/master"])
os.chdir(cwd)
if local_sha != remote_sha:
    print("Not using the most up-to-date template! Please `git pull` and try again")
    sys.exit(1)
```

divick:
+1

cookiecutter/generate.py
```diff
@@ -72,18 +72,19 @@ def apply_overwrites_to_context(context, overwrite_context):
             context[variable] = overwrite
 
 
-def generate_context(context_file='cookiecutter.json', default_context=None,
-                     extra_context=None):
+def generate_context(repo_dir, context_file='cookiecutter.json',
```
theodesp: Why did you add the parameter in the beginning? That could potentially have breaking changes. It's safer if you move it at the end

natetarrh: It's intentional: if the parameter's added at the end, then every callsite has to be updated not just with the value of `repo_dir`, but with values for all the other parameters in between when they're not supplied. This simplifies things, since I can ignore whether other callsites are relying on default values.

I'm not too concerned about breaking changes since there's a single callsite in the source code. This mostly affects the test code.

cookiecutter/generate.py
```diff
@@ -72,18 +72,19 @@ def apply_overwrites_to_context(context, overwrite_context):
...
     Loads the JSON file as a Python object, with key being the JSON filename.
 
+    :param repo_dir: Project template input directory.
     :param context_file: JSON file containing key/value pairs for populating
         the cookiecutter's variables.
     :param default_context: Dictionary containing config to take into account.
     :param extra_context: Dictionary containing configuration overrides
     """
-    context = {}
+    context = {'_repo_dir': os.path.abspath(repo_dir)}
```
michaeljoseph: Could we move this outside the call to `generate_context` rather?

Something like #774 .

ssbarnea:
@natetarrh closing it as it appears abandoned, but ping me if you fix it.

Issue #380, opened 2015-01-30
Remove 'cookiecutter' namespacing


vaab:
I'm not very fond of this 'cookiecutter.' everywhere. I'm aware this might be a controversial subject.
What is the rationale behind it ?

Here's a link to a branch  https://github.com/0k/cookiecutter/tree/pr-remove-namespace (I can't propose it as PR as it's based on another PR #379 of mine).

This change is compatible with the existing namespaced templates.

pydanny:
The namespacing there is to make it obvious that a particular template was a cookiecutter one (there are many project template renderers in the world). Additionally, if you have a project that uses Jinja2, Django, Mustache, Liquid, or a hundred other template engines, not having the namespace makes it really hard to figure out what's going on. That said, We've discussed the ability to have a mutable namespace. So you could shorten it to something like 'cc'.

As to your implementation, the problem with it is that it doesn't just break backward's compatibility, it shatters it. ;)

This isn't a simple issue. Let's discuss it.

vaab:
About attributing the package to cookiecutter, I figured that it would be enough to have a cookiecutter-* named repository/folder, and a cookiecutter.{json,yml} at the base.
For the second point, I thought that if you had to generate any other template files for your project you would have to resort to somesort of quoting anyway.
And for the last point, can you explain what did I shatter that the actual current test suite didn't see ? (The only thing I can see is the usage of a template variable named "cookiecutter" that is not anymore available for template maker).
I had it renamed to "cc", but it seemed to me that if I had to put it everywhere, this was a good sign I could remove it. Are there some cases I had missed where you could actually not prefix your variables in the current usage ?
Many thanks for any of your explanations. I still feel that I miss some important reason why this namespace is here, and I'm confident that you'll make it clear.

pydanny:
@vaab Here we go!

1. You are quite possibly correct about the `cookiecutter.{json,yml}` file at the base.
2. It's common to use the `{% raw %}` tag from Jinja to manage what should and shouldn't be rendered. Even then, it gets confusing even with the `cookiecutter` namespace. 
3. By breaking backwards compatibility, I don't mean the tests. Specifically I mean the 250-odd public cookiecutter templates on GitHub, then dozens on Bitbucket and GitLab, not to mention the potential of hundreds that are used privately by companies (Eventbrite, Spotify, Dreamhost, etc are users) and individuals. Any change to the name space has to be done in a way to protect those users from having to revamp all their hard work.

[...]

PR #908, opened 2017-03-06
Return generated context from cookiecutter API

rmedaer:
I patched cookiecutter main function to return both result of `generate_files` function and generated context. Indeed the generated context is really useful for my application which using cookiecutter as Python module.

An example of usage: https://github.com/rmedaer/milhoja/blob/b1207e79/milhoja/milhoja.py#L178

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -89,4 +89,4 @@ def cookiecutter(
         context=context,
         overwrite_if_exists=overwrite_if_exists,
         output_dir=output_dir
-    )
+    ), context['cookiecutter']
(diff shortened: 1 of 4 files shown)


hackebrot:
Hi @rmedaer! 👋 

Thank you for this PR. I'm afraid, we will need to wait for the next major release to integrate this change, as it's breaking an existing API.

As you a workaround you can include a file in your template to store the context as a python dict with ``{{ cookiecutter }}``.

theodesp:
@rmedaer Personally I don't particularly like this approach of returning 2 variables instead of one. How about adding a new CLI call to just provide you the context that was resolved?

michaeljoseph:
Another workaround would be to read the context from the [`replay_dir`](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/main.py#L84)?

insspb:
Hi @rmedaer, thank you for contribution. And excuse us for long delay with answer
Please check #380 discussion. The resulting patch for #380 will make 'cookiecutter' a variable name. 
In this PR you are returning only part of context, not whole item. 
To be reusable such thing should return whole item.

kurtmckee:

PR #910, opened 2017-03-10
pyramid-cookiecutter-starter add Chameleon choice

stevepiercy:
We've added an option to pyramid-cookiecutter-starter for the user to select Chameleon for templating language in addition to Jinja2


stevepiercy:
Superceded by PR #911

PR #912, opened 2017-03-14
Fixes broken link to "Copy without render"

coreysnyder04:
I ran into a broken link and this should fix it.


hackebrot:
Oops! Thank you @coreysnyder04! 🙇

Issue #846, opened 2016-11-02
git pull instead of re-clone


ryanolson:
* Cookiecutter version: latest
* Template project url: n/a
* Python version: n/a
* Operating System: n/a

### Description:

Instead of deleting a git repo that has already been pulled, it seems most logical that we should perform a pull instead of deleting the repo and recloning.

https://github.com/audreyr/cookiecutter/blob/e041f215e5def801529a67af03f0bcf283af38aa/cookiecutter/vcs.py#L28

If the pull fails, then a wipe and reload is in order.

pydanny:
Interesting idea, certainly could speed things up with Cookiecutter Django.

The downside is the additional complexity for Cookiecutter.

ryanolson:
I know you guys don't like to add a lot of external dependencies; however you do require `git` to be installed, perhaps it's not too much to  add [pygit2](https://github.com/libgit2/pygit2) to the set of dependencies.  That greatly simply the `vcs.py` code and make pulling much easier.

pydanny:
Thanks for the other suggestion, but the installation issues across multiple platforms are a show stopper for Cookiecutter integration (http://www.pygit2.org/install.html#libgit2-within-a-virtual-environment)

ryanolson:
Gotcha.  If I have time, I'll see about adding some submodule calls for performing a pull.

pydanny:
Umm... Pygit has dependency issues that are way too broad for the Cookiecutter maintainers to cover. Any simplification of code that it will provide will be offset many times over by users across many platforms asking for help with libgit installation help.

kurtmckee:
It may be possible to implement this using dulwich nowadays, but this represents additional complexity for cookiecutter, and it's outside of what the team can write and maintain at this time.

I'm closing this ticket due to the feature's complexity and maintenance effort.

PR #914, opened 2017-03-14
If repo exist, try to pull before recloning it.

rlaverde:
Fixes: #846

This is missing a test, I'm not sure how this could be tested, someone could give a hint?

### DIFF
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -39,6 +39,7 @@ def prompt_and_delete_repo(repo_dir, no_input=False):
     else:
         question = (
             "You've cloned {} before. "
+            "And It wasn't possible to pull new changes. "
             "Is it okay to delete and re-clone it?"
         ).format(repo_dir)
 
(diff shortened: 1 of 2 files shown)


hackebrot:
Hi @rlaverde! 👋 

Thank you for your contribution! I wonder if we should prompt the user whether they want to perform a git pull first. Otherwise we overwrite their local template and they won't even know.

What do you think, @audreyr @pydanny @michaeljoseph?

rlaverde:
> I wonder if we should prompt the user whether they want to perform a git pull first. Otherwise we overwrite their local template and they won't even know.

I'm not sure, I didn't add a prompt because I think that normally people will want to use the latest version of the cookie(template)

hackebrot:
I'm not sure, right now as a user you can decide not to modify the local template when using the CLI. With this change we take that option away from them.

🔔 @michaeljoseph

michaeljoseph:
Yeah, I think I'm with @hackebrot on this one.  It would make more sense to me to add this functionality when the user declines deletion and re-cloning, since we currently just exit (which has always seemed surprising).
I think we should ask the user if they'd like to `pull`, and then use the existing repo if they decline.

cookiecutter/vcs.py
```diff
@@ -39,6 +39,7 @@ def prompt_and_delete_repo(repo_dir, no_input=False):
     else:
         question = (
             "You've cloned {} before. "
+            "And It wasn't possible to pull new changes. "
```
theodesp: Lower case `It` here --> `it'

tests/vcs/test_clone.py
```diff
@@ -73,10 +73,6 @@ def test_clone_should_abort_if_user_does_not_want_to_reclone(mocker, tmpdir):
         side_effect=SystemExit,
         autospec=True
     )
```
theodesp: I would like more tests here covering those scenarios you mentioned

theodesp:
@rlaverde You need also to fix the CI errors 👍

PR #916, opened 2017-03-19
Adds command "cookiecutter installed" to list installed (locally available) templates.

mikeckennedy:
Adds new command: `cookiecutter installed`

This lists locally available templates, for example:

```
user$ cookiecutter installed
7 installed templates: 
 * cookiecutter-bottle
 * cookiecutter-pyramid-talk-python-starter
 * cookiecutter-sublime-text-3-plugin
 * cookiecutter-template
 * puppet-openstack-cookiecutter
 * pyramid-cookiecutter-starter
 * python-macon-template
```

I was not able to integrate it into click without breaking the existing functionality. Someone who knows click better than me please adjust this if possible.

I’d like to see just `cookiecutter —installed` as the command (so it would be listed in help output) but I only get a missing template error and all attempts to add the command as an option met with my lack of experience with click.

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -15,6 +15,7 @@ import click
 
 from cookiecutter import __version__
 from cookiecutter.log import configure_logger
+from cookiecutter.config import get_user_config, get_config, DEFAULT_CONFIG
 from cookiecutter.main import cookiecutter
 from cookiecutter.exceptions import (
     OutputDirExistsException,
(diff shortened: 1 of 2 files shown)


hackebrot:
Hey @mikeckennedy! 👋 

Thanks for the PR. I'll have a look into the CLI implementation.

What does the ``-27`` exit code mean and is that platform agnostic?

mikeckennedy:
Thanks @hackebrot that would be great to see this feature.

-27 is just a number to be unique. Feel free to use any failure code that makes the most sense.

cookiecutter/cli.py
```diff
@@ -106,6 +134,9 @@ def main(
     if template == u'help':
         click.echo(click.get_current_context().get_help())
         sys.exit(0)
+    if template == u'installed' or template == u'i':
```
theodesp: I think it's better if you name it --list-installed as its is more accurate

cookiecutter/cli.py
```diff
@@ -48,6 +49,33 @@ def validate_extra_context(ctx, param, value):
...
+    Lists installed (locally cloned) templates. Use cookiecutter installed
+    """
+
+    config = get_user_config(passed_config_file, default_config)
+    cookiecutter_folder = config.get('cookiecutters_dir')
+
+    if not os.path.exists(cookiecutter_folder):
+        click.echo('Error: Cannot list installed templates. ' +
+                   'Folder does not exist: {}'.format(cookiecutter_folder))
+        sys.exit(-27)
```
theodesp: Why don't you just return `sys.exit(-1)` it's more ordinary that way

cookiecutter/cli.py
mikeckennedy: Sounds great to me. I'm open to basically any name for the command. But it's a real pain to hunt down the template names over and over. Once you have them downloaded it would be nice to use it as a list of what you've used before.

cookiecutter/cli.py
mikeckennedy: I simply wanted a different code than other errors so things scripting cookiecutter can use that as a signal for why / how something failed. If everywhere you return -1 how do you know what went wrong?

theodesp:
@mikeckennedy It looks like you need to add a few more tests to maintain the coverage 👍

cookiecutter/cli.py
theodesp: Ok, But I think this is a separate consideration to handle as in the project for now exit returns with -1 or 0. Maybe log a ticket to have this extra functionality.

mikeckennedy: Sure. For the sake of this PR I'm happy to make it -1. But this could be something to consider in general for cookiecutter.

PR #883, opened 2017-01-05
Add cookiecutter-bare to README.rst

tuxredux:
I added [cookiecutter-bare](https://github.com/webevllc/cookiecutter-bare) to the **Cookiecutter (Meta)** category, because it is a _programming language agnostic_ project. Decline this pull request, and let me know if I should create a new category instead! Thanks in advance for your time!

PR #921, opened 2017-03-25
Add cookiecutter-git to README

tuxredux:
This is a follow-up of PR #883 ! I renamed the repository, so I thought it would be best to create a new pull request (also as a way to reduce commit history bloat). I added bitbucket.org support like I said I would, and I also created a few official releases. I also named and added the **Git** category, as per @michaeljoseph 's suggestion. Thank you in advance for your time! :smile:


michaeljoseph:
Thanks @tuxredux!

PR #923, opened 2017-03-28
[WIP][add] ability to override render context before rendering.

simahawk:
A common requirement - at least in our projects - is to generate new variables on the fly based on user input.
Now you can use `get_render_context` to generate the context and manipulate it before passing it to cookiecutter.
Example:
```
from cookiecutter.main import cookiecutter, get_render_context
from cookiecutter.config import get_user_config

config_dict = get_user_config()
context_file = os.path.join(template_path, 'cookiecutter.json')
# this triggers questions prompt
render_context = get_render_context(context_file, config_dict)
render_context['cookiecutter']['whatever_you_want_to_render'] = 'oh yeah!'
cookiecutter(template_path, render_context=render_context)
```
TODO: update tests and docs.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -24,7 +24,7 @@ logger = logging.getLogger(__name__)
 def cookiecutter(
         template, checkout=None, no_input=False, extra_context=None,
         replay=False, overwrite_if_exists=False, output_dir='.',
-        config_file=None, default_config=False):
+        config_file=None, default_config=False, render_context=None):
     """
     API equivalent to using Cookiecutter at the command line.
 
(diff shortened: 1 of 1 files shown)


ericof:
@simahawk Do you a use case for that? (If it is the same as mine we could join forces)

simahawk:
@ericof hey pal :) Use case is mostly the one described in the description :stuck_out_tongue_winking_eye:  You want to play with variables at runtime based on user input. 

For instance, we use it to generate our cloud platform project settings by computing secure passwords, random database names,  memory size, workers number, etc.

Is not published but I can give you a private picture of it if you need it :wink:

ericof:
@simahawk I have the same user case on Websauna, so let's work together on this.
It seems you need the tests and docs, correct?

simahawk:
@ericof that would be awesome :)

ssbarnea:

PR #929, opened 2017-04-13
Don't dereference symlinks when using copytree

bittner:
This change addresses issues with dereferencing symlinks. See https://github.com/audreyr/cookiecutter/issues/865#issuecomment-293891002 for details.


pjbull:
FYI, there is an existing PR for the referenced issue already that fixes this, adds tests, handles rendered and non-rendered symlinks, etc:
https://github.com/audreyr/cookiecutter/pull/868

That one is waiting on review--presumably because the Windows story is so rough.

bittner:
Fascinating, symlinks on Windows. Is there such a thing? Glad I don't need to bother.

I'm happy to close this PR again.

PR #932, opened 2017-04-21
Implement pep257 on exceptions.py

terryjbates:


cookiecutter/exceptions.py
```diff
@@ -1,103 +1,77 @@
...
+    """Raised when a template uses an undefined variable in the context."""
+
     def __init__(self, message, error, context):
+        """Initialize UndefinedVariableInTemplate instance."""
         self.message = message
         self.error = error
         self.context = context
 
     def __str__(self):
+        """Return informal string representation of object."""
```
hackebrot: ``informal``?

cookiecutter/exceptions.py
```diff
@@ -1,103 +1,77 @@
...
     """
 
 
 class UnknownTemplateDirException(CookiecutterException):
-    """
-    Raised when Cookiecutter cannot determine which directory is the project
-    template, e.g. more than one dir appears to be a template dir.
+    """Raised when project template directory cannot be determined.
+
+    Example: more than one dir appears to be a template dir.
```
hackebrot: I think I like the original message better.

cookiecutter/exceptions.py
```diff
@@ -1,103 +1,77 @@
...
-    """
+    """Raised when get_config() is passed an invalid path to a config file."""
 
 
 class InvalidConfiguration(CookiecutterException):
-    """
-    Raised if the global configuration file is not valid YAML or is
-    badly contructed.
-    """
+    """Raised if global configuration file is invalid or badly constructed."""
```
hackebrot: I think ``invalid`` is not as explicit as ``not valid YAML``

PR #934, opened 2017-04-23
Add support for symlinks

pjbull:
~~This is in place of PR #868, which also include fixes for Windows that seem too messy to be worth pulling in to maintain.~~

~~Caveat: Symlinks will _not_ work as expected on Windows with Python versions < 3.2 (but, they don't now anyway...)~~

Updated PR to support symlinks in templates across platforms now that we don't support older versions of Python.

### DIFF
--- a/AUTHORS.md
+++ b/AUTHORS.md
@@ -170,3 +170,4 @@
 * Tom Forbes ([@orf](https://github.com/orf))
 * Xie Yanbo ([@xyb](https://github.com/xyb))
 * Maxim Ivanov ([@ivanovmg](https://github.com/ivanovmg))
+* Peter Bull ([@pjbull](https://github.com/pjbull))
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -228,10 +228,9 @@ def render_and_create_dir(
         link_tmpl = environment.from_string(symlink)
         rendered_link = link_tmpl.render(**context)
 
-        logger.debug('Creating symlink from {} to {}'.format(
-            dir_to_create,
-            rendered_link
-        ))
+        logger.debug(
+            'Creating symlink from {} to {}'.format(dir_to_create, rendered_link)
+        )
 
         os.symlink(rendered_link, dir_to_create)
     else:
@@ -366,7 +365,12 @@ def generate_files(
                 unrendered_dir = os.path.join(project_dir, root, d)
                 try:
                     render_and_create_dir(
-                        unrendered_dir, context, output_dir, env, overwrite_if_exists, symlink=symlinks.get(d, None)
+                        unrendered_dir,
+                        context,
+                        output_dir,
+                        env,
+                        overwrite_if_exists,
+                        symlink=symlinks.get(d, None),
                     )
                 except UndefinedError as err:
                     if delete_project_on_failure:
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -354,6 +354,7 @@ def generate_files(
 
             for copy_dir in copy_dirs:
                 indir = os.path.normpath(os.path.join(root, copy_dir))
+                outdir = os.path.normpath(os.path.join(project_dir, indir))
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
                 shutil.copytree(indir, outdir, symlinks=True)
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -218,11 +218,14 @@ def render_and_create_dir(
             logger.debug(
                 'Output directory %s already exists, overwriting it', dir_to_create
             )
+
+            # must be removed for symlink to be created successfully
+            if symlink is not None:
+                shutil.rmtree(dir_to_create)
+
         else:
             msg = 'Error: "{}" directory already exists'.format(dir_to_create)
             raise OutputDirExistsException(msg)
-    else:
-        make_sure_path_exists(dir_to_create)
 
     if symlink is not None:
         link_tmpl = environment.from_string(symlink)
@@ -357,7 +360,9 @@ def generate_files(
                 outdir = os.path.normpath(os.path.join(project_dir, indir))
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
-                shutil.copytree(indir, outdir, symlinks=True)
+                shutil.copytree(
+                    indir, outdir, symlinks=True, dirs_exist_ok=overwrite_if_exists
+                )
 
             # We mutate ``dirs``, because we only want to go through these dirs
             # recursively
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -361,7 +361,7 @@ def generate_files(
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
                 shutil.copytree(
-                    indir, outdir, symlinks=True, dirs_exist_ok=overwrite_if_exists
+                    indir, outdir, symlinks=True,
                 )
 
             # We mutate ``dirs``, because we only want to go through these dirs
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -199,7 +199,7 @@ def generate_file(project_dir, infile, context, env, skip_if_file_exists=False):
 
 
 def render_and_create_dir(
-    dirname, context, output_dir, environment, overwrite_if_exists=False
+    dirname, context, output_dir, environment, overwrite_if_exists=False, symlink=None
 ):
     """Render name of a directory, create the directory, return its path."""
     name_tmpl = environment.from_string(dirname)
@@ -221,6 +221,19 @@ def render_and_create_dir(
         else:
             msg = 'Error: "{}" directory already exists'.format(dir_to_create)
             raise OutputDirExistsException(msg)
+    else:
+        make_sure_path_exists(dir_to_create)
+
+    if symlink is not None:
+        link_tmpl = environment.from_string(symlink)
+        rendered_link = link_tmpl.render(**context)
+
+        logger.debug('Creating symlink from {} to {}'.format(
+            dir_to_create,
+            rendered_link
+        ))
+
+        os.symlink(rendered_link, dir_to_create)
     else:
         make_sure_path_exists(dir_to_create)
 
@@ -323,9 +336,15 @@ def generate_files(
             # unrendered directories, since they will just be copied.
             copy_dirs = []
             render_dirs = []
+            symlinks = dict()
 
             for d in dirs:
                 d_ = os.path.normpath(os.path.join(root, d))
+
+                if os.path.islink(d_):
+                    logger.debug('Processing symlink at {}...'.format(d))
+                    symlinks[d] = os.readlink(d_)
+
                 # We check the full path, because that's how it can be
                 # specified in the ``_copy_without_render`` setting, but
                 # we store just the dir name
@@ -336,10 +355,9 @@ def generate_files(
 
             for copy_dir in copy_dirs:
                 indir = os.path.normpath(os.path.join(root, copy_dir))
-                outdir = os.path.normpath(os.path.join(project_dir, indir))
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
-                shutil.copytree(indir, outdir)
+                shutil.copytree(indir, outdir, symlinks=True)
 
             # We mutate ``dirs``, because we only want to go through these dirs
             # recursively
@@ -348,7 +366,7 @@ def generate_files(
                 unrendered_dir = os.path.join(project_dir, root, d)
                 try:
                     render_and_create_dir(
-                        unrendered_dir, context, output_dir, env, overwrite_if_exists
+                        unrendered_dir, context, output_dir, env, overwrite_if_exists, symlink=symlinks.get(d, None)
                     )
                 except UndefinedError as err:
                     if delete_project_on_failure:
--- a/docs/advanced/index.rst
+++ b/docs/advanced/index.rst
@@ -23,3 +23,4 @@ Various advanced topics regarding cookiecutter usage.
    template_extensions
    directories
    new_line_characters
+   symlinks
--- a/docs/advanced/symlinks.rst
+++ b/docs/advanced/symlinks.rst
@@ -1,0 +1,58 @@
+.. _symlinks:
+
+Symlinks
+----------------------
+
+Symlinks are virtual files or folders that simply point to another location on the
+file system. For example, you may add a symlink to your project template ``system_logs``
+that points to ``/var/log`` on the system where the template is rendered. Symlinks can
+be thought of as a shortcut to a specific file or folder.
+
+Symlinks will work on most platforms with cookiecutter. However, if you expect your
+template to be used on Windows systems with Python < 3.2, it is recommended that you
+do not use symlinks or that you handle symlinking in a post-generation hook.
+
+
+Using symlinks
+~~~~~~~~~~~~~~~~~~~~~~~
+
+`Symbolic links`_ are commonly used in posix systems. Cookiecutter supports symlinks
+in templates both as rendered and unrendered content. That is, the symlink itself
+both be named with a variable or point to a destination with a variable in it.
+
+On posix systems (see below for Windows systems) you can simply create a symbolic link
+as normal in the template directory::
+
+    ln -s existing_file_path new_symlink_path
+
+As stated above, either ``existing_file_path`` or ``new_symlink_path`` can contain
+variables that will be templated in braces ``{{ cookiecutter.variable }}``. These
+will be replaced when the template is rendered.
+
+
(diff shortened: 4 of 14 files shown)


bittner:
Build fails on Travis for three easy-to-fix `flake8` complaints in `tests/test_generate_symlinks.py`.

On AppVeyor it's more weird: All Python 3 tests fail.

pjbull:
Yeah, I know why this is happening. Should have a fix shortly.

pydanny:
I like that this PR is simpler than #868. On the other hand, it's always tricky when we have a feature that works in one operating system and not another that doesn't throw warnings or exceptions to users. Considering our supported versions of Python in https://github.com/audreyr/cookiecutter/blob/master/setup.py, how about this:

If:

* on Windows
* running Python 3.2 or lower
* symlinks are attempted

Then Cookiecutter should throw an explicit error called something like: `NoSymlinksOnWindowsPython27` and `NoSymlinksOnWindowsPython32`, followed by a concise but understandable explanation. That will allow us (the maintainers) to have an easy resolution when someone files an issue for a Cookiecutter template taking advantage of symlinking.

pydanny:
Also, @pjbull and @bittner, is there any reason why you want Cookiecutter to handle symlinks in the templating section? Why not just write a post gen hook? Seems to me that the better way to do things - you certainly have a lot more fine-grained control!

bittner:
I'm not sure I understand @pydanny's intent in "why you want Cookiecutter to handle symlinks in the templating section".

I assume reproducing a cookiecutter template "as is" is meant to be the default behavior of the Cookiecutter tool. Why exclude symlinks? Isn't the current behavior a bug? (Imagine a relative symlink pointing to the current version of, say, a logfile or the active configurations of Apache.)

Cookiecutter should reproduce. Period.
--------------------------

It should be the responsibility of the cookiecutter template author to ensure the cookiecutter (template, not the tool) works across different environments and operating systems (I [commented on this in PR #868](https://github.com/audreyr/cookiecutter/pull/868#issuecomment-294288951) already). If there is a symlink in a cookiecutter I should know as an author that this may only work on Unix-like OSes. Cookiecutter should create a symlink, not a copy of real file, if I put a symlink under version control.

(Interesting, it looks like symlinks are [supported on Windows since Vista](https://blogs.windows.com/buildingapps/2016/12/02/symlinks-windows-10/), but has anyone seen  anyone using that feature?)

pjbull:
There's a hiccup to raising an error on an unsupported platform if we see a symlink: the `islink` function doesn't work, so we'll never know that we're in an unsupported case. I think the best we can do is have a warning on the platform, and maybe in the documentation.

Would this PR work for you if we add:
 - [ ] `warning` in main.py if symlinks are not supported
 - [ ] Add documentation about symlink support to the docs (including which OS/Python combos are supported)

In my mind, the justification for symlink support in `cookiecutter` is that it is a filesystem feature that some people just expect to work (right now it fails silently). We see it more often in the data science use cases that we work on where you want to consistently point collaborators to a data source that's mounted elsewhere. It's not the only way to do that, but it is a reasonable way.

The more aggressive approach is to use the bigger PR in #868, trust in the tests, and support symlinks across platforms. Hopefully some day that all gets deprecated when official support goes to > Python 3.2 only.

The more conservative approach is just to have a `warning` if we see a symlink that symlinks are not supported and authors of templates should handle linking in a post gen hook.

My preference in is the middle-of-the-road option, but I can see justification for the more conservative/aggressive approaches as well.

pfmoore:
> warning in main.py if symlinks are not supported (this would always pop-up, regardless of if symlinks are attempted)

So you're suggesting that cookiecutter would warn every time it's used on Windows? Regardless of whether symlinks (a feature that's presumably not used in any existing cookiecutter) are used in the cookiecutter the user requests?

That seems pretty hostile to Windows users :-(

pjbull:
@pfmoore Yes, it would warn on Windows + Python < 3.2 (so, not all Windows instances).

I definitely don't want to be hostile to Windows users (for proof, see #868). I'm open to other solutions if you have a suggestion.

pfmoore:
Just fail if the template tries to use a symlink on a platform that doesn't support them? You say islink doesn't work if symlinks aren't supported - I don't follow. If you git clone a cookiecutter with a symlink in it, what would you get on a platform without symlinks? I'd assume you'd get a warning/error from git that the clone didn't work as expected - why not catch that and report it to the user? You certainly wouldn't get a symlink in the checkout (It's worth noting that on Windows, even in Win 10, users can only create symlinks if they have a specific privilege - in practical terms, that means you're in an elevated prompt, which people won't be 99% of the time, so "you can't realistically use symlinks on Windows" is a pretty good rule of thumb).

If you have a public repo with symlinks, I'd be happy to test by doing a clone of that repo and reporting the results back here.

bittner:
I think it would help to outsource the `win_utils.py` module to a compatibility project on PyPI that handles `islink`, `readlink`, `symlink`, `copytree` transparently. This can be managed with additional care outside the scope of this project.

We could then simply go ahead with this PR, and when the compatibility project is ready we simply change the imports from `from os import ...` and `from os.path import ...` to, say, `from os_path_compat import ...`, and we're done. (I would also ask @jaraco before creating that compatibility project, maybe he or someone else knows if such a project already exists.)

pydanny:
> I think it would help to outsource the win_utils.py module to a compatibility project on PyPI that handles islink, readlink, symlink, copytree transparently. This can be managed with additional care outside the scope of this project.

That would be pretty awesome. I did something similar to that with https://github.com/pydanny/whichcraft, which was me pulling out the code of Cookiecutter so it could be easily tested independently. I know it seems like a tiny, silly package, but moving this stuff out means the problem becomes more shallow and easier for more people to contribute to fixing issues.

bittner:
Oooops, this project seems to already exist!  :tada: 

- `pip install`[jaraco.windows](https://pypi.python.org/pypi/jaraco.windows)
- e.g. [from jaraco.windows.filesystem import symlink](https://github.com/jaraco/jaraco.windows/blob/master/jaraco/windows/filesystem/__init__.py)

Found on StackOverflow for [os.path.islink on windows with python](http://stackoverflow.com/questions/15258506/os-path-islink-on-windows-with-python).

pjbull:
@bittner As I wrote in #868, the existing `jaraco.windows` functions do not work as expected. If you just use those imports, the symlink tests fail. There is an error when rmtree tries to remove the directories. That's the reason `win_utils` got rolled in the first place. Before creating the posix-only PR, I spent a solid half day in the weeds on finding a workaround that used `jaraco.windows` and still passed the tests. Here's a branch to work off of if you want to test that out:
https://github.com/pjbull/cookiecutter/tree/win-symlink-jaraco

@pfmoore Pretty sure that the default behavior is just to produce a text file with the path that is supposed to be linked to if you clone a git repo with symlinks.  Here are their docs for that: https://github.com/git-for-windows/git/wiki/Symbolic-Links Also, not sure that relying on `git` errors/warnings would work anyway since there's no guarantee a template is cloned with`git`, right?

@pydanny Yeah, the purpose of this new PR was just to extract the functionality that we can use now and then solve the Windows symlinks problem elsewhere. Maybe that's a new package, maybe there's a fix for a bug `jaraco.windows`. Either way, I guess the question is what is the stop gap that let's posix folks use symlinks while the Windows work is in progress? Is it a warning? Is it a docs change? Happy to do whatever unblocks those concerns so this could be merged.

bittner:
For me merging this PR is fine, my use case is only affected by [a single line](https://github.com/audreyr/cookiecutter/pull/934/files#diff-b4c1da8a76ea242bf99a397dcb90b2a6R332) though, to be honest.

``` python
shutil.copytree(indir, outdir, symlinks=True)
```

My problem, if that is not yet clear, is that Cookiecutter does not create what I put under version control. A postgen hook would be pointless (makes a special case that needs postprocessing for every single symlink put under version control, or requires moving file system content into postgen hook business logic, if you want so).

pfmoore:
@pjbull OK, I'm somewhat confused here. AIUI, cookiecutter creates a directory from a template that's copied to the user's machine in ```~/.cookiecutters```. If I'm not able to create symlinks (non-elevated, no special privs enables) then that template won't have symlinks in there. So by the time we're creating the target directory from the template it's already too late.

So how would I run cookiecutter on a template that contains symlinks? I don't think I can even create one...

FWIW, ```git clone https://github.com/sschuberth/git-playground``` on my PC gives no errors, and creates a directory ````linux-links``` with 3 text files:

```
>dir

    Directory: C:\Work\Scratch\git-playground\linux-links

Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       25/04/2017     22:52             14 this_is_a_hard_link_to_file
-a----       25/04/2017     22:52             14 this_is_a_regular_file
-a----       25/04/2017     22:52             22 this_is_a_symbolic_link_to_file

>type .\this_is_a_hard_link_to_file
Hello world!
>type .\this_is_a_regular_file
Hello world!
>type .\this_is_a_symbolic_link_to_file
this_is_a_regular_file
```

Anyway, this is pretty off-topic. My only point was that I don't think it's reasonable for cookiecutter to always warn the user that symlinks aren't supported, without even checking if it matters for what the user's actually doing.

pjbull:
@pfmoore Changing your git config enables symlinks on Windows `git config --global core.symlinks true`, so you can have a template with symlinks that you cloned from a git repository (which is what this PR does on Appveyor). Also, for the record, Python supports creating, reading, and testing for symlinks on Windows in versions >= 3.2, so it's not crazy to think users could have symlinks on their Windows systems.

Agreed that it is not ideal to have super chatty warnings about symlinks if users aren't using them. My only point is _if_ a user cloned a git repo that did have symlinks by changing their config, we still can't warn in `cookiecutter` because `islink` will be False for those links when Python < 3.2 (even though they are actually symlinks on the filesystem).

I'd like to do the right thing for Windows users and posix users. Unfortunately, it just doesn't seem like there's an easy Windows + Py3.2 fix at the moment. My current vote would be to add a section in the docs about symlinks that includes information on the git config and the fact that they won't work on Windows systems with Python < 3.2. Happy to write that up if that's the preferred fix.

pfmoore:
@pjbull My point is that you don't *just* need the git setting, you also need some fairly annoying to set privileges. But yes, if you *have* set things up so you can use symlinks, why not have them work?

If it's just Python 3.2 that's the problem (sorry, hadn't spotted that) then it's not a big deal. Very few people even bother supporting 3.2 any more, so limited functionality seems reasonable in that case. One thought - would it be possible to patch the symlink functions to immediately give an error on unsupported systems? Then you could ensure that you fail if the codepath that uses symlinks is taken, without affecting other codepaths.

But a doc section saying "unsupported on Windows with Python < 3.2" seems like a sufficient approach.

pjbull:
@pfmoore It's actually all Python versions < 3.2 (including all of Python 2, including 2.7). I probably had admin privileges on my Azure VMs, but I never needed any changes beyond the git config to use symlinks in my testing.

I think we can patch `readlink`/`symlink` if we also take the `jaraco.windows` dependency and use the `islink` function from there (otherwise `islink` is always false and we'll never hit read or create). That may actually be good since it makes it easier for someone to figure out the bugs that were preventing `jaraco.windows` just working for our use case (as mentioned above).

So, just to make sure that we're on the same page after the chatter:
 - [ ] Docs update with specific info about symlinks
 - [ ] On Windows + Py < 3.2, use the `jaraco.windows` version of `islink`. Patch `readlink` and `symlink` to raise `NoSymlinksOnWindowsPython27` with a clear message that symlinks are not supported in that configuration.
 - [ ] Add bug for fixing symlinks on Windows + Python < 3.2 (the fix may ultimately be in `jaraco.windows`, but the current failure I see is in our `rmtree` function so it needs a little more investigation) and dump all this info in there.

pfmoore:
@pjbull Sounds reasonable to me.

jaraco:
> the existing jaraco.windows functions do not work as expected. ... There is an error when rmtree tries to remove the directories.

I'd be delighted to see a bug report and if possible a PR to test and fix this issue with [jaraco.windows](/jaraco/jaraco.windows). I'll get it set up on AppVeyor for CI support.

bittner:
I'd be happy to see this merged anytime soon, as discussed above. :+1:

pjbull:
I've completed all the tasks above including docs and a new issue. 

#947 now tracks implementing symlinks for Windows + Python < 3.2. I'll update that bug with a couple of relevant links once this is merged.

I'll also close the following:
 - #868 
 - #865

bittner:
Is this PR ready for merging?

theodesp:
@pjbull Can you review the conflicting files?

milancermak:
A project I'm working on would benefit from this functionality. Is there anything I can do to help with moving this forward?

pydanny:
Yes, you can expedite the review of this and other pull requests by getting your employer to increase our funding:

* https://www.patreon.com/hackebrot ($39/month)
* https://www.patreon.com/danielroygreenfeld ($2/month)

For reference, Cookiecutter works in a stable manner on all operating systems and supported versions of Python. Our plan is to maintain the status quo moving forward unless we can get more funding. Adding more features increases the workload of the maintainers, who are fundamentally unpaid for working on this project.

milancermak:
All clear. I've became a small patreon myself, will talk to the folks at my company to see if we can increase that. Thanks to all of you!

pydanny:
@milancermak I'm very sorry about what I posted in another place. It was inexcusable and I apologize. I've deleted what I said.

pydanny:
We're still having failures on testing on Windows with Python 3.x. I'm working to resolve that right now.

alexz-kh:
Bump for cool feature :)

alexz-kh:
One more bump for cool feature :|

@pydanny @milancermak @audreyr  @theodesp any chance to get into it ?

ssbarnea:

bittner:
This would have fixed an actual bug for anyone using symlinks in a cookiecutter template. Sad to see this go.

SamSchott:
Is there any interest on picking this up again? Symlinks can be an essential part of a project structure in Unix and being able to preserve those from a cookiecutter template would be great.

One example use case are templates which contain macOS Frameworks.

bittner:
> Is there any interest on picking this up again? Symlinks can be an essential part of a project structure in Unix and being able to preserve those from a cookiecutter template would be great.

This issue should definitely be picked up again. IMHO, fixing this bug is not about taste or personal preference. A file system-based template should reproduce the exact original state that it was drafted from.

The fact that symlinks need to be [handled in a special manner](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/mklink) on Mircosoft Windos should be regarded as a technical detail, not a go/no-go style decision. It could be handled as a (temporarily) unsupported feature to start with ([as suggested](https://github.com/cookiecutter/cookiecutter/pull/934#issuecomment-296767763) by @pydanny above).

pjbull:
What's the end-of-life story for Python 2.7 and cookiecutter? IIRC, once support for 2.7 is officially dropped, this PR can be simplified to remove any of the Windows-specific pieces since the standard library supports everything we need.

That said, even on 2.7 this PR improves the experience on Windows over the current behavior.

pjbull:
@ssbarnea PR is updated and tests are passing except one complexity check in code climate. I don't think the complexity is too drastic, and it is a straightforward way to support symlinks without changing any core existing logic, so I think it may be worth an exception. I could look at a refactor if that is preferable.

ssbarnea (approved):

ssbarnea:
That is one of the oldest PRs I ever seen and getting it to green shows a lot of dedication. I approved it but with the side note that my review was not really extensive, so please I ask at least another core to review it before making a decision, nobody likes regressions.

pjbull:
> so please I ask at least another core to review it before making a decision, nobody likes regressions.

Agreed, I was leaning on the test cases that I wrote way back when, so if there are ideas for scenarios that need more comprehensive tests, I would be happy to write them!

ssbarnea:
Closing as OP lost interest in driving it. We can reopen as I was supporting the change.

pjbull:
@ssbarnea Sorry, I thought this was waiting for review from another maintainer. Is there something else I can do to support this change getting merged?

ssbarnea:
Sure, just keep it reporting green on CI and ping/poke a maintainer for review. We all struggle with available time...

pjbull:
Just made some updates to this branch to resolve conflicts. Could this PR be re-opened (looks like I don't have permissions for that), or should I start a new one?

insspb:
Hi @pjbull 
Thank you for interest. this PR cannot be reopened because it was force pushed. So you need to open new one. 
Looks like big feature and a lot of discussion here. As I was thinking about some different problems, to speed up review of this please make some more changes

- Move everything new from os.path to pathlib, as it is what considered next huge merge (before this)
- Link to this discussion, so I can reread all of this
- Update docs (we are supporting python 3.7 and up, I see 3.2 there
- Update logger statements to logger.debug("message {}", var_in_message)
- remove from __future__ if not required
- extend functions :param variable: definition in functions where you declare variables. If function does not contain :param for other variables, please add them too

After all done all tests should pass, coverage should be 100% and I will be glad to look on this. 
I am very interested to test it on windows, where I will be in few days.

Thank you.

pjbull:
Thanks for the detailed comments @insspb. Will do!

insspb:
@pjbull You are welcome. I recommend to start from everything except moving to pathlib, and do pathlib move after other codebase will be updated. In this case it will be much more easy to you how to move this in each case of functionality usage. It will take few days to wait.

PR #941, opened 2017-05-09
Use a more lightweight python porting library.

asottile:
There's a few beefs I have with using `python-future`, notably:

- It installs the "new" module names in python2 (has some negative
  compatibility with libraries that poorly detect python2 / python3).
- Introduces `newbytes` and `newstr` whick leak out of modules into code which
  doesn't understand how to handle their types (often leads to errors in
  python 2 where `str(x)` causes `"b'...'"`).
- These are summarized in more detail
  [here][1]

Some other useful links from changes made in this PR:

- I dropped `iteritems(...)`, it's not going to make a performance difference
  here.  If you want an additional read on this, openstack has a nice
  [writeup][2] on this.

[1]: https://github.com/Yelp/yelp_clog/pull/18#issue-143837308
[2]: http://lists.openstack.org/pipermail/openstack-dev/2015-June/066391.html

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -6,9 +6,7 @@ from collections import OrderedDict
 import json
 
 import click
-from past.builtins import basestring
-
-from future.utils import iteritems
+import six
 
 from jinja2.exceptions import UndefinedError
 
@@ -164,7 +162,7 @@ def render_variable(env, raw, cookiecutter_dict):
             render_variable(env, v, cookiecutter_dict)
             for v in raw
         ]
-    elif not isinstance(raw, basestring):
+    elif not isinstance(raw, six.string_types):
         raw = str(raw)
 
     template = env.from_string(raw)
@@ -199,7 +197,7 @@ def prompt_for_config(context, no_input=False):
     # First pass: Handle simple and raw variables, plus choices.
     # These must be done first because the dictionaries keys and
     # values might refer to them.
-    for key, raw in iteritems(context[u'cookiecutter']):
+    for key, raw in context[u'cookiecutter'].items():
         if key.startswith(u'_'):
             cookiecutter_dict[key] = raw
             continue
@@ -224,7 +222,7 @@ def prompt_for_config(context, no_input=False):
             raise UndefinedVariableInTemplate(msg, err, context)
 
     # Second pass; handle the dictionaries.
-    for key, raw in iteritems(context[u'cookiecutter']):
+    for key, raw in context[u'cookiecutter'].items():
 
         try:
             if isinstance(raw, dict):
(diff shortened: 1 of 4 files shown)


asottile:
I'd understand completely if you don't want this PR, thought I'd throw it out there anyway :)

six is much more lightweight than future though :)

```
$ du -hs six-1.10.0.tar.gz
32K	six-1.10.0.tar.gz
$ du -hs future-0.16.0.tar.gz 
808K	future-0.16.0.tar.gz
```

hackebrot:
Hi @asottile! 👋 

Thanks a bunch for your contribution. I think we used six a while back, but can't remember why we moved away from it. No objections from me, if we can make sure it doesn't break compatibility.

Any thoughts @audreyr @pydanny @michaeljoseph?

asottile:
@hackebrot I've rebased this (it had bitrotted) -- would love to get some feedback on this PR!

michaeljoseph:
I'm 👎 on this, given https://docs.python.org/3/howto/pyporting.html recommendations...
Specifically, I prefer the approach of writing Python 3 code that can run on 2.7...

michaeljoseph:
Having just looked at the diff, I'm reasonably impressed that our compatibility exposure is so small 🙌

michaeljoseph:
👀 #301

asottile:
@michaeljoseph are you still thumbsdown given how dead simple this patch is? (fwiw future isn't even being used the way it is designed, cookiecutter is really only using the six-like bits)

michaeljoseph:
@asottile given how minimal this changeset is, i'm willing to be convinced by the majority...
and of course my 👎 is imho...

pydanny:
@asottile, thanks for this contribution. I understand your desire to get your pull request in, but I honestly don't see what this is fixing. Cookiecutter works in these environments without a hitch:

* Python 2.7
* A bunch of of Python 3s
* All major operating systems
* All major packaging systems

If one of those is breaking, then adopting your fix should be adopted. Otherwise, I just don't see what it gets us.

asottile:
@pydanny the problem isn't that cookiecutter is broken, it's that `python-future` breaks other things just by being installed.  I've linked my rationale in the original post -- here's an example where it broke [pytest](https://github.com/pytest-dev/pytest/pull/4074) for example

FlorianLudwig:
With python 2 being retired in 4 month the solution to `python-future` breaking stuff is to drop `python-future`?

asottile:
> With python 2 being retired in 4 month the solution to `python-future` breaking stuff is to drop `python-future`?

As things stop using python2 yeah there's no need for python-future -- that said it can mostly be replaced today with other equivalent porting libraries (such as `six`) which aren't as intrusive / problematic.

ssbarnea:
@asottile Can you please rebase it? I recently become core on cookiecutter and I do fully subscribe to your idea. In fact I did the same in other projects in the past, long time ago.

asottile:
I have rebased this!  thanks for the ping

PR #944, opened 2017-05-15
Allow Jinja extensions to be defined/bundled in the template repository

ramnes:
This commit permits users to define custom Jinja extensions to be used in the `cookiecutter.json` file.

For example, with something like this in `template/helpers.py`: 
```python
from jinja2.ext import Extension


def get_app_slug(app_name):
    return app_name.lower()


class HelpersExtension(Extension):

        super(Extension, self).__init__()
        environment.filters["get_app_slug"] = get_app_slug

__all__ = ["HelpersExtension"]
```

It allows to use that Jinja filter directly:
```json
{
    "app_name": "ExampleProject",
    "app_slug": "{{cookiecutter.app_name|get_app_slug}}",
    "_extensions": ["helpers.HelpersExtension"]
}
``` 

This is really just a dumb example, but I'm pretty confident that it could be a powerful feature.

I'm kinda new to Cookiecutter, so please point me to the right direction if there's already another way to achieve something like this that I'm not aware of.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -10,6 +10,7 @@ library rather than a script.
 from __future__ import unicode_literals
 import logging
 import os
+import sys
 
 from .config import get_user_config
 from .generate import generate_context, generate_files
@@ -62,6 +63,7 @@ def cookiecutter(
         no_input=no_input,
         password=password
     )
+    sys.path.insert(0, repo_dir)
 
     template_name = os.path.basename(os.path.abspath(repo_dir))
 


ramnes:
@hackebrot any feedback?

theodesp:
@ramnes Looks good. Have you got any real cases when you've used Jinja Extensions?

ramnes:
Yeah, at numberly we use our fork and we leverage this feature a lot.

We use filters almost everywhere, because it's so convenient to output different things based on a single input.

The example in my first comment is very basic but shows pretty well how we're using it.

theodesp:
Could you PM the rest of the Team on Gitter about your progress? 👍 
https://gitter.im/audreyr/cookiecutter

ramnes:
@theodesp I've already done that in July, without much success. :disappointed:

michaeljoseph:
I would support a more explicit solution to this issue:
* validate the presence of a well known directory in a template repo (called `extensions` or similar)
* add this directory to the import path when we initialise the Jinja environment
* tests and documentation

ramnes:
This implementation is how I expected cookiecutter to work in the first place, and it's good to have instinctive behaviors working out of the box.

I'd be open to implement it the way you say, but it would take more time, wouldn't be as obvious since you'd have to look at the docs, whereas here it's super simple and just works…

Don't want to keep it simple? What are the downsides?

ilsken:
I [hacked together](https://pypi.python.org/pypi/cookiecutter_repo_extensions/0.2.1) something to make this available without a fork.

`cookiecutter.json`:
```
{
  "repo_name": "test",
  "foo": "bar",
  "_extensions": [ "cookiecutter_repo_extensions:Extension", "ccext:TemplateUtils" ]
}
```

```
# template/extensions/ccext.py
# -*- coding: utf-8 -*-

from jinja2.ext import Extension


def split(value, delim=','):
    return [ x.strip() for x in value.split(delim) ]

class TemplateUtils(Extension):
        super(TemplateUtils, self).__init__(environment)
        environment.filters['split'] = split

```

Only requires you to have `cookiecutter_repo_extensions` alongside `cookiecutter` and then all repo local extensions should load fine (as long as this extension is loaded first).

Not pretty and I'm not great at python so forgive the quality. Just a work-around for people who need want this feature but don't want to maintain a fork.

ramnes:
@pydanny gentle ping

As said on gitter two months ago, done is better than perfect, and this is a two lines diff that's sitting here for more than a year now. Can we do something about it?

ramnes:
Note to myself: just implement that `extension` directory idea.

insspb:
@ramnes 
Thank you for contribution. While this was in review, more correct approach in PR #1240 was introduced. 
This one closed as duplicate.

ramnes:
Great!

PR #945, opened 2017-05-15
Add test_requirements.txt (fix #377)

### LINKED ISSUE #377: Improve the local machine virtualenv test story
If you are manually testing instead of using `tox`, to determine the test requirements is too challenging. While we should be testing in `tox`, this is a barrier to easily working on Windows.

Proposals:
- Document the test requirements
- Provide a `test_requirements.txt` file.

ramnes:
Tox is great. Although, a lot of people are used to the simpler `pip install -r test_requirements.txt` and `pytest` workflow, including me.

Adding this `test_requirements.txt` file doesn't cost much and makes things easier for new developers and for rapid testing.

This PR is related to #377.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -14,3 +14,6 @@ ignore = E731
 [bdist_wheel]
 universal = 1
 
+[tool:pytest]
+testpaths = tests
+addopts = --cov=cookiecutter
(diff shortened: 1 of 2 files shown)


hackebrot:
Hi @ramnes! 👋 

Thank you for your contribution. 🍪

hackebrot:
FYI I removed ``-e .`` from the requirements file, so we can better use it from tox. Otherwise tox would test against the intermediate distribution.

PR #951, opened 2017-05-21
Add cookie-cookie to README.rst

tuxredux:
I created my own version of [cookiecutter-template](https://github.com/eviweb/cookiecutter-template) based on my [cookiecutter-git](https://github.com/tuxredux/cookiecutter-git) project for automatic remote repository creation! I also transferred **cookiecutter-git** from the `webevllc` namespace to `tuxredux`, so I went ahead and fixed the broken link in `README.rst`. Thanks in advance for your time!

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -459,6 +459,10 @@ Meta-templates for generating Cookiecutter project templates.
 
 .. _`cookiecutter-template`: https://github.com/eviweb/cookiecutter-template
 
+* `cookie-cookie`_: A project template for... project templates...
+
+.. _`cookie-cookie`: https://github.com/tuxredux/cookie-cookie
+
 Ansible
 ~~~~~~~
 
@@ -471,9 +475,9 @@ Ansible
 Git
 ~~~
 
-* `cookiecutter-git`_: Git Cookiecutter. A git repository project template!
+* `cookiecutter-git`_: A git repository project template!
 
-.. _`cookiecutter-git`: https://github.com/webevllc/cookiecutter-git
+.. _`cookiecutter-git`: https://github.com/tuxredux/cookiecutter-git
 
 
 C


hackebrot:
Hey @tuxredux! 👋 

Thank you for your PR. To be frank, I'm not sure if having this much logic in hooks, especially remote calls, is a good idea. I realize that this is a general problem with how hooks work in cookiecutter 😶 

What do you think @pydanny @audreyr @michaeljoseph?

tuxredux:
@hackebrot Thanks for your reply!

Sorry, I'm confused :thinking: What do you mean by

> general problem with how hooks work

If you're worried about too much logic I would kindly point out that the `post_gen_hook.py` in `cookiecutter-django` for example is [299 lines long](https://github.com/pydanny/cookiecutter-django/blob/master/hooks/post_gen_project.py#L299), whereas the `post_gen_hook.py` in `cookie-cookie` and `cookiecutter-git` are only [168 lines long](https://github.com/tuxredux/cookie-cookie/blob/master/hooks/post_gen_project.py#L168). A good chunk of the hook script is also comprised of utility code. A `run` function to wrap and simplify `subprocess` calls, and a `requests` class to use the builtin `urllib2` library instead of the third party [requests](http://docs.python-requests.org/en/master/).

If you're worried about excess network usage then a good place to start in my humble opinion would be the initial cookiecutter download. It seems if you're using a remote cookiecutter, the whole cookiecutter repository is deleted and re-cloned even if it's the same commit, and doing so is required otherwise the prompt exits and the project is not generated. If you're worried about security, all password prompts are using the builtin python library `getpass`!

Now I'll be honest I still need to add full coverage unit testing with continuous integration, but the cookiecutters have been working great for me without any major problems so far. I believe they embody the true spirit of the cookiecutter project! A major task in managing a modern software project is creating a remote VCS repository, so why not automate it like the project's structure has been?

All in all, I hope I don't come across as disrespectful in anyway as I have the utmost respect for you and the rest of the cookiecutter team! You guys have done amazing work, and while you may disagree with what I have to say I'm still eager to receive your feedback! :smile:

michaeljoseph:
@hackebrot I agree that the extensive `post_gen_hook.py` may be brittle, but I don't think that should be a reason for not including this template.

michaeljoseph:
Thanks @tuxredux 👋

hackebrot:
You're right @michaeljoseph! 👍 

Thank you @tuxredux 🙇

PR #958, opened 2017-06-05
Add temporary debug tox env

hackebrot:
⚠️ DO NOT MERGE ⚠️ 

Just testing if our tests actually run on both x86 and x64.


hackebrot:
Works on appveyor 😸

Issue #845, opened 2016-11-01
Add support for "private" cookie cutters to provide commercialisation option


freakboy3742:
This is a broad strokes feature proposal; I'm motivated to develop it further and implement it if it is of interest to the cookiecutter team.

The problem: Cookiecutter templates represent an expression of knowledge, experience, and opinion.  The templates take time to develop, as does the knowledge and experience that leads to their development.

In parallel, the open source community has a general problem with fundraising. 

There is, and will always be, a place for fully open, public repositories sharing cookiecutter templates. However, the expertise contained in cookiecutter templates also represents an opportunity for monetisation - selling access to templates.

I'm not proposing to add credit card processing or any of the *financial* side of the process to cookiecutter. What I'm proposing is modifications that would enable an end user to *use* a template in a way that is compatible with monetisation.

This could take many forms, including, but not necessarily limited to:
* Cloning private URLs (is this possible already?)
* Template rollout from password protected zip files, encrypted bundles, etc.
* "non-cacheable" flags for templates, so that the the raw template is never saved permanently.
* "registration keys" that can be entered at time of use that contain pre-set values for certain template values (for example, a template for a website might have ``domain_name`` as a template key; the registration key would set this value in the template, rather than making it user-customisable.
* Pinging a registration server to receive decryption keys or registration keys and track usage.

These changes are mostly intended to modify Cookiecutter in such a way that an end user can *use* a template, without *seeing* the raw template; and, beyond that, constrain the use of the template to those use cases that have been paid for (by install count, target, or whatever)

I don't imagine that these changes will be piracy proof. At the end of the day, you have to transmit the bytes. A motivated individual could get those bytes, and thus the template, and give it away. These measures are aimed at making it easier to do the right thing.

I expect that this proposal will be at least mildly controversial, because it relies on "hiding" information (the template), which some might object to because of an aggressive "everything should be free as in freedom" worldview. My argument here is that what a cookiecutter template provides is "distilled expertise" - it's consulting without actually paying for the dedicated time of a consultant. Providing a mechanism to require people to *pay* for that expertise is, in some regards, the exact model that free software advocates often point to as a reliable model for funding open source. The only difference is that the role of the consultant has been automated.

Is this an idea that the cookiecutter core team finds interesting?

pydanny:
This is a very interesting idea. Initially, I'm for it!

Do keep in mind that git cloning from private URLs is already a feature of Cookiecutter. The rest of it is up in the air.

[...]

PR #961, opened 2017-06-17
Add option to use templates from Zip files or Zip URLs 

freakboy3742:
There are two reasons to add this feature:

Firstly, zip files are a convenient way to distribute a large number of files in a fixed structure. This means a cookiecutter can literally be a single file, rather than a directory, making it easier to distribute.

Secondly, this can be used to address #845 without cookiecutter itself needing to incorporate commercialisation features. The potential commercialisation path is as follows:
* Developer uploads a Zip file to a private S3 bucket.
* Developer deploys a website to sell their template.
* User buys a license from the website
* Developer website provides the user a public url, protected by username/password or access credentials of some kind
* Any attempt to access the public URL is validated against the sales database. This can limit template uses to a particular count, or only allow access for a period of time, or any other scheme that the developer chooses.

Of course, this won't stop a malicious user from copying the template from the cache or anything like that - but from a "make it easy to do the right thing" perspective, this would be sufficient to monetise a template.


hackebrot:
Hey @freakboy3742! 👋 

Thanks for this PR! I like the idea and would be happy to add this functionality.

I haven't had the time to give this a thorough review yet. Hopefully I'll find some time during/after PyData Berlin and EuroPython. :crossed_fingers: 

What do you think @audreyr @pydanny @michaeljoseph?

michaeljoseph:
Yeah, I'm 💯👍  on this- I'll try to find some time to review this weekend...

cookiecutter/zipfile.py
```diff
@@ -0,0 +1,88 @@
...
+def unzip(zip_url, is_url, clone_to_dir='.', no_input=False):
+    # Ensure that clone_to_dir exists
+    clone_to_dir = os.path.expanduser(clone_to_dir)
+    make_sure_path_exists(clone_to_dir)
+
+    if is_url:
+        # Build the name of the cached zipfile,
+        # and prompt to delete if it already exists.
+        identifier = zip_url.rsplit('/', 1)[1]
+        zip_path = os.path.join(clone_to_dir, identifier)
```
hackebrot: We could then also delete the zip archive once it has been extracted

cookiecutter/zipfile.py
```diff
@@ -0,0 +1,88 @@
...
+    else:
+        question = (
+            "You've downloaded {} before. "
+            "Is it okay to delete and re-download it?"
+        ).format(path)
+
+        ok_to_delete = read_user_yes_no(question, 'yes')
+
+    if ok_to_delete:
+        if os.path.isdir(path):
```
freakboy3742: Zip archives are; but the unpacked zip directory isn't. There's a second usage of this function on line 81 to purge an unpacked zip directory.

cookiecutter/zipfile.py
```diff
@@ -0,0 +1,88 @@
...
+        zip_path = os.path.abspath(zip_url)
+        ok_to_delete = None
+
+    # Now unpack the repository. The zipfile will include
+    # the name of the template as the top level directory;
+    # Check if that directory already exists, and if so,
+    # prompt for deletion. If we've previously OK'd deletion,
+    # don't ask again.
+    zip_file = ZipFile(zip_path)
+    unzip_name = zip_file.namelist()[0][:-1]
```
freakboy3742: Good point. I'll beef up the validation here.

cookiecutter/zipfile.py
freakboy3742: I did think about this; I ended up keeping both on the "what's a few more bytes?" principle. Effectively, we have three options:
1. Keep both (what is currently implemented)
2. Keep the zip file, and unpack to a temp directory
3. Keep the unpacked victory, and delete the zip file

Moving to (2) is a simple solution; but (3) gives us the opportunity (at some point in the future) use a password encoded zip file, adding to the potential security for commercial templates. 

So - any preferences?

michaeljoseph: I'm voting for (3), largely because the mechanics of downloading and unzipping are just a means to an end (the template).

Also, I like my victory unpacked 😉

michaeljoseph: I think this would also make it a no-brainer to subsequently re-use `prompt_and_delete` (`cookiecutter.repository` is probably the natural destination of this refactor?)

cookiecutter/zipfile.py
michaeljoseph: Perhaps we should also attempt to catch exceptions related to `ZipFile` errors to wrap in a custom exception?

tests/zipfile/test_prompt_and_delete.py
```diff
@@ -0,0 +1,111 @@
...
+def test_prompt_should_ask_and_rm_dir(mocker, tmpdir):
+    """In `prompt_and_delete()`, if the user agrees to delete/reclone the
+    repo, the repo should be deleted.
+    """
+    mock_read_user = mocker.patch(
+        'cookiecutter.zipfile.read_user_yes_no',
+        return_value=True,
+        autospec=True
+    )
+    dir = tmpdir.mkdir('repo')
```
michaeljoseph: Let's avoid shadowing builtin names?

tests/zipfile/test_unzip.py
```diff
@@ -0,0 +1,229 @@
...
+
+    clone_to_dir = tmpdir.mkdir('clone')
+
+    output_dir = zipfile.unzip(
+        'tests/files/fake-repo-tmpl.zip',
+        is_url=False,
+        clone_to_dir=str(clone_to_dir)
+    )
+
+    assert output_dir == os.path.join(str(clone_to_dir), 'fake-repo-tmpl')
```
michaeljoseph: Here and below, we can use the `py.path.local` api for consistency: `clone_to_dir.join('fake-repo-tmpl')`

theodesp:
@freakboy3742 Can you review the requested changes?

freakboy3742:
@theodesp My apologies - I've been snowed under with other work. I'll take a look ASAP.

freakboy3742:
Apologies for the delay in updating this patch. I've now addressed the issues raised in the two reviews by @michaeljoseph and @hackebrot, and added a couple of extra improvements for good measure:
* Updated docstrings as requested
* Combined the `vcs` and `zipfile` versions of `prompt_and_delete`, and moved the result to the utils package (it can't be in repository as suggested because of circular dependencies)
* Added an extra question to `prompt_or_delete` step that allows the user to reuse an existing template, instead of forcing a re-download. This is necessary for using cookiecutter on planes, on unreliable conference WiFi, etc.
* Improved error handling for badly formed or invalid zip archives
* Restructured some tests to make better use of `pytest.mark.parametrize`
* Modify zipfile behavior to:
  - Keep the original zipfile (if downloaded from a URL) in the `.cookiecutter` directory;
  - When generating a template, unpack the zipfile into a temporary directory; and
  - Delete the unpacked contents after use.
  This ensures that if the user retrieves a Zipfile repository, they won't have an unzipped version visible on their system for any longer than is necessary to generate the templated output. 
* Added ability to extract password-protected Zipfiles. Combined with the previous change, this means you can deliver a commercial template in a password-protected Zipfile. End users will only ever have access to generated output, not the raw template itself. This solution won't stop determined attackers, but it's strong enough to foil casual attempts at circumventing a commercial license.

freakboy3742:
Regarding the coverage results - there are 4 lines missed. Two of those lines are in `cookiecutter/utils.py` and were pre-existing coverage misses. The other two lines are in `cookiecutter/zipfile.py`, but aren't coverable - they're a backwards compatible import shim for the naming of the `zipfile.BadZipFile` exception ([Lines 9-10](https://github.com/freakboy3742/cookiecutter/blob/24420ada0ba7cef7cf9781361171b61f1b3133db/cookiecutter/zipfile.py#L9)) .

cookiecutter/repository.py
```diff
@@ -23,6 +24,11 @@ def is_repo_url(value):
     return bool(REPO_REGEX.match(value))
 
 
+def is_zip_file(value):
+    """Return True if value is a zip file."""
+    return value.endswith('.zip')
```
hackebrot: AFAIK file extensions on Windows are not case sensitive, so maybe we want to change this ``return value.lower().endswith('.zip')``.

cookiecutter/zipfile.py
```diff
@@ -0,0 +1,123 @@
+from __future__ import absolute_import
+
+import os
+import requests
+import tempfile
+from zipfile import ZipFile
+try:
+    from zipfile import BadZipFile
+except ImportError:
```
hackebrot: Can you please add a comment mentioning that ``BadZipfile`` was deprecated in Python 3.2?

docs/usage.rst
```diff
@@ -69,6 +69,42 @@ type of repo that you want to use prepending `hg+` or `git+` to repo url::
 
     $ cookiecutter hg+https://example.com/repo
 
+Works with Zip files
```
hackebrot: I love that you always write great documentation for your changes, @freakboy3742!!! 📝 🙇

freakboy3742:
@hackebrot Thanks for the review - changes have been made!

hackebrot:
Thank you, @freakboy3742! 🙇

hackebrot:
Thank you for your work @freakboy3742! 🙇 🍪

PR #963, opened 2017-06-24
Add gitlab shorthand

hackebrot:
Nothing much to add here 😁 

I think we should support a GitLab abbreviation.

Please let me know your thoughts! Ty 🍪


hackebrot:
Thank you for the quick review @michaeljoseph! 🍪

PR #684, opened 2016-04-28
don't remove project directory on failed hook w/ --overwrite-if-exists

rsyring:
fixes audreyr/cookiecutter#629

Rough draft.  Let me know if you don't like this approach for any reason.

No tests for this change.  I looked at the current tests and didn't see anything test_hooks.py that was checking cleanup functionality.  Is that where the tests for this should go?

PR #964, opened 2017-06-28
Preserve existing output directory on failure

### LINKED ISSUE #629: _run_hook_from_repo_dir should not remove files if they already existed before
The current behavior makes `--overwrite-if-exists` remove all the files (undesirable) if some hook fails.

michaeljoseph:
If a generated output directory already exists, we don't delete it on project generation failure.

Supersedes #684.
Fixes #629.

### DIFF
--- a/tests/test_generate_hooks.py
+++ b/tests/test_generate_hooks.py
@@ -31,14 +31,16 @@ def remove_additional_folders(request):
     Remove some special folders which are created by the tests.
     """
     def fin_remove_additional_folders():
-        if os.path.exists('tests/test-pyhooks/inputpyhooks'):
-            utils.rmtree('tests/test-pyhooks/inputpyhooks')
-        if os.path.exists('inputpyhooks'):
-            utils.rmtree('inputpyhooks')
-        if os.path.exists('tests/test-shellhooks'):
-            utils.rmtree('tests/test-shellhooks')
-        if os.path.exists('tests/test-hooks'):
-            utils.rmtree('tests/test-hooks')
+        directories_to_delete = [
+            'tests/test-pyhooks/inputpyhooks',
+            'inputpyhooks',
+            'inputhooks',
+            'tests/test-shellhooks',
+            'tests/test-hooks',
+        ]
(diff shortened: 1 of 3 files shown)


michaeljoseph:
👀 @rsyring @ionelmc @eliasdorneles @sodre

rsyring:
I would love to see this get merged!  To this day, when iterating on cookie cutters for some types of projects, I have to separate the cookie cutter output from my actual git repo and rsync the changes over so the entire project, including `.git`, doesn't get clobbered.

I appreciate the effort here @michaeljoseph and apologize for not finishing what I started.  Getting into the tests ended up being too high a bar for the time I had.

tests/test_generate_hooks.py
```diff
@@ -115,6 +119,65 @@ def test_oserror_hooks(mocker):
...
+
+    with pytest.raises(FailedHookException) as excinfo:
+        generate.generate_files(
+            context={
+                'cookiecutter': {'hooks': 'hooks'}
+            },
+            repo_dir='tests/test-hooks/',
+            overwrite_if_exists=True
+        )
+        assert 'Hook script failed' in str(excinfo.value)
```
hackebrot: The assertion needs to be outside of the context manager, since the ``FailedHookException`` will cause the block to exit on ``generate_files()``.

tests/test_generate_hooks.py
```diff
@@ -115,6 +119,65 @@ def test_oserror_hooks(mocker):
...
+    os.mkdir('inputhooks')
+    with pytest.raises(FailedHookException) as excinfo:
+        generate.generate_files(
+            context={
+                'cookiecutter': {'hooks': 'hooks'}
+            },
+            repo_dir='tests/test-hooks/',
+            overwrite_if_exists=True
+        )
+        assert 'Hook script failed' in str(excinfo.value)
```
hackebrot: The assertion needs to be outside of the context manager, since the ``FailedHookException`` will cause the block to exit on ``generate_files()``.

rsyring:
@hackebrot THANK YOU!

Issue #510, opened 2015-09-09
Customizable .cookiecutterrc location


nfarrar:
Any chance cookiecutter could allow us to read from a cookiecutterrc file stored in another location (i.e. the XDG default: `~/.config/cookiecutter/config`). I *really* like having my home uncluttered - and lately have been trying to move my configuration files out as much as possible.

This could be done several simple ways:

1. Make cookiecutter check (by default) the current location & for an XDG compatible configuration file.
2. Add a *config* flag (`cookiecutter <...> --config=~/.config/cookiecutter/config`)
3. Read an environment variable, if it exists:

```
import os
config_file = os.getenv('COOKIECUTTER_CONFIG', '~/.cookiecutterrc`)
```

audreyr:
Closing as duplicate of #104, but I think this is something we should do sooner than later, since it's a violation of the XDG spec. Marking #104 as high-priority. Thanks @nfarrar.

Issue #104, opened 2013-11-06
Support alternate config location as per XDG Base Directory Spec


audreyfeldroy:
Suggested by @scanny:

> Might I suggest considering ~/.config/cookiecutter/config as a supported alternate location for the configuration file(s)? This is the location specified by the XDG Base Directory Specification (at least the ~/.config/ prefix is :).
> 
> I first came upon this location when GitHub for Mac put my .gitignore there as ~/.config/git/ignore. Git supports this location from I believe v1.7. I really like keeping as many as possible of my .dotfiles there now because it keeps my home directory more manageable; otherwise I'd have to scroll to see all the .xyz config files there.
> 
> I'm not sure if it would be the right place to place local templates or not. I suppose it depends on whether you consider those configuration or not. I personally wouldn't mind at all having them there, perhaps in ~/.config/cookiecutter/templates.

TODO: look at other examples of this.

nfarrar:
This would be great. :) Cached cookiecutters in $XDG_CACHE_HOME/cookiecutter and user templates in $XDG_DATA_HOME/cookiecutter/templates. :)

bersace:
I workaround this in my dotfiles to unclutter `~/`. https://github.com/bersace/dotfiles/commit/0bbcc2457b578aa9fde8593ca724442c2e7f27e3

That would be nice to support both rc location and move to cache home.

scanny:
Nice workaround @bersace! I think I'll do it that way as well :)

bersace:
@scanny you're welcome ! :)

ssbarnea:
I am closing this because we have an updated version at https://github.com/cookiecutter/cookiecutter/pull/1355 which we will likely merge soon.

HappyEinara:
NB the PR at #1355 seems to use XDG spec for the cache (working) files but does not address this specific issue: the use of `$XDG_CONFIG_HOME/cookiecutter/<some appropriate filename>`  as an alternative path to `~/.cookiecutterrc` in the config flow.

From the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html):
> $XDG_CONFIG_HOME defines the base directory relative to which user specific configuration files should be stored. If $XDG_CONFIG_HOME is either not set or empty, a default equal to $HOME/.config should be used. 

Looks like this could be done with a few elifs:

1. If `--default-config` is forced, no config file
2. If a file is specified on the commandline, use that.
[...]

Issue #517, opened 2015-09-18
Tests should not fiddle with ~/.cookiecutterrc


vincentbernat:
Hey!

Tests are directly using ~/.cookiecutterrc. They fail with mine. I see that there is code to move it out of the way, then move it back. It doesn't seem to work as expected. I am also a bit uncomfortable with tests that move files in my home. It would be better to set `HOME` environment variable to some temp directory instead, wouldn't it?

pydanny:
:+1: to this as a problem.

hackebrot:
I think all of the things we do in the ``clean_system`` fixture should not operate on the users' system.

:+1: for having an isolated test env!

bionikspoon:
ref #623

This is a copy and paste from the pull request.  I wrote up a summary of changes.

## Summary of changes

Use temp directories instead of modifying the file system. 

The tests make use of pytest's `tmpdir` fixture.  Each test works in a unique folder instead of CWD.

### Modifications to the main `cookiecutter` module

- `cookiecutter.config`: replaced `USER_CONFIG_PATH` and `DEFAULT_CONFIG` with function call to 'lazy load' the values

  Reasoning: part of the fix is setting the `HOME` env variable.  However, these static variables are being called in too many places before the earliest test setup function can run.  

  It could've been monkeypatched I guess, but it would've had to been patched several places--everywhere it's being used, and this seems shady.

### Changes to the tests

- New global setup fixture that sets `HOME` env var and changes CWD for the test to the unique test folder.

- Improved cleanup routine, none.
  
  There was a ton of teardown/cleanup fixtures,  I deleted all of that.  With this setup you don't need it.   

  The concept: Each test gets a unique folder for that test.  Each time you run the tests, each of those unique folders are grouped in a folder for the run.  Pytest automatically preserves the trailing 5 test runs.

  tldr; pytest does the cleanup.

- absolute references.
  The problems created by using `tmpdir` and changing CWD:  references to the "tests" folder need to be absolute.   I created a helper function in a new file `tests.utils` called `dir_tests` to make this easier. It's just `os.join` partial.   And I basically did a search and replace  for references to the 'tests' folder.

  _Note: it was originally called `tests_dir`, which is way more conventional, but pytest was testing it._

#### Other changes to the tests

- removed a few gratuitous fixtures. Someone was going fixture happy :smile: 

[...]

PR #971, opened 2017-07-09
add xdg & basic AppData support; fix existing tests

jmsdvl:
Adds support for the XDG spec as per #104 and (sort of) #510. 

Users can specify a location for the config file and through the config file specify locations for cookiecutter related data, which I assume is new since the above mentioned issues were broached; however, it is somewhat roundabout and it would be preferable to explicitly support the "official" locations given by each platform for where an application is supposed to store its stuff.  Therefore, this patch implements the following behavior:

The user configuration file is searched for in the directories given by:

1.  `$COOKIECUTTER_CONFIG`: if set, always takes priority
2.  `~/.cookiecutterrc`: dotfiles in homedir take priority
2.  `$XDG_CONFIG_HOME/cookiecutter/config`: mostly Linux, I imagine, although I've heard of some OS X devs setting this envvar to aid Linux-native apps.
3.  `%APPDATA%\cookiecuttter\config`: Windows
4.  `~/.config/cookiecutter/config`:  this is for lazy-Linux; fairly often the env vars specified by XDG are not actually set, applications running on Linux just behave as though they were.
5.  `~/Library/Application\ Support/cookiecutter/config`: OS X specific; probably won't see much use

Similarly, the locations of `cookiecutters_dir` and `cookiecutter_replay` can now be overridden via the environment (`$COOKIECUTTERS_DIR` and `$COOKIECUTTER_REPLAY` respectively) and are searched for in the following list:

1.  The directory given by the appropriate environment var
2.  `$XDG_DATA_HOME/cookiecutter/<dir>`
3.  `%APPDATA%\<dir>`
4.  `~/.local/share/cookiecutter/<dir>` - see above note about lazy-Linux
5.  `~/Library/Application\ Support/cookiecutter/<dir>`
6.  The existing default locations of `~/.cookiecutters_dir/` and `.cookiecutter_replay`

In both cases the first location that exists is used.

**NOTE:** In order to preserve backwards compatibility and a seamless user experience, if the pre-existing default location for a file or dir is already populated, cookiecutter returns that path immediately instead of searching past item no. 1 on the above lists (i.e., this change _should_ only affect new installs and users who explicitly decide to move their config and/or data).

The existing test suite has been fixed to work with these changes.  Some work remains but I've decided to open this now for feedback, in case anyone has any. What remains is to:

1. Write tests that exercise the configuration / data directory finding functions _per platform_ - probably  by mocking out `sys.platform`
2. Update the documentation appropriately.

I could also whip up a quick CLI command to dump where cookiecutter thinks its config and data dirs are located if you think it would be helpful.

### DIFF
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -5,7 +5,9 @@
 from __future__ import unicode_literals
 import copy
 import logging
-import os
+from os.path import (
+    abspath, expandvars, expanduser, isfile, isdir, join
+)
 import io
 
 import poyo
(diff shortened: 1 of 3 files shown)


jmsdvl:
In connection with this PR, given that I'll need to add at least one more commit to add some tests, I'd be willing to take on #517.  It will probably be easiest to address them concurrently.

verginer:
Support for the SDG spec would be really nice. Any progress on this? 👍

ssbarnea:
Closing due to lack of any update in almost two years. Feel free to reopen if rebased.

Issue #973, opened 2017-07-25
mixed hooks (both sh and py) in the same project


milonimrod:
* Cookiecutter version: 1.5.1
* Python version: 2.7.13
* Operating System: Linux

### Description:

I was wandering if there is an option to add the ability to have both _sh_ and _py_ hook scripts in the package. It seems very straight forward to implement.

Thank you,
Nimrod

milonimrod:
I've created a pull request about it #974

michaeljoseph:
I'm not entirely convinced of the generic utility of this functionality?

Could you give us an example use-case?

milonimrod:
Yes, I have an example (which is the reason I found the need). In our fork we wanted to run after each creation:
1. Git initiation and adding of files
2. Registration of the project in internal remote git server
3. Run a few calls to several APIs
4. Unit a connection to docker

Some of the above are suited for python scripts and some are for bash scripts. So to make the development and maintenance easier we looked for the above feature.

michaeljoseph:
I'm still not convinced that we should be adding this as a core feature for a somewhat minor (IMO) convenience.
🤔 @hackebrot @pydanny @audreyr ?

michaeljoseph:
@milonimrod #824 may be a solution to your problem by allowing some organisation of utility hook code.

milonimrod:
@michaeljoseph I'm not sure how #824 may help as it still require that i'll implement either a python script or a bash script. then I'll need to orchestrate all the calls.

The only caveat that I see in my approach is that it is not clear which script runs first. I would suggest to extend the hooks scripting into an approach like linux services. I mean that the name of the hook will determine the order. For example: 01_pre_gen_project.py, 02_pre_gen_project.sh, 03_pre_gen_project.py

lszyba1:
Couldn't you add your shell code to post_gen
1. Generate initial template
2. Run your SH script
3. Use examples below to modify the already generated templates from part 1 with additional changes after shell returned the results
http://lucasmanual.com/blog/cookiecutter-modify-context-in-post_gen_project-py/
Lucas

milonimrod:
Hi @lszyba1,
I think that solving using 1 & 2 is too and requires the user to have long and cluttered scripts. 3 seems very useful for many tasks but I don't think this is a good approach. As I see it, it can help organize the hooks in a tidy manner

yggdr:
[...]

PR #974, opened 2017-07-25
Changed: Generated projects can use multiple type hooks at same time. (sh + py)

milonimrod:
I've added support for multiple hook types in the same project (both _sh_ and _py_)
#973


theodesp:
Looks good. I would expect more tests though that cover this case.

milonimrod:
@theodesp  - as requested, I added a test

theodesp:
@milonimrod Good Just PM the Team in Gitter for the updates https://gitter.im/audreyr/cookiecutter

jaklan:
@theodesp 2 years later, why it's still not merged? It's already done, it's so trivial, but really, really useful, but still it's not available in master...

PR #980, opened 2017-08-15
Fixed[#824] - Cannot use local imports in hooks

theodesp:
Added **hooks** folder to PYTHONPATH so that the hooks can reference the current folder


cookiecutter/hooks.py
```diff
@@ -82,6 +82,7 @@ def run_script(script_path, cwd='.'):
     try:
         proc = subprocess.Popen(
             script_command,
+            env=get_hooks_env(cwd),
```
theodesp: Include env variable here

cookiecutter/hooks.py
```diff
@@ -142,3 +143,23 @@ def run_hook(hook_name, project_dir, context):
...
+    Create and return a copy of the environmental variables including
+    the current hooks path appended to the PYTHONPATH. This is used for
+    Popen to enable local imports when running hooks.
+
+    :param cwd: The directory to run the script from.
+    :param hooks_dir: The hook directory in the template.
+    :return: A copy of the environmental variables with the addition of
+    the current hooks path appended to the PYTHONPATH.
+    """
+    hooks_env = os.environ.copy()
```
theodesp: Attach a copy of the current env with the hooks folder included in PYTHONPATH

tests/test_generate_hooks.py
```diff
@@ -57,6 +58,19 @@ def test_ignore_hooks_dirs():
...
+def test_imports_in_hooks():
+    try:
+        generate.generate_files(
+            context={
+                'cookiecutter': {'pyhooks': 'pyhooks'}
+            },
+            repo_dir='tests/test-hook-imports/',
+            output_dir='tests/test-hook-imports/')
+    except FailedHookException:
+        pytest.fail("Unexpected FailedHookException...")
```
theodesp: Test it does not raise an ** FailedHookException** when we import  something from hooks

tests/test_generate_hooks.py
```diff
@@ -120,6 +134,32 @@ def test_oserror_hooks(mocker):
 
 
 @pytest.mark.usefixtures('clean_system', 'remove_additional_folders')
+def test_popen_includes_hooks_in_python_path(mocker):
+    err = OSError()
+    prompt = mocker.patch('subprocess.Popen')
```
theodesp: Test that the Popen receives a copy of the current env with the **hooks** folder included in PYTHONPATH

theodesp:
It seems that there are issues on Windows ! Will try to fix them

michaeljoseph:
@theodesp From https://docs.python.org/2/using/windows.html#finding-modules:
> If the environment variable PYTHONPATH exists, as described in Environment variables, its entries are added next. **Note that on Windows, paths in this variable must be separated by semicolons, to distinguish them from the colon used in drive identifiers** (C:\ etc.).

theodesp:
Thanx @michaeljoseph I've fixed it now! woohoo

nikordaris:
what is the status of this?

theodesp:
Hey sorry I will revisit this week

theodesp:
@michaeljoseph Sorry for the insanely long delay. I've updated the PR Thanx.

tests/test_generate_hooks.py
```diff
@@ -57,6 +58,19 @@ def test_ignore_hooks_dirs():
...
 @pytest.mark.usefixtures('clean_system', 'remove_additional_folders')
+def test_imports_in_hooks():
+    try:
+        generate.generate_files(
+            context={
+                'cookiecutter': {'pyhooks': 'pyhooks'}
+            },
+            repo_dir='tests/test-hook-imports/',
+            output_dir='tests/test-hook-imports/')
+    except FailedHookException:
```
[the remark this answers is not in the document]
theodesp: This check is to test that a valid import does not throw an exception. I could not think a way to check otherwise. What's the alternative?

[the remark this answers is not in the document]
michaeljoseph: You could remove the `try: except`- the test will fail if an exception gets thrown?
And then assert that the expected files are present in `output_dir`?

ahobsonsayers:
Is this PR ok to be merged? Are the any blockers i can help with. This feature would be super useful!

insspb:
This branch is ready to be merged, but i have many concerns about security. So I will hold it for hour, or for day, or until 2.0 for final decision. Need to think about it once again.

michal-halenka-datasentics:
@insspb May I ask what is the current status on this?

jackkeck:
Not sure if this has been resolved yet. I am open to jumping in to lend a hand on getting this over the finish line if that helps. Please let me know! Thanks!

devmcp:
Is there any chance this will be implemented? It would be very useful!

jnvilo:
came here looking for a solution to this too but looks like no one with write access has the time to do the review.   I am desperate for this so in the meantime will have to work with a fork  that adds the hooks dir to the python path until it finally gets merge. 

Perhaps a better solution would just be create a hookslib python module and install that so its available for import into our hooks and just close this issue.

kurtmckee:
It may be possible to implement this with fewer surprises by copying the _entire_ hooks directory to a temporary directory (with Jinja rendering the pre- and post- hooks files) and then adding that temporary directory to the PYTHONPATH when Python is run on those files.

My reasoning is that people may write their hooks with syntax like

```python
from . import utils
```

which can be accommodated better by copying all of the files in the hooks directory.

Another possibility is to modify `cookiecutter._template` so that it's an absolute path, and then add boilerplate to the hook files so that they modify their own paths at runtime, like:

```python
import sys

sys.path.insert(0, "{{ cookiecutter._template }}")

import utils
```

For now, I'm closing this PR due to its age.

PR #982, opened 2017-08-21
Copy symlinks as symlinks when supported by the platform

mcfletch:
This allows your templates to use relative or absolute symlinks and have the symlink reproduced in the generated project. This change copies the link-target directly, so if you have a relative link it will wind up relative to the final location, while absolute references will continue to be absolute.

It is *possible* that people are doing strange things in their templates where the rely on symlinks being copied as files, but at least in our case it was surprising to discover that our symlinks were not being reproduced in the final project. It was also surprising to find the cookiecutter crashing when the symlinks were to locations not on the development machine.


insspb:
Should be switchable.

seanh:
Just a quick note to say that this will break our cookiecutters if it isn't switchable: https://github.com/cookiecutter/cookiecutter/issues/1420#issuecomment-1177891798

kurtmckee:

It may be possible for template authors to create symlinks, rather than copying them, using a post-generation hook. For example:

```python
import os

os.symlink('{{ cookicutter._output_dir }}/relative', 'relative-symlink')
os.symlink('{{ cookicutter._output_dir }}/absolute', '/etc/nginx/nginx.conf')
```

Note that on Windows platforms, only Python 3.8 and higher can reliably create symlinks without admin privileges.

PR #990, opened 2017-10-02
Adds wemake-django-template

sobolevn:
Adds new link into `Python-Django` section.

https://github.com/wemake-services/wemake-django-template


sobolevn:
@hackebrot any feedback please?

hackebrot:
Looking good 👍

PR #946, opened 2017-05-15
Give access to Jinja2 environment variables

ramnes:
See the rationale behind this PR: https://github.com/audreyr/cookiecutter/issues/637#issuecomment-262207477

This PR fixes #704.

PR #993, opened 2017-10-11
 Allow passing of Jinja Environment initializer kwargs from the python API

pnegahdar:


insspb:
Hi @pnegahdar 
Thank you for contribution and excuse us for long answer.
This is same feature, that was implemented in #946

But your implementation require some user interaction. But this feature is more required for template developers, so should be available from template. That's why #946 will be merged, and this PR will not. 

If you have other concerns, feel free to ping me via issue or another PR.

Anyway no docs here, no tests here.

pnegahdar:
Thanks! Kind of you to follow up, you didn't have to :)

PR #996, opened 2017-10-14
pydocstyle corections on extentions.py for #742

meahow:
This is for #742
I changed the format of the docstrings for cookiecutter/extentions.py to followe pep257 style


hackebrot:
Hi @meahow! 👋 

Thank you for your contribution. 🍪

PR #997, opened 2017-10-14
pydocstyle for exceptions.py

meahow:
This is for #742
I changed the format of the docstrings for cookiecutter/exceptions.py.
Few of the docstrings required minimal rewording to fit pep257 style.
I hope this is fine.

### DIFF
--- a/cookiecutter/exceptions.py
+++ b/cookiecutter/exceptions.py
@@ -21,7 +21,8 @@ class NonTemplatedInputDirException(CookiecutterException):
 class UnknownTemplateDirException(CookiecutterException):
     """Raised when Cookiecutter cannot determine template directory.
 
-    For example when more than one dir appears to be a template dir.
+    Raised when Cookiecutter cannot determine which directory is the project
+    template, e.g. more than one dir appears to be a template dir.
     """
 
 
@@ -116,7 +117,7 @@ class RepositoryCloneFailed(CookiecutterException):
 
 
 class InvalidZipRepository(CookiecutterException):
-    """Raised or invalid zip repositories.
+    """Raised for invalid zip repositories.
 
     Raised when the specified cookiecutter repository isn't a valid
     Zip archive.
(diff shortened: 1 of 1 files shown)


cookiecutter/exceptions.py
```diff
@@ -1,103 +1,101 @@
...
     The name of the input directory should always contain a string that is
     rendered to something else, so that input_dir != output_dir.
     """
 
 
 class UnknownTemplateDirException(CookiecutterException):
-    """
-    Raised when Cookiecutter cannot determine which directory is the project
-    template, e.g. more than one dir appears to be a template dir.
+    """Raised when Cookiecutter cannot determine template directory.
```
hackebrot: Please keep the original wording.

cookiecutter/exceptions.py
```diff
@@ -110,17 +108,16 @@ class UnknownExtension(CookiecutterException):
...
+    """Raised when the specified cookiecutter repository doesn't exist."""
 
 
 class RepositoryCloneFailed(CookiecutterException):
     """Raised when a cookiecutter template can't be cloned."""
 
 
 class InvalidZipRepository(CookiecutterException):
-    """
+    """Raised or invalid zip repositories.
```
hackebrot: Please fix the typo here: ``Raised for``

cookiecutter/exceptions.py
meahow: I changed it so that the full description is kept as it was initially, but for the summary it is not as easy as it has to fit in one line (and one line budget is 80 chars according to pep8).
So I kept my version here.
If this doesn't match your intent, please let me know how would you like this phrased. 

As a side note, it's worth mentioning that this exception does no seem to be used anywhere in the code (I grepped for it both in the current master and in the history).

meahow:
@hackebrot,
Thanks for quick feedback.
I addressed your comments, please let me know if this is fine.

insspb:
@meahow thank you for your contribution. 
Unfortunately another PR for the same file was merged, and now files pass all pep checks. 
So this would not be merged and closed as a duplicate. 
I am sorry for not keeping FIFO order of PRs.

PR #998, opened 2017-10-14
pydocstyle for __init__.py

meahow:
This is for #742
I changed the format of the docstrings for cookiecutter/__init__.py to follow pep257 style


hackebrot:
Thank you for your contribution @meahow! 🍪

PR #1000, opened 2017-10-14
pyocstyle for main.py

meahow:
This is for #742
I changed the format of the docstrings for cookiecutter/main.py to follow pep257 style


hackebrot:
Thanks for the PR @meahow! 🍪

PR #1006, opened 2017-10-30
Allow non-json-serializable context

### LINKED ISSUE #1005: Allow for non-json-serializable context
I'm moving from a bespoke jinja2 templating system to cookiecutter (used as a python [library](https://cookiecutter.readthedocs.io/en/latest/cookiecutter.html#module-cookiecutter.main)), and am trying to pass arbitrary python objects as extra context, which is something allowed in jinja2. For example, in my current template, I pass a formatting function as context and want to call it like:

```
{% for something in cookiecutter.things.items() %}
{{ cookiecutter.my_formatter(something) }}
{% endfor %}
```

The nice thing about this is that I can arbitrarily adjust the way that the code is formatted, just by injecting different functions.

The problem I'm hitting right now, is that the objects are getting stringified by this line https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/prompt.py#L168. I'm not that familiar with the codebase, but this seems strange that the prompt code path is being used when I am setting `no_input=True`. Though I do understand that `prompt_for_config` is doing the cookiecutter config templating (i.e. [this](https://cookiecutter.readthedocs.io/en/latest/advanced/templates_in_context.html)). Anyway, removing the string casting seems to mostly do the trick...

Except that replayability then breaks - [this](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/main.py#L87) line chokes with `Object of type 'Foo' is not JSON serializable`. We could do something like pickling the context (yuck). But personally, i'd be fine with loudly disabling replaying if a non-json-serializable context is passed. 

It also makes things a little more challenging because now you might have to handle the non-string context case in your template. But again, I think that this is another case of caveat emptor.

Does this sound like a reasonable idea? Is there something horrible that i'm missing?

bplotnick:
This fixes #1005 by no longer stringifying context. This, however also may break json serialization so it adds an option, `no_dump`, to disable json serialization of context (which is used for replays). You should set this if you are passing non-json-serializable context.

This would most likely be a breaking change for many if people are assuming that non-strings in context get stringified (some of my existing cookiecutter code did).

Let me know if there is a better way to handle this.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 def cookiecutter(
         template, checkout=None, no_input=False, extra_context=None,
         replay=False, overwrite_if_exists=False, output_dir='.',
-        config_file=None, default_config=False, password=None):
+        config_file=None, default_config=False, password=None, no_dump=False):
     """
     API equivalent to using Cookiecutter at the command line.
 
@@ -84,7 +84,8 @@ def cookiecutter(
         # include template dir or url in the context dict
         context['cookiecutter']['_template'] = template
 
-        dump(config_dict['replay_dir'], template_name, context)
+        if not no_dump:
+            dump(config_dict['replay_dir'], template_name, context)
 
     # Create project from local context and project template.
     result = generate_files(
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -25,7 +25,8 @@ logger = logging.getLogger(__name__)
 def cookiecutter(
         template, checkout=None, no_input=False, extra_context=None,
         replay=False, overwrite_if_exists=False, output_dir='.',
-        config_file=None, default_config=False, password=None, dump_replay_data=True):
+        config_file=None, default_config=False, password=None,
+        dump_replay_data=True):
     """
     API equivalent to using Cookiecutter at the command line.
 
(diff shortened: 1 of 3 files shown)


pydanny:
@bplotnick, thanks for the PR.

However, please reverse how your `no_dump` flag works so your pull request doesn't cause breaking changes by default.

bplotnick:
`no_dump` is `False` by default, which means dump, which is the current default... I can see the confusion now 😄 . I can change it to be a positive like `dump_replay_data` and default to True. Would that be clearer?

bplotnick:
@pydanny I renamed it to `dump_replay_data` (and flipped the default) so it's clearer that it isn't a breaking change by default.

bplotnick:
I had forgotten about this PR. I fixed a flake8 issue, and I think it's ready to go now. What do you think @pydanny

bplotnick:
@pydanny ping?

bplotnick:
@pydanny any chance you can have a look at this?

kurtmckee:

It may be possible to support this by serializing your object into individual keys. For example:

```python
class MyObject:
    x = 1
    y = {"a": 10}
    z = [1, 2, 3]

o = MyObject()

serialized = {
    "x": o.x,
    "y_a": o.y["a"],
    "z_0": o.z[0],
    "z_1": o.z[1],
    "z_2": o.z[2],
}
```

Issue #851, opened 2016-11-17
Access the Context dict in the Pre/Post-Generate Hook scripts


tuxredux:
Hey, just want to start off by saying thanks for making this awesome project! :)

The title says it all--is there any way to access the CLI Context dictionary in the `pre_gen_project.py` and `post_gen_project.py` hook scripts? I want to be able to add an extra context variable [(a timestamp just like the example)](https://cookiecutter.readthedocs.io/en/latest/advanced/injecting_content.html#example-injecting-a-timestamp), but without having to write my own script that wraps the `cookiecutter.main.cookiecutter` function. I also wanted to be able to reassign the context variables if validation fails, [instead of just raising an exception](https://cookiecutter.readthedocs.io/en/latest/advanced/hooks.html#example-validating-template-variables).

Any help is greatly appreciated! Keep up the great work!

Cadair:
:+1: I came here to ask exactly the same thing!

tuxredux:
Another quick question, is there anyway to access the created project files in the `post_gen_project.py` script? That could solve the extra context variables injection problem for now, as well as [this other issue](https://github.com/audreyr/cookiecutter/issues/853). Thanks in advance for your time!

hackebrot:
Hi @tuxredux! 👋 

Thank you for your feedback. I'm happy to hear that you find cookiecutter useful.

As to your question, hooks cannot modify the context at this point. They are being rendered though, so you could do something like the following in a hook (untested 😁 ):

```python
context = """
{{ cookiecutter | jsonify }}
"""
```

detroitcoder:
Asking a similar question. I wanted to use cookie cutter to deploy applications that require a lot of configuration and was hoping to use cookiecutter to templatize the deployment. I want to add variables that are like {{cookiecutter.current_directory}} or {{cookiecutter.current_user}} without needing to pass these in from the command line. I was hoping there was a hook somewhere that I could set these but I guess not.

I could pass these as extra context using the python api, but that requires a user calling a python script that calls out to the cookiecutter function which requires them to do more pre-installation. Would be great if there was a hook where I could modify the cookiecutter jinja variables.

Pawamoy:
[...]

PR #1022, opened 2017-12-16
ENH: Python module as hooks

0x0L:
Thanks for this project!

This PR deals with #851. Pre and post python hooks become modules that implement the `hook` functions.

Add a file named `python_gen_project.py` in the `hooks/` directory containing

```python
def pre_gen_project(project_dir, context):
    context['some_key'] = 'abc'

def post_gen_project(project_dir, context):
    pass
```

`some_key` will be available while templating files


0x0L:
Alternatively I also implemented a version that `json.loads()` stdout of scripts and merge it into `context['cookiecutter']`.

glenfant:
Is there a chance we have such pre/post-render hooks that can change the context one of these days. I'll work around this with custom markup placeholders that are replaced in post_gen_project.py by my own computed text (including user name, actual date, computed paths, ...). It's dirty, noisy and slow but it works.

0x0L:
ping @audreyr @hackebrot

The CI looks broken `ERROR: arrow requires Python '>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*' but the running Python is 3.4.6`

0x0L:
Could be done through extensions.

PR #1023, opened 2017-12-18
add more test to improve branch coverage

amey589:


insspb:
@amey589 thank you for contribution. 
Just one note. Tests should be separated from each other and minimised. 

As it is old pull request, I refactor it and will merge in #1348

PR #1025, opened 2017-12-22
Updating tutorial1

### LINKED ISSUE #942: Explain how cookiecutter finds template directory
I'm trying to learn cookiecutter, and in the tutorials and such, it doesn't explain how it chooses a template directory. It's just kind of magic that it somehow knows to look for ```{{cookiecutter.project_slug}}``` or any of a variety of other values that are used in other cookiecutter examples.

Source-diving led me to `find.py` and `find_template()` which suggest that it's the first jinja-like thing with "cookiecutter" in the directory name. Which is cool; however, it would be super helpful to make that explicit in the tutorial.

Thanks for the sweet project!

accraze:
Adding more context on how cookiecutter finds the template file

closes #942

### DIFF
--- a/docs/tutorial1.rst
+++ b/docs/tutorial1.rst
@@ -108,6 +108,8 @@ Find the directory called `{{ cookiecutter.project_slug }}`. Click on it. Observ
 the files inside of it. You should see that this directory and its contents
 corresponds to the project that you just generated.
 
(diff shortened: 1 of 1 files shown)


hackebrot:
Hi @accraze ! 👋 

Thank you for this PR.

@michaeljoseph @willingc thoughts? I'm 👍

PR #1034, opened 2018-01-21
Added: _output_dir to cookiecutter context

Casyfill:
trying to add `output_dir` to `context` object - just like a `_template`


Casyfill:
hm, not sure what went wrong... :-/

insspb:
Hi @Casyfill, I reworked and merged this pull request to current master.
It probably will be available in cookiecutter 2.0, if overall context format will not be changed. 
At this time you can use master at current commit to use it.

insspb:
@Casyfill PS thank you for contribution.

PR #1037, opened 2018-01-30
ignore *.pyc files in hooks/

Avsecz:
Cookiecutter should ignore the pyc files in hooks as it leads to decoding errors:

```
UnicodeDecodeError: 'utf8' codec can't decode byte 0xf3 in position 1: invalid continuation byte
```

I'm getting this error on python 2.7

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -36,8 +36,9 @@ def valid_hook(hook_file, hook_name):
     matching_hook = basename == hook_name
     supported_hook = basename in _HOOKS
     backup_file = filename.endswith('~')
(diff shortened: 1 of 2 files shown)


hackebrot:
Hi @Avsecz! 👋 

Thank your for this PR. Would you mind adding a test for that?

Avsecz:
Some tests failed due to a ci-code: `curl: (7) Failed to connect to s3.amazonaws.com port 443: Connection timed out`

ssbarnea:
Closing as this is not fixing an existing bug. If this aims to fix a bug we better: file a bug, create test that replicates the bug markes with xfail, propose PR to fix the bug and make the test active.

PR #1045, opened 2018-02-27
Feature/update cookiecutter to use different delimiters

mkevanz:


pydanny:
Hi @mkevanz, thank you for this contribution.

However, the vision as put down by @audreyr since the beginning is we will always stayed with stock Jinja2 delimiters as it makes it easier for people to use Cookiecutter. As both users and template creators. Therefore, we can't accept this pull request.

pydanny:
Do'h, I see you just closed this PR.

mkevanz:
@pydanny Correct, I accidentally submitted it to the wrong repo. Thanks.

PR #1047, opened 2018-03-02
Adding tox to the list of modules in the test_requirements manifest.

jgrasser:


insspb:
No explanation provided. Have no sense in the current state.

PR #1063, opened 2018-04-09
Support template in a subfolder of a repo

DmitrySavinBloomberg:
[**Bloomberg LP**](https://www.bloomberg.com/) contribution.

Allows to specify a subfolder in the repo in which to search for the template.

## Usage
At Bloomberg we find it convenient to keep the templates in one repo (like templates belonging to the same team, or grouped by any other criteria), i.e.:
```sh
.
|-- cookiecutter-pipproject
|   |-- README.md
|   |-- cookiecutter.json
|   |-- tasks.py
|   `-- {{cookiecutter.app_name}}
`-- cookiecutter-python-app
    |-- LICENSE
    |-- README.rst
    |-- cookiecutter.json
    |-- requirements.txt
    |-- tests
    `-- {{cookiecutter.project_slug}}
```
Unfortunatelyn, `cookiecutter` doesn't allow you to have such a configuration easily.

This patch makes it possible with `cookiecutter' using the following syntax:

```sh
cookiecutter --subfolder cookiecutter-python-app <YOU REPO>
```
where `<YOU REPO>` is the git repo where the templates are and `cookiecutter-python-app` is the name of the template inside the repo (as in the example above) to be used for generating the code.

I hope all the guidelines have been followed, but please feel free to point me out to any problem with the PR!

Many thanks


pydanny:
First off, thanks for this pull request. However, is there any reason why this can't be placed in a tool that employs the internal Cookiecutter API?

vivainio:
FWIW I was just looking for this feature (just to avoid repo profileration). Having it as a flag in the main cookiecutter binary would aid discovery

heitorlessa:
That is awesome! I essentially did this the other day using the Cookiecutter API, but this PR would simplify all the boilerplate and the concerns I had with repo proliferation too for those using the CLI.

+1 to that if the owners agree to that as this adds a huge value

vivainio:
In case this is rejected because it adds a feature to a set that we want to keep minimal, is there an alternative cookiecutter cli application that can accept features like this? Or are users expected to whip up custom applications of their own?

luzfcb:
Speaking as a long-time cookiecutter user ( i'm not a maintainer of this project. ).

I believe that any new feature, however useful and cool it may be, is unlikely to be accepted in this momment, because this project, even if it is used by a lot of people, including extra large companies, is not sufficiently funded, so that their maintainers can not devote more time to it than they currently do.

Maintainers can not cope with an even greater burden of to support this library without sufficient funding. https://github.com/audreyr/cookiecutter/pull/848#issuecomment-259575460.

Current project has almost seven thousand stars in github, but still failed to get even $60 per month of funding.

https://www.patreon.com/hackebrot
https://www.patreon.com/danielroygreenfeld

(Although it is obvious, as incredible as it sounds, software developers, also have the day with 24h like everyone, of which, have to use to get the money to pay the bills. :-) )
 
If you work in a company (or not), and cookiecutter helps you Create Value ( saving you time with boring and repetitive tasks, necessary to create a new project ), and you have some money and want to support this incredible software, talk about it with the maintainers.


related links:

https://www.fordfoundation.org/library/reports-and-studies/roads-and-bridges-the-unseen-labor-behind-our-digital-infrastructure/

https://github.com/pybee/paying-the-piper

pydanny:
@vivainio Cookiecutter does one thing, and it does it quite well. Cookiecutter supports every supported version of Python in every environment, so increasing it's features is often more complicated than it seems. And as much as we love working on this project, our work on it comes out of our free time. Adding more features means the maintainers have to do even more unpaid volunteer work.

@luzfcb is right to point out this project is not sufficiently funded. It's frustrating knowing how many Fortune 500 companies use this tool and yet all that we make from it is worth what in the USA we call "Enough for Taco Bell twice a month". 

That said, I hear your plea and those of the others on this thread. We love this project and want to improve it. We're working to change the funding method. To that end, we are launching paid courses that use Cookiecutter extensively. A substantial amount from these courses will be going to support Cookiecutter and its ecosystem.

So if you want this feature and this project to move forward, please go to https://courses.twoscoopspress.com/courses/creating-and-distributing-python-packages

pydanny:
@DmitrySavinBloomberg with the 5 minutes of unpaid volunteer time I can afford to spare, it doesn't appear to have any problems.

vivainio:
Assuming cookiecutter won't be growing in features much due to resource starvation, it seems like the best workaround is to write own cli's using cookiecutter api (but cli would deal with discovery etc, and just do the rendering with cookiecutter lib). Are there best practice or documentation examples on how to use the api, or is it just about reading the source? An example of such an utility I was thinking of would be seeking the templates from package.json up the tree, and then render them in current directory.

PR #1066, opened 2018-04-16
Add link to python-module-cookiecutter

gsemet:
Hi,

This is the cookiecutter I use everytime I need to create a python module. I started using pipenv a months ago, and combined with PBR and a proper tooling (doing here around a Makefile), it applies very well to library and application use cases. It is generic, that is does not bootstrap the code, only the tooling and build system.

PBR and pipenv handles pretty much all key things:
- dependencies declaration
- requirements.txt generation (for library) to support services such as readthedocs that does not support properly Pipfile yet
- library tracks only Pipfile, PBR handle the injection into setup.py properly
- application tracks both Pipfile and Pipfile.lock to maximize reproductibility
- PBR allows semantic commit ("sem-ver: feature" in the commit message means a jump of minor versioning, "sem-ver: api-break" means a jump of major version)
- of course "make tag-pbr" can automatically create the right tag according to the latest semantic commit.
- travis.yml secret automatically retrieved from travis
- travis pushes to pypi on successful tag build
- yapf/isort formatting (enforcing 1 import per line to simplify merges)
- test with pytest of course
- fancy classic stuff: coverage report, editorconfig, dockerfile, ...

To see its usages:
- library: https://github.com/gsemet/cfgtree
- application: https://github.com/Guake/guake

planned: backport the automatic release note generation using OpenStack's reno project.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -336,6 +336,7 @@ Python
 * `cookiecutter-bottle`_ : A cookiecutter template for creating reusable Bottle projects quickly.
 * `cookiecutter-openstack`_: A template for an OpenStack project.
 * `cookiecutter-docopt`_: A template for a Python command-line script that uses `docopt`_ for arguments parsing.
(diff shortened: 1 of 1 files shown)


ssbarnea:
We are no longer maintaining a list of templates inside README file and instead we use github topics feature. Please read https://github.com/cookiecutter/cookiecutter#a-pantry-full-of-cookiecutters and add the needed topics to your repository and people will be able to find your template. The most important one is ``cookiecutter-template``.

We are also going to remove most of already listed templates from there, in order not avoid being biased.

PR #1072, opened 2018-04-24
Add cookiecutter-pipenv: Cookiecutter Python Package Template with Pipenv

elgertam:
I forked @audreyr's cookiecutter-pypackage so that it now uses Pipenv and Pipfile instead of pip and requirements.txt. In addition, I updated the docs to recommend using pipsi instead of manually installing Cookiecutter or Pipenv into virtualenvs. Other than a few further changes to the docs, this is substantially similar to the cookiecutter-pypackage.

I'd love it if you would add the link to my forked template. 😀

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -327,6 +327,7 @@ Python
   template.
 * `cookiecutter-pipproject`_: Minimal package for pip-installable projects
 * `cookiecutter-pypackage-minimal`_: A minimal Python package template.
+* `cookiecutter-pipenv`_: Python Package Template with Pipenv.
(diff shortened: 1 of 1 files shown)


ssbarnea:
We are no longer maintaining a list of templates inside README file and instead we use github topics feature. Please read https://github.com/cookiecutter/cookiecutter#a-pantry-full-of-cookiecutters and add the needed topics to your repository and people will be able to find your template. The most important one is ``cookiecutter-template``.

We are also going to remove most of already listed templates from there, in order not avoid being biased.

PR #1076, opened 2018-05-28
Add skip-if-file-exists when doing overwrite-if-exists

chhsiao1981:
This PR is useful for the following scenario:

I've already setup a project, but I would like to have some cookiecutter for additional modules.

Basically I can't achieve the above scenario by just having "overwrite_if_exists",
because the existing files will be reset if it happened to be in the cookiecutter template-files.

This PR intends to skip the existing files when doing overwrite_if_exists.


insspb:
@chhsiao1981 thank you for contribution. Contribution mention restored in #1345

chhsiao1981:
@insspb 

Thank you so much for merging this pull-request!

Issue #1088, opened 2018-07-30
Customize jinja parser


fatal10110:
Hey guys, I'm trying to use the cookiecutter to generate a code base for my k8s + Helm project.
Helm uses "go templates" that has the same syntax like jinja templates, it causes conflicts, so I came up with a new feature... Jinja provides an ability to change it's syntax by modifying the environment.

We can provide these parameters as object in `cookiecutter.json`

```
    {
        "_environment": {
            "variable_start_string": "[[",
            "variable_end_string": "]]"
        }
    }
```

WDYT?

Here is the PR #1089

fatal10110:
@matusf Nope, I can not... Because I need also cookiecutter tags there
example:

```
spec:
   container:
       image: {{ .Values.repository }}/{{ cookiecutter.image_name }}
```

matusf:
@fatal10110,  this can be done with `raw statement` as well.
```
spec:
   container:
       image: {% raw %}{{ .Values.repository }}{% endraw %}/{{ cookiecutter.image_name }}
```
But I understand that it may not be the most elegant solution for every situation

fatal10110:
Yeah, in big files that contains a lot of tags it may be a problem (almost hell :D)
https://github.com/helm/charts/blob/master/stable/acs-engine-autoscaler/templates/deployment.yaml

senderle:
New cookiecutter user here, trying to understand the problem to avoid future issues like this. Aren't cookiecutter variables explicitly namespaced? I expected that when you render the template, cookiecutter would only touch the namespaced tags. As long as you've defined all the variables in your template, how do conflicts arise? Is this some kind of testing issue?

fatal10110:
@senderle Nope, it uses template engine called "Jinja 2", so it doesn't really knows about cookiecutter...
It gets two inputs 1. Template string 2. Template params, so it will try to match any variables between `{{` and `}}`

senderle:
@fatal10110 yeah, I use Django. Jinja is a clone of Django's template system. But this didn't seem to pose problems for [cookiecutter-django](https://github.com/pydanny/cookiecutter-django).

senderle:
Though looking more closely at the code I see they are using 

    {% raw %} 

tags in some places. OK!

ldaverio:
Hello everybody,

I would also like to be able to access the Jinja2 environment easily, typically to enable a line statement prefix such as `//#`, allowing me to write JS templates such as:

```javascript
[...]

PR #1089, opened 2018-07-30
Feature: Custom environment variables

fatal10110:
Implements #1088


pouledodue:
is possible to merge this, I need to set `keep_trailing_newline=True` in jinja

cookiecutter/environment.py
```diff
@@ -7,6 +7,31 @@
 from .exceptions import UnknownExtension
 
 
+class CustomEnvironmentArgsMixin(object):
```
quantumlicht: While this is convenient, it seems like the cookiecutter prefers users using jinja2 extensions to extend functionality.

You can write a simple extension like this for your use case

```
class CustomDelimitersExtension(Extension):

        super(CustomDelimitersExtension, self).__init__(environment)
        environment.variable_start_string = "[["
        environment.variable_end_string = "]]"
```

quantumlicht: I would probably remove the addition of the mixin from this PR to make it easier to accept.
I think the key point you covered, is passing the env variable to `find_template`. Extensions are loaded when env is instantiated, so your variable delimiters would be taken into account.

quantumlicht:
Can somebody from the maintainer team take a look at this? With my recommendation (to remove the new mixin), I think this serves as a bug fix to allow using custom delimiters properly.

JustinPlute:
Any update on this?

con-f-use:
See #1339 for an alternative implementation.

stefan-caraiman:
bump

kurtmckee:

PR #1112, opened 2018-10-01
pep257 fixed for exceptions.py

gahan9:


insspb:
Both files currently pass pep257 check. Declined.

PR #1111, opened 2018-09-30
Unbreak tests

HaraldNordgren:
While working on
- https://github.com/audreyr/cookiecutter/pull/1106
- https://github.com/audreyr/cookiecutter/pull/1107

I noticed that the tests are failing on unrelated issues. Created the PR here with vanilla `master` first and verified that the tests fail already there.

Seems that `click` v7.0 introduces some new error reporting. So I think either we change the tests like this or figure out a nice way to freeze the click dependency at `v6`.

PR #1119, opened 2018-10-28
Fix several warnings

BoboTiG:
Note that failing tests are fixed in #1111, I did not want to backport them but can do it if requested :)

### DIFF
--- a/cookiecutter/repository.py
+++ b/cookiecutter/repository.py
@@ -15,7 +15,7 @@ REPO_REGEX = re.compile(r"""
  |                                      # or
  (\w+@[\w\.]+)                          # something like user@...
 )
-""", flags=re.X)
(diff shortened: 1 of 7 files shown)


cookiecutter/repository.py
```diff
@@ -11,12 +11,11 @@
 from .zipfile import unzip
 
 REPO_REGEX = re.compile(r"""
-(?x)
 ((((git|hg)\+)?(git|ssh|https?):(//)?)  # something like git:// ssh:// etc.
  |                                      # or
  (\w+@[\w\.]+)                          # something like user@...
 )
-""")
+""", flags=re.X)
```
nicktimko:
```suggestion
""", flags=re.VERBOSE)
```
would be nice if you forget what `X` is.

PR #1120, opened 2018-11-01
Added cookiecutter-amanita minimal python template to the list

constrict0r:
Added a minimal python template with amanita flavor.


insspb:
Target removed. Link broken.

PR #1133, opened 2018-12-28
Implement cookie-pick feature

nalch:
TODO: Put description here

Fixes: #1134

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -236,4 +236,4 @@ Contributors
 .. _`@asottile`: https://github.com/asottile
 .. _`@jonathansick`: https://github.com/jonathansick
 .. _`@hugovk`: https://github.com/hugovk
-
+.. _`@nalch`: https://github.com/nalch
--- a/README.rst
+++ b/README.rst
@@ -163,6 +163,8 @@ Did someone say features?
 
 * Projects are always generated to your current directory.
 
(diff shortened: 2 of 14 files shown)


ssbarnea:
Closing this for several reasons: outdated, never got attention from anyone and maybe the most important aspect: it does not solve an existing bug or feature request. By having a bried look at this change I say "that is a lot of code added for something that looks like a corner case".

If you really want this to happen, you should better start with a feature request that describes very well the use case. If it manages to get the buy-in from few others, it worth creating a PR to fix it.

nalch:
Maybe the look was too brief. The PR solves the linked feature request #1134. On top of that the added documentation gives details about the use case.
This exact use case of migrating cookiecutter changes came up again and again over the years (f.e. #784 and #1173).

That being said, I'm just interested in the functionality to go into cookiecutter. So I'll leave it to the newer PR.

ssbarnea:
@nalch I know, the project lacked active maintenance for a very long time but hopefully this we are going to revive its community. Few months ago I was ready to scrap it from other projects where I was using due to this but recently I become core and I can help a little bit keeping it afloat.

As you mentioned the original feature request, I added it to the description and also added a comment there. I will also reopen this until we can make a decision.

nalch:
No worries. Let's take the newer pull request instead to get the functionality in.

Issue #970, opened 2017-07-07
Use YAML for cookiecutter configuration


Varriount:
The current "configuration" file for cookiecutter templates uses the JSON format. While this is fine for simple templates, it quickly falls apart for more complex requirements:
    - Limited prompt capabilities (only single-line freeform input or single-choice selection)
    - No comments 
    - No multi-line strings (useful for long jinja templates)
    
Using YAML allows for [tags](http://www.yaml.org/spec/1.2/spec.html#id2761292) to be used, which allows for tagging a data type with extra information (such as whether a list should be multiple-choice or single-choice). YAML also allows for comments and multi-line strings (including heredocs)

Issues that might be better resolved using features from YAML:
    - #138 (Using tags to denote input type)
    - #126 (Using a tag to denote that an input is a boolean type)
    - #853 (Using tags to denote that an input is a sensitive string)
    - #433 (Using objects marked with a tag, to differentiate from regular maps)

luzfcb:
Comparison of configuration file languages: 
https://gist.github.com/njsmith/78f68204c5d969f8c8bc645ef77d4a8f

Varriount:
The main draw for YAML is the ability to tag pieces of data in the document. The next-best format would be TOML (which at least has comments and multi-line strings).
Using the INI file format would be the worst choice. Every project that uses it eventually runs into some sort of limitation (everything's a string or possibly a list, limited nested) and has to create hacks.

brutus:
I came here looking for YAML support, because writing JSON is so tedious. It's unfortunate to see YAML support was dropped, but understandable. 

I have to agree with @Varriount, though, INI support would be the worst choice as an alternative format — it is not standardized in any way and not that big of a gain in terms of writability. __TOML__ might be a way better alternative.

I think __HJSON__ might be an even better alternative. We stick with JSON and get a lot of nice writability improvements. Especially for those used to Python syntax. It's basically JSON for humans ;)

http://hjson.org/

techdragon:
TOML is likely to be a better long term choice since future versions of packaging tools such as `pip` and `setuptools` will be using TOML since it was chosen as the format for the future Pipfile feature.

Using TOML will mean at some point in the future you can rely on TOML support coming out of the box with the users Python install just like `pip` does now. (for current versions of python)

Varriount:
@techdragon Neat. Where are these plans for pip posted?

pfmoore:
[...]

PR #1136, opened 2019-01-04
Use yaml optionally

umeboshi2:
This PR addresses #970.
If a ```cookiecutter.yml``` is present and a ```cookiecutter.json``` file is absent, it will read the config from the yaml file.  If the json file is present, it will be used for config even if the yaml file exists.  AFAICT, the only foreseeable problem with this PR is that pyyaml is inluded in the requirements in ```setup.py```.  If you can accept an "optional requirements" file, I can move the dependency to that and add instructions in the readme.  Except for the requirement being listed, I have tried to make this feature as optional as possible.  There is also an ```import yaml``` statement in ```generate.py`` ` that can be placed in a "try except" statement if desired.

I have been using this branch for the last year successfully and just rebased from current master before making this PR.  All of the templates that I have created are using yaml for config and they have been working just fine.

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -37,6 +37,7 @@ def ordered_load(stream, Loader=yaml.Loader, object_pairs_hook=OrderedDict):
     ordered_load(stream, yaml.SafeLoader)"""
     class OrderedLoader(Loader):
         pass
+
     def construct_mapping(loader, node):
         loader.flatten_mapping(node)
         return object_pairs_hook(loader.construct_pairs(node))
@@ -118,7 +119,7 @@ def generate_context(context_file='cookiecutter.json', default_context=None,
         # FIXME handle potential errors
         with open(context_file) as file_handle:
             obj = ordered_load(file_handle, yaml.SafeLoader)
-        
+
     # Add the Python object to the context dictionary
     file_name = os.path.split(context_file)[1]
     file_stem = file_name.split('.')[0]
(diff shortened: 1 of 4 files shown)


pydanny:
The problem with this PR is that it's reliant on PyYAML. As of a few years ago, PyYAML could not be reliably installed on all platforms, something that is an absolute requirement for Cookiecutter. This is what predicated the creation of Poyo, a simple and pure python YAML parser.

If the installation issues for PyYAML on windows have been resolved, then we can consider this PR.

ssbarnea:
@pydanny PyYAML had a period where it lacked maintenance and that is why runamel was also created but those times are now gone. Considering https://github.com/hackebrot/poyo/issues/23 I am inclined to support more the use of pyyaml as is more mature.

Pure Python is a benefit but only when both projects are actively maintained. If we are to count popularity I would say that Pyyam is about 8x more popular.

Lets see if there are any installation issues on Windows with pyyaml.

insspb:
Hi @umeboshi2  and thank you for contribution. 
it will be greate to include it in nearest cookiecutter release, as it will be major refactoring.
Can you do three things with it:

1. Rebase to current master
2. Write documentation example with template in yaml. 
3. Write some tests.

ssbarnea:
@umeboshi2 Yep, keep it fresh and it will happen very soon. Many things changed with cookiecutter maintenance since you originally raised it.

lastthyme (approved):

kurtmckee:

PR #1151, opened 2019-02-19
Using custom context for hooks

andresaco:
Using an extended version of the rendering context, including the repository dir variable `_repo_dir`


insspb:
@andresaco thank you for contribution. 
But it will not be merged. This is not correct place to extend context. Check issue #1034 for correct implementation.

PR #1184, opened 2019-07-04
Add file:// as a suitable protocol (#1)

ivanovmg:
Add file:// as a suitable protocol

Required if the repo is stored in Windows server, or other shared folder.

Enables the following:
$ cookiecutter "file://path/to/server/repo.git"

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -49,9 +49,7 @@ repository::
     $ cookiecutter https://github.com/audreyr/cookiecutter-pypackage.git
     $ cookiecutter git+ssh://git@github.com/audreyr/cookiecutter-pypackage.git
     $ cookiecutter hg+ssh://hg@bitbucket.org/audreyr/cookiecutter-pypackage
-    $ cookiecutter file://path/to/server/repo.git
-
-
+    
 You will be prompted to enter a bunch of project config values. (These are
 defined in the project's `cookiecutter.json`.)
 
@@ -69,6 +67,11 @@ If you want to work with repos that are not hosted in github or bitbucket you ca
 type of repo that you want to use prepending `hg+` or `git+` to repo url::
 
     $ cookiecutter hg+https://example.com/repo
+
+In addition, one can provide a path to the cookiecutter stored
+on a local server::
+
+    $ cookiecutter file://server/folder/project.git
 
 Works with Zip files
 --------------------
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -49,6 +49,7 @@ repository::
     $ cookiecutter https://github.com/audreyr/cookiecutter-pypackage.git
     $ cookiecutter git+ssh://git@github.com/audreyr/cookiecutter-pypackage.git
     $ cookiecutter hg+ssh://hg@bitbucket.org/audreyr/cookiecutter-pypackage
+    $ cookiecutter file://path/to/server/repo.git
 
 
 You will be prompted to enter a bunch of project config values. (These are
--- a/cookiecutter/repository.py
+++ b/cookiecutter/repository.py
@@ -11,7 +11,8 @@ from .vcs import clone
 from .zipfile import unzip
 
 REPO_REGEX = re.compile(r"""
-((((git|hg)\+)?(git|ssh|https?):(//)?)  # something like git:// ssh:// etc.
+# something like git:// ssh:// file:// etc.
+((((git|hg)\+)?(git|ssh|file|https?):(//)?)
  |                                      # or
  (\w+@[\w\.]+)                          # something like user@...
 )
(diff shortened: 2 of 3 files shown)


docs/usage.rst
```diff
@@ -49,6 +49,7 @@ repository::
     $ cookiecutter https://github.com/audreyr/cookiecutter-pypackage.git
     $ cookiecutter git+ssh://git@github.com/audreyr/cookiecutter-pypackage.git
     $ cookiecutter hg+ssh://hg@bitbucket.org/audreyr/cookiecutter-pypackage
+    $ cookiecutter file://path/to/server/repo.git
```
ssbarnea: The file:// protocol requires 3 slashes as it works only with full paths (mostly due to security). 

I think that the examples should include correct file examples.

ivanovmg: Thank you for the feedback!

The problem is that when I try to clone repo using the command with 3 slashes, it leads to an error.

$ git clone file://server/folder/project.git
Cloning into 'project'...
fatal: 'C:/Program Files/Git/server/folder/project.git' does not appear to be a git repository
fatal: Could not read from remote repository.

For some reason it would search under the folder with git executable, rather than in the absolute path. When I use 2 slashes, like in my example, it works.
OS Windows 10, GitBash or cmd.

Technically, I just need to clone from this kind of path //server/folder/project.git. I figured out, that using the current code and simply allowing file:// (with 2 slashes) to be treated as an url, makes all work.

Maybe, you can suggest how we can tell cookiecutter to treat //server/folder/project.git as a remote location to make it clone from it?

Thank you in advance!

ssbarnea: Please check https://tools.ietf.org/html/rfc8089 and its original rfc. I am unable to see any valid use of `file://` because the first part after it would be a server, not a PATH. Pointing something to local machine woudl require 3 slashes.

ivanovmg: I should have mentioned that more clearly in the docs example. My intention was to use file://server/folder/project.git as I mentioned in my previous message. And it stays in accordance with what you have just mentioned.

Should I just update the example to file://server/folder/project.git ?

ssbarnea: Yeah, I think that using server is a good compromise. I just wanted to avoid confusing people into using incorrect file URIs, ones that are not portable.

OvalAdmin:
I'm trying to run cookiecutter against a local, version controlled, repository and the `--checkout BRANCH`  option seems to largely be ignored. Similarly if I invoke cookiecutter with `checkout="BRANCH"`. I suspect it is not picking up that the cookiecutter resides within a repository as I am passing in a local path for the repository argument. As a consequence I have tried a number of variants of the `file:///...folder` both with forward and backward slashes and double/triple leading slashes, My results are as follows :

```
e:\Python>cookiecutter file:///e:/python/biscuits --checkout python
A valid repository for "file:///e:/python/biscuits" could not be found in the following locations:
file:///e:/python/biscuits

e:\Python>cookiecutter file:///e:\python\biscuits --checkout python
A valid repository for "file:///e:\python\biscuits" could not be found in the following locations:
file:///e:\python\biscuits

e:\Python>cookiecutter file:///e:\\python\\biscuits --checkout python
A valid repository for "file:///e:\\python\\biscuits" could not be found in the following locations:
file:///e:\\python\\biscuits

e:\Python>cookiecutter file://e:/python/biscuits --checkout python
A valid repository for "file://e:/python/biscuits" could not be found in the following locations:
file://e:/python/biscuits

e:\Python>cookiecutter file://e:\python\biscuits --checkout python
A valid repository for "file://e:\python\biscuits" could not be found in the following locations:
file://e:\python\biscuits

e:\Python>cookiecutter file://e:\\python\\biscuits --checkout python
A valid repository for "file:///e:/python/biscuits" could not be found in the following locations:
file:///e:/python/biscuits
```

I am also uncertain of where the `.git` in `.../folder.git` originates from ? If I create a normal repository the result is `.../folder/.git` and if I push this into a "bare" repository there isn't a `.git` folder. Is this some naming convention that you use to identify folders under version control e.g. `../folder.git`, `../folder.svn`, `.../folder.cvs` etc. or is there some other format for a git folder that I'm unaware of ?

OvalAdmin:
After a quick skim through `cookiecutter/repository.py` it seems as though cookiecutter only tests for the presence of a local folder and does not inspect it for a versioning mechanism. I did notice that `REPO_REGEX` does match my use cases above and that the "repository" code should execute in preference to the "folder" but this does not appear to be happening and I fall through to the raised exception at the end of `determine_repo_dir`. 

*Aside: *  If I follow correctly the code to distinguishes `git|svn|cvs+file://some/path`, under version control, from `/some/path`, without version control. Perhaps it is possible to allow local paths, that are version controlled repositories, to be duplicated into the users' `.cookiecutter` folder in the same manner that a remote repository might be checked out. It seems the most natural mechanism given how cookiecutter normally behaves otherwise.

PR #1200, opened 2019-08-07
Add a random_ascii_string() global function.

orf:
Including a random string of some kind in a template is often a use case. Currently the workarounds are really quite annoying: You can use a post_hook to replace tokens in a file, but this is super hacky.

This feature has been requested a number of times, and I really think including a `random_ascii_string` global function would be really useful.

### DIFF
--- a/cookiecutter/extensions.py
+++ b/cookiecutter/extensions.py
@@ -3,6 +3,13 @@
 """Jinja2 extensions."""
 
 import json
+import string
+
+try:
+    # Python 3.6 and above
+    from secrets import choice
+except ImportError:
+    from random import choice
 
 from jinja2.ext import Extension
 
@@ -18,3 +25,19 @@ class JsonifyExtension(Extension):
             return json.dumps(obj, sort_keys=True, indent=4)
 
         environment.filters['jsonify'] = jsonify
+
+
+class RandomStringExtension(Extension):
+    """Jinja2 extension to create a random string."""
+
+    def __init__(self, environment):
+        super(RandomStringExtension, self).__init__(environment)
+
+        def random_ascii_string(length, punctuation=False):
+            corpus = string.ascii_letters
+            if punctuation:
+                corpus = "".join((string.ascii_letters, string.punctuation))
+            else:
+                corpus = string.ascii_letters
+            return "".join(choice(corpus) for _ in range(length))
+        environment.globals.update(random_ascii_string=random_ascii_string)
--- a/tests/test_generate_file.py
+++ b/tests/test_generate_file.py
@@ -83,6 +83,29 @@ def test_generate_file_jsonify_filter(env):
 
 
 @pytest.mark.usefixtures('remove_cheese_file')
+@pytest.mark.parametrize("length", (10, 40))
+@pytest.mark.parametrize("punctuation", (True, False))
+def test_generate_file_random_ascii_string(env, length, punctuation):
+    infile = 'tests/files/{{cookiecutter.random_string_file}}.txt'
+    data = {'random_string_file': 'cheese'}
+    context = {
+        "cookiecutter": data,
+        "length": length,
+        "punctuation": punctuation
+    }
+    generate.generate_file(
+        project_dir=".",
+        infile=infile,
+        context=context,
+        env=env
+    )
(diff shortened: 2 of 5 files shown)


cookiecutter/extensions.py
```diff
@@ -18,3 +20,17 @@ def jsonify(obj):
...
         environment.filters['jsonify'] = jsonify
+
+
+class RandomStringExtension(Extension):
+    """Jinja2 extension to create a random string."""
+
+    def __init__(self, environment):
+        super(RandomStringExtension, self).__init__(environment)
+
+        def random_ascii_string(length, punctuation=True):
```
insspb: I do not like punctuation inclusion by default. 
I am always expect more problems from special characters.

tests/test_generate_file.py
```diff
@@ -82,6 +82,23 @@ def test_generate_file_jsonify_filter(env):
         assert json.loads(generated_text) == data
 
 
+@pytest.mark.usefixtures('remove_cheese_file')
+@pytest.mark.parametrize("length", (10, 40))
```
insspb: you make matrix with 'length', but did not do with 'punctuation'

cookiecutter/extensions.py
```diff
@@ -18,3 +20,17 @@ def jsonify(obj):
...
+            corpus = string.ascii_letters
+            if punctuation:
```
insspb: I do not like reassignments at all. And ``+=`` with strings.
Consider: 
```python
if punctuation:
    corpus = "".join((string.ascii_letters, string.punctuation))
else:
    corpus = string.ascii_letters
```

insspb:
Oh, just forgot main problem. 
This should be documented in Advanced usage docs. In same place where time extension documented. 
With usage examples. 
Thank you. Let's merge it before 1.7.0

cookiecutter/extensions.py
orf: Ok 👍

tests/test_generate_file.py
orf: Done 👍

orf:
Done @insspb. I've updated the advanced usage page to cover the jsonify extension as well. Let me know what you think!

insspb:
@orf Thank you for your efforts. I made a pull request to your repo after manual check of this plugin. Please merge it, msg me, and I will mark it as needs review and needs-merge.

orf:
Thanks for your MR @insspb. I've left a comment: I'm against not using the 'secrets' module here. Please remove that deletion and I'll merge, and we can get this in for 1.7.0.

insspb:
@orf Ok, not big deal, maybe you right, changed in PR. 

P.S., By the way, please review #1216 here, as we have not enough reviewers for my big commit.

hackebrot:
Hi @orf! 👋 

Thank you for submitting a pull-request to cookiecutter!

You said that this feature has been requested a number of times and that workarounds for this are annoying to use. Can you please provide example use cases where template authors would want to create a generate a random string? Also have you considered publishing this extension and make it [available for the template](https://cookiecutter.readthedocs.io/en/latest/advanced/template_extensions.html#template-extensions) rather than shipping it with cookiecutter?

We need to be deliberate with adding new features to cookiecutter since we have limited bandwidth to maintain this project and care about backwards-compatibility, meaning that if we merge a feature we will most likely have to maintain it for quite some time. 🍪

orf:
The case that spurred this MR was generating an Elixir project. Elixir requires three secret keys embedded in a configuration file. The creator of the Elixir template had written a pre_hook in Python that would write three keys to three files, and a post_hook that would do some `sed` magic to add the random strings to the right places. This is absolute madness but after spending some time I couldn't find a better way.

Maintenance is a burden, as always, but this is a not uncommon use case. I don't have the time now, but I did some searching and there are several requests for this in the repo across the years.

Cookiecutter is used in more contexts than just Python, and all of our internal users of it install via homebrew. Asking them to install a third party Python package before running a template is a big hurdle that not everyone will be able to complete, and plus I'm not even sure if it's possible with Homebrew (it installs cookiecutter to an isolated virtualenv).

orf:
Any news on this? Can we get it merged soonish? Not sure when the release is, but I'd really love this to be in it.

PR #1203, opened 2019-08-10
Convert README from rst to md

millatidy:
Referencing [Issue](https://github.com/cookiecutter/cookiecutter/issues/1179). Added acknowledgment to @audreyr  and contributors.


setup.py
```diff
@@ -19,7 +19,7 @@
     os.system("git push --tags")
     sys.exit()
 
-with io.open('README.rst', 'r', encoding='utf-8') as readme_file:
+with io.open('README.md', 'r', encoding='utf-8') as readme_file:
```
luzfcb: ⚠️ Since Readme has been converted to markdown, it is also necessary to inform setuptools which markdown is being used. Just add `long_description_content_type='text/markdown'` to the `setup` (after line 47)

Read more here: https://packaging.python.org/guides/making-a-pypi-friendly-readme/

luzfcb: Fixed on e986419

PR #1205, opened 2019-08-10
Replace flake8 with black

millatidy:
Replace flake8 with black

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,9 @@ clean-pyc: ## Remove Python file artifacts
 clean: clean-tox clean-build clean-pyc ## Remove all file artifacts
 
 .PHONY: lint
-lint: ## Check code style with flake8
+lint: ## Check code style with black
 	@echo "+ $@"
-	@tox -e flake8
+	@tox -e black
 
 .PHONY: test
 test: ## Run tests quickly with the default Python
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -3,3 +3,4 @@ pytest-cov
 pytest-mock
 pytest-catchlog
 freezegun
+black
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,9 +8,7 @@ repos:
       - id: check-byte-order-marker
       - id: check-executables-have-shebangs
       - id: check-merge-conflict
-  - repo: https://gitlab.com/pycqa/flake8
-    rev: 3.7.8
+  - repo: https://gitlab.com/psf/black
+    rev: stable
     hooks:
-      - id: flake8
-        additional_dependencies:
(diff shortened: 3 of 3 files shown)


Makefile
```diff
@@ -37,9 +37,9 @@ clean-pyc: ## Remove Python file artifacts
 clean: clean-tox clean-build clean-pyc ## Remove all file artifacts
 
 .PHONY: lint
-lint: ## Check code style with flake8
+lint: ## Check code style with black
 	@echo "+ $@"
-	@tox -e flake8
+	@tox -e black
```
michaeljoseph: I don't _think_ we have a toxenv called `black`. `lint` maybe?

michaeljoseph:
Not entirely sure I agree with this PR, `flake8` can do a lot more things than just paint it black?

test_requirements.txt
```diff
@@ -3,3 +3,4 @@ pytest-cov
 pytest-mock
 pytest-catchlog
 freezegun
+black
```
michaeljoseph: We probably don't need this because `pre-commit` will install `black`?

hackebrot:
Hi @millatidy! 👋 

Thank you for your pull-request to Cookiecutter! 🍪

I'm afraid I'm with @michaeljoseph here. I'm happy to auto-format source code with black, but we will need to continue running flake8 as a code linter.

millatidy:
> Hi @millatidy! 👋
> 
> Thank you for your pull-request to Cookiecutter! 🍪
> 
> I'm afraid I'm with @michaeljoseph here. I'm happy to auto-format source code with black, but we will need to continue running flake8 as a code linter.

I noticed some PR's are failing due to vague linting issues. That is why I made this PR. How the the failing PR's be fixed for instance this PR?

insspb:
I played with black a little bit and have some thoughts about it. 
1. It is not style checker, it is formatter. So we should not change flake8 for it. Moreover in next versions of python we, I hope, will extend flake8 with mypy type checking plugin.
2. Implementation of black with default settings will require a complete codebase code change. That will make all existing pull requests problematic. Here is example of what black with do with one of our files, when black allowed to any change.
Screenshot from 2019-08-25 17-11-09
3. In other places, black sometimes do a good job. (With restricted quotes change) I like this type of formatting:
Screenshot from 2019-08-25 17-10-02
But if line not so long it can break already formatted to oneliner. I do not like it for lines where it was already formatted vertically.
Screenshot from 2019-08-25 17-10-24

So at the end of the day. this commit will break existing codebase and will require next commit to fix it. 
This commit is not ready to merge.
Black CAN be implemented as precommit hook on github, and only in this way. But this should be done when most current PRs are closed/merged.
I will close it. You can open an issue to discuss this feature.

PR #1185, opened 2019-07-04
Convert README and other non-Sphinx docs to Markdown

wagnernegrao:
Referencing the issue [1179](https://github.com/cookiecutter/cookiecutter/issues/1179)

I'm working to convert all files.

Files:
- [x] Authors
- [x] Backers
- [x] Contributing
- [x] History
- [x] Readme
 - [x] case_studie

Work in production!

PR #1209, opened 2019-08-14
Update README from rst to md

millatidy:
Convert Readme from rst to md


insspb:
Thank you for efforts, but this is duplicate work. The same done in #1185 
setup.py changes not included in #1185 at this point. If you remove all changes except setup.py and make new PR it will be merged with #1185 if will not be added at #1185 at merge time. 
Thank you for work anyway.

PR #1210, opened 2019-08-15
Convert JsonifyExtension class to a function.

yudori:
This PR converts the custom jinja2 extension called JsonifyExtension to a function to conform with the standard stated here: https://github.com/cookiecutter/cookiecutter/blob/master/CONTRIBUTING.rst#L165 which states that this project should prefer functions to classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -61,3 +61,7 @@ target/
 # PyCharm
 .idea/
 .env
+
+# vscode
+.vscode/
+
--- a/cookiecutter/environment.py
+++ b/cookiecutter/environment.py
@@ -3,6 +3,7 @@
 """Jinja2 environment and extensions loading."""
 
 from jinja2 import Environment, StrictUndefined
+from jinja2.utils import import_string
 
 from .exceptions import UnknownExtension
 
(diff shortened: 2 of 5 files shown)


michaeljoseph:
This seems fine, code-wise, but I'm not sure preferring functions to classes means we should rewrite existing code without getting additional benefits?

insspb:
Code looks good. But overall I do not think that this implementation makes all codebase easier to read and understand. Let's hear from other developers.

hackebrot:
Hi @yudori! 👋 

Thank you for your pull-request. Based on our [coding standards documentation](https://github.com/cookiecutter/cookiecutter/blob/master/CONTRIBUTING.rst#L165), you were right to assume that this jinja2 extension needed to me updated to be using functions instead of a class. In my opinion, this specific guideline applies to cookiecutter's own code, while the extension follows the [best practices of the jinja2 project](https://jinja.palletsprojects.com/en/2.10.x/extensions/#example-extension).

I'm sorry, but I will have to close your pull-request. I created #1214 to update the coding standards section in our contributing guide to make sure it doesn't happen again. 📝 

Thanks again for your work. I hope to see another pull-request from you @yudori! 🍪

PR #1213, opened 2019-08-22
removed pyramid pcreate and Paste link in docs

seanmavley:
This goes to fix this: https://github.com/cookiecutter/cookiecutter/issues/1204


insspb:
Thank you for notice. It will be implemented in #1185 before the merge.

PR #1215, opened 2019-08-26
Require pip 9.0.0 or newer for tox environments

hackebrot:
Fix [Python 2.7 test environment](https://ci.appveyor.com/project/cookiecutter/cookiecutter/branch/master/job/ec787v2rawykero1). See https://github.com/tox-dev/tox/issues/1406

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,7 @@ init:
 
 install:
   - "%PYTHON%/Scripts/easy_install -U pip"
-  - "%PYTHON%/Scripts/pip install tox"
+  - "%PYTHON%/Scripts/pip install -U tox virtualenv"
   - "%PYTHON%/Scripts/pip install wheel"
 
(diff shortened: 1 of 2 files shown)


insspb:
As more and more projects drop support of python 2.7 we will face such issues more frequently. 
If I understand correctly this PR does not fix it in appveyor.
I think we should make some sprint in September to close as many current issues as possible, release 1.7.0 and drop python 2.7 support after it.

hackebrot:
Thanks for the review @ssbarnea! 🍪

ssbarnea:
@insspb Lets not diverge, I hope we do not discuss dropping py27 before 2020. I use cookiecutter in molecule and I am sure I will not drop it until it ends EOL. Even after I am likely to keep it until I am forced to drop it. I am more likely to replace a library that drops support for it than dropping support for py27.

PR #1217, opened 2019-08-27
Update file mode for hooks.py to be not executable

hackebrot:
Our linter failed because hooks.py does not have a shebang. But I don't think it's supposed to be executable. Please let me know your thoughts! 🙇 

https://travis-ci.com/cookiecutter/cookiecutter/jobs/228509876

See also #1195


hackebrot:
Thanks for the review @insspb! 🍪

PR #1221, opened 2019-09-01
Use sys.executable when invoking python interpreter from tests

vincentbernat:
When we only have python3 installed, the test for missing argument is
failing because there is no "python" executable. Use `sys.executable`
instead. Also set environment correctly, like done in 7024d3b36176.


insspb:
LGTM

PR #1225, opened 2019-09-09
Don't install whichcraft on Py>=3.

anntzer:
Uses the second (setuptools>=18) approach documented at
https://hynek.me/articles/conditional-python-dependencies/
to make the dependency Py2-only.

### DIFF
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -7,7 +7,10 @@ import logging
 import os
 import subprocess
 
-from whichcraft import which
+try:
+    from shutil import which
+except ImportError:
+    from whichcraft import which
 
 from .exceptions import (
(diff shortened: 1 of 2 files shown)


cookiecutter/vcs.py
```diff
@@ -7,7 +7,10 @@
 import os
 import subprocess
 
-from whichcraft import which
+try:
+    from shutil import which
+except ImportError:
+    from whichcraft import which
```
michaeljoseph: @anntzer I'm a little confused by this, isn't the point of `whichcraft` that you _don't_ have to do this? 🤔

anntzer: I don't think saving three lines of code is worth an additional dependency, but ultimately it's not my call.

hackebrot:
@ssbarnea did you see @michaeljoseph's comment? 🤔

PR #1227, opened 2019-09-12
Allow specifying custom j2 delimiters in environment variables

dbevacqua:
First off, thank you for creating this excellent project!

We are templating a large number of pre-existing complex jinja2 templates and, to avoid having to use a workaround like the one described in https://github.com/cookiecutter/cookiecutter/issues/1068#issuecomment-406107802, we have added the ability to create the j2 [Environment](https://jinja.palletsprojects.com/en/2.10.x/api/#jinja2.Environment) with custom delimiters.

The approach in this PR is to source these overrides from the environment. For example, we run `cookiecutter` with:

```
J2_VARIABLE_START_STRING='<<'
J2_VARIABLE_END_STRING='>>'
J2_BLOCK_START_STRING='<%'
J2_BLOCK_END_STRING='%>'
J2_COMMENT_START_STRING='<#'
J2_COMMENT_END_STRING='#>'
```

We tried another approach using cookiecutter.json (similar to the way you already allow custom j2 modules to be loaded) but for various reasons (mainly that there are two places where the variable start and end delimiters are hard-coded from which config would not be easily accessible), we pursued this one.

We have not yet updated docs as we wanted to validate the general approach first.

### DIFF
--- a/cookiecutter/environment.py
+++ b/cookiecutter/environment.py
@@ -54,7 +54,8 @@ class ExtensionLoaderMixin(object):
         else:
             return [str(ext) for ext in extensions]
 
-class CustomDelimiterConfigurerMixin(object):
+
+class DelimiterMixin(object):
     """Mixin to initialize the Jinja2 Environment with custom delimiters
     """
     def __init__(self, **kwargs):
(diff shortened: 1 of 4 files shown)


Oliboy50:
I wonder why is this not available yet? (> 3 years after)

It seems to be a critical feature for this kind of tool 🤔

kurtmckee:

PR #1229, opened 2019-09-17
replace 'bdist_wheel' with 'sdist' from 'make clean' because it breaks pypi releases.

aidanmelen:
The issue can be seen on early versions of [awscli-bastion](https://pypi.org/project/awscli-bastion/0.5.0/).

### Error
We can reproduce the issue with running ``make clean`` with ``python setup.py bdist_wheel``:

```
$ make clean
$ make release
```

then install will fail

```
$ pip install awscli-bastion==0.5.0                                                                                                                                         master
Collecting awscli-bastion==0.5.0
  ERROR: Could not find a version that satisfies the requirement awscli-bastion==0.5.0 (from versions: 0.1.0.macosx-10.14-x86_64, 0.2.0.macosx-10.14-x86_64, 0.3.0.macosx-10.14-x86_64, 0.4.0.macosx-10.14-x86_64, 0.5.0.macosx-10.14-x86_64)
ERROR: No matching distribution found for awscli-bastion==0.5.0
```

### Solution

Edit ``make clean`` with ``python setup.py sdist``:

```
$ make clean
$ make release
```

then install will succeed

```
$ pip install awscli-bastion==0.6.0

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ submodules: ## Pull and update git submodules recursively
 .PHONY: release
 release: clean ## Package and upload release
 	@echo "+ $@"
-	@python setup.py sdist bdist_wheel
+	@python setup.py sdist 
 	@twine upload -r $(PYPI_SERVER) dist/*
 
 .PHONY: sdist


luzfcb:
@aidanmelen This issue is not related to cookiecutter itself but related to https://github.com/audreyr/cookiecutter-pypackage

luzfcb:
Anyway, the [wheel package](https://pythonwheels.com/) format [(PEP 427)](https://www.python.org/dev/peps/pep-0427/) is the standard and work side-by-side with the old egg packages format.

https://packaging.python.org/guides/distributing-packages-using-setuptools/#wheels

aidanmelen:
That is good to hear. Do you know why my releases were failing with bdist_wheel? why does it only work with sdist?

luzfcb:
@aidanmelen https://github.com/aidanmelen/awscli_bastion/commit/49ac464dbf311aafcb2bb6f514aefc80b4d5ad89#diff-b67911656ef5d18c4ae36cb6741b7965

This commit seems to introduce the bug in your code because of the argument expected by the [`wheel`](https://wheel.readthedocs.io/en/latest/) library is `bdist_wheel` and not [`bdist`](https://docs.python.org/3.7/distutils/builtdist.html#creating-built-distributions).
Is expected that the `wheel` package is installed and updated to the latest version.

Also, without debugging information, it's hard to give any help

aidanmelen:
thanks I appreciate that. That gives me enough context to sort my stuff out.

Issue #1166, opened 2019-05-02
Support for renaming the cookiecutter dict


trondhindenes:
### Description:
Is there any support for renaming the cookiecutter object to something shorter? I'm finding `cookiecutter` to be one of those words that are hard to type, and would like to use something like `cc` in Jinja templates instead

hackebrot:
@Granitosaurus, this is your first and last warning about [Code of Conduct][code-of-conduct] violations against other contributors to Cookiecutter. ⚠️

The item of the Code of Conduct you are specifically violating is:

- [Trolling or insulting/derogatory comments][code-of-conduct item]

A sample of your insulting/derogatory writing (screen captured) from this comment thread and your pull-request (#1242):

> It's absolutely mindboggling how this project survived so long with such an absurd gap in basic UX. 

> what the hell is this?

> alias "cc" to "cookiecutter" for templating and user's sanity

Any future violations will result in an immediate banning from this project. While we encourage debate, we refuse to condone this kind of ongoing abusive behavior. We wouldn't allow this kind of treatment as paid employees, and we certainly won't take it as volunteers for an unpaid project.

[code-of-conduct]: https://github.com/cookiecutter/cookiecutter/blob/master/CODE_OF_CONDUCT.md
[code-of-conduct item]: https://www.pypa.io/en/latest/code-of-conduct/

Granitosaurus:
@hackebrot I might have been a bit rough with my wording sorry. Been working on updating two huge cookiecutter templates and I guess typing cookiecutter a million times blew some of my sanity off lol. 

For real though, someone should look into merging my PR eitherway.

wadkar:
Personally, I am not worried about typing `cookiecutter` because you one reads a variable an order of magnitude more than one writes. Perhaps `explicit is better than implicit` can be considered useful here?

Not to disregard any usability concerns, I do like the third option suggested earlier, i.e.  `{{foo}}.{{bar}}` though I am not sure how it can be achieved without [surprising the user too much](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

As it stands, I don't mind the status-quo and use the cookiecutter variable in jinja2 template. As for providing a shorter alias or renaming, wouldn't something like `{%- cc = cookiecutter -%}` at the beginning of template work for those who want shorter/cryptic name?

Granitosaurus:
I've been using `cc` alias from the MR since I posted the merge request and never looked back.

[...]

Issue #822, opened 2016-09-09
Custom context template holder name


urosjarc:
Currently you can't replace `{{coockiecutter.something}}` with some shorter version like
`{{cc.something}}`... [see the link...](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/main.py#L80)

I would really like to discuss how to implement this feature to the coockiecutter, I can make pull requests to make this happend.

hackebrot:
Maybe you can add a comment to the discussion on #380?

urosjarc:
@hackebrot : I change my focus to yeoman generator.

hackebrot:
Okay, fair enough! ☺️ I'm closing as a duple of #380 then.

PR #1233, opened 2019-09-30
Parametrize namespace

miklosaubert:
This allows for specifying a different context namespace than the default `'cookiecutter'`, when using cookiecutter's methods from Python. The behaviour of the command-line tool is unchanged.

This should be a first step towards fixing issues #380 and #1166 .

Rationale:

- The code was already partially supporting custom namespaces, in the `generate_context` method of `generate.py`. The method takes a context file name as parameter, with default value `'cookiecutter.json'`, and then generates a context object, using the file stem as the namespace: https://github.com/cookiecutter/cookiecutter/blob/673f773bfaf591b056d977c4ab82b45d90dce11e/cookiecutter/generate.py#L104 So with a context file name "`foo.json`", we then end up with a namespace of "`foo`". This PR extends this idea by making other methods "symmetric" with that behaviour. Slight change: we decouple the namespace from the context file name stem, by also introducing a namespace parameter to the `generate_context` function.
- Changes are voluntarily kept minimal, to ensure backwards compatibility and to ease code review. The issue of aligning the CLI, to accept e.g. an extra `--namespace` flag would have to be addressed separately, after merging this.

The last commit (f6645be) is just some bonus documentation for VS Code users, and is unrelated to the main concern of this PR. I can remove it from the branch and make a new PR if preferred.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -160,6 +160,7 @@ def is_pirate(message):
 ```
 
 * Write new code in Python 3.
+* If you are using VS Code with the Python extension, it's a good idea to set it up to use `flake8` for linting and `black` for formatting. Once your virtualenv is set up, run `pip install flake8 black`. Then in VS Code, go to Settings > Workspace, look for "python linting flake8" and check the "Python > Linting: Flake8 Enabled" checkbox. Look for "python black" and make sure the "Python > Formatting: Provider" setting is set to "black". In the "Python > Formatting: Black Args", add the following command-line arguments: `--skip-string-normalization` and `--line-length 79` to make it conform to the above standard for string quotes, and to flake8 settings.
 
 ## Testing with tox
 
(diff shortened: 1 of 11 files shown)


insspb:
Hi @miklosaubert , thank you for contribution. 

I am thinking about this issue for a while, when was worked on 1.7.x

Many PRs related to namespace variables, and I am inviting all their contributors here for discussion.

Related issues and PRs: 

* #380 - Remove cookiecutter namespace discussion
* #822 - Custom context template holder name
* #1166 - Renaming cookiecutter dictionary
* #1242 - Alias "cc" to "cookiecutter"
* #1244 - Alias "cc" to "cookiecutter"

My main concern is that namespace cli or class setting has no sense. This option will require user to write `cookiecutter --namespace=cc` or something like this, just to make some template work on their side. Or will require us to hardcode some value in our internals, as 'cookiecutter' now.

As I see it now: Correct solution is to implement namespace naming based on raw variable in template. Like it done with [`_copy_without_render`](https://cookiecutter.readthedocs.io/en/latest/advanced/copy_without_render.html) or [`_new_lines`](https://cookiecutter.readthedocs.io/en/latest/advanced/new_line_characters.html)

So we need to implement `_namespace` inside template, and check it in mentioned `cookiecutter/cookiecutter/generate.py` and pass it in next stages as independent variable.

I will do it myself before 2.0 if nobody will not be able to do it before, but this implementation is another way of hardcodding, that should be avoided.

I propose to make any further discussion in original issue #380

PR #1239, opened 2019-10-08
Pep257

firejava:
fix pep257 and flake8 on remote.py


firejava:
duplicates work done in [1234] (https://github.com/cookiecutter/cookiecutter/pull/1234) which looks good.

PR #1240, opened 2019-10-09
Feature/local extensions

mwesterhof:
this is a proposed change to add a "_local_extensions" feature.
It will enable the use of an additional "_local_extensions" element in the cookiecutter.json file. This will have the exact same functionality as the "_extensions" element, except that it can be used to load jinja2 template extensions straight from the project template.

I'm not certain if this is the best approach to achieve this, but I'm happy to have some feedback, of course.

```json
{
    "directory_name": "directory_name",
    "module_name": "module_name",
    "_local_extensions": ["local_extensions.SomeExtension"]
}
```

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -75,7 +75,7 @@ def cookiecutter(
         password=password,
         directory=directory,
     )
-    import_patch = patch_import_path_for_repo(repo_dir)
+    import_patch = _patch_import_path_for_repo(repo_dir)
 
     template_name = os.path.basename(os.path.abspath(repo_dir))
 
@@ -127,7 +127,7 @@ def cookiecutter(
     return result
 
 
-class patch_import_path_for_repo:
+class _patch_import_path_for_repo:
     def __init__(self, repo_dir):
         self._repo_dir = repo_dir
         self._path = None
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -4,8 +4,11 @@ Main entry point for the `cookiecutter` command.
 The code in this module is also a good example of how to use Cookiecutter as a
 library rather than a script.
 """
+from __future__ import unicode_literals
+from copy import copy
 import logging
 import os
+import sys
 
 from cookiecutter.config import get_user_config
 from cookiecutter.exceptions import InvalidModeException
@@ -72,15 +75,17 @@ def cookiecutter(
         password=password,
         directory=directory,
     )
+    import_patch = patch_import_path_for_repo(repo_dir)
 
     template_name = os.path.basename(os.path.abspath(repo_dir))
 
     if replay:
-        if isinstance(replay, bool):
-            context = load(config_dict['replay_dir'], template_name)
-        else:
-            path, template_name = os.path.split(os.path.splitext(replay)[0])
-            context = load(path, template_name)
+        with import_patch:
+            if isinstance(replay, bool):
+                context = load(config_dict['replay_dir'], template_name)
+            else:
+                path, template_name = os.path.split(os.path.splitext(replay)[0])
+                context = load(path, template_name)
     else:
         context_file = os.path.join(repo_dir, 'cookiecutter.json')
         logger.debug('context_file is %s', context_file)
@@ -93,7 +98,8 @@ def cookiecutter(
 
         # prompt the user to manually configure at the command line.
         # except when 'no-input' flag is set
-        context['cookiecutter'] = prompt_for_config(context, no_input)
+        with import_patch:
+            context['cookiecutter'] = prompt_for_config(context, no_input)
 
         # include template dir or url in the context dict
         context['cookiecutter']['_template'] = template
@@ -104,17 +110,31 @@ def cookiecutter(
         dump(config_dict['replay_dir'], template_name, context)
 
     # Create project from local context and project template.
-    result = generate_files(
-        repo_dir=repo_dir,
-        context=context,
-        overwrite_if_exists=overwrite_if_exists,
-        skip_if_file_exists=skip_if_file_exists,
-        output_dir=output_dir,
-        accept_hooks=accept_hooks,
-    )
+    with import_patch:
+        result = generate_files(
+            repo_dir=repo_dir,
+            context=context,
+            overwrite_if_exists=overwrite_if_exists,
+            skip_if_file_exists=skip_if_file_exists,
+            output_dir=output_dir,
+            accept_hooks=accept_hooks,
+        )
 
     # Cleanup (if required)
     if cleanup:
         rmtree(repo_dir)
 
     return result
+
+
+class patch_import_path_for_repo:
+    def __init__(self, repo_dir):
+        self._repo_dir = repo_dir
+        self._path = None
+
+    def __enter__(self):
+        self._path = copy(sys.path)
+        sys.path.append(self._repo_dir)
+
+    def __exit__(self, type, value, traceback):
+        sys.path = self._path
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -4,7 +4,6 @@ Main entry point for the `cookiecutter` command.
 The code in this module is also a good example of how to use Cookiecutter as a
 library rather than a script.
 """
-from __future__ import unicode_literals
 from copy import copy
 import logging
 import os
--- a/cookiecutter/utils.py
+++ b/cookiecutter/utils.py
@@ -8,6 +8,7 @@ import stat
 import sys
 
 from cookiecutter.prompt import read_user_yes_no
+from jinja2.ext import Extension
 
 logger = logging.getLogger(__name__)
 
@@ -105,3 +106,15 @@ def prompt_and_delete(path, no_input=False):
             return False
 
         sys.exit()
+
+
+def simple_filter(filter_function):
+    """Decorate a function to wrap it in a simplified jinja2 extension."""
+
+    class SimpleFilterExtension(Extension):
+        def __init__(self, environment):
+            super(SimpleFilterExtension, self).__init__(environment)
+            environment.filters[filter_function.__name__] = filter_function
+
+    SimpleFilterExtension.__name__ = filter_function.__name__
+    return SimpleFilterExtension
--- a/docs/advanced/local_extensions.rst
+++ b/docs/advanced/local_extensions.rst
@@ -1,0 +1,36 @@
+.. _`template extensions`:
+
+Local Extensions
+----------------
+
+*New in Cookiecutter X.x*
+
+A template may extend the Cookiecutter environment with local extensions.
+These can be part of the template itself, providing it with more sophisticated custom tags and filters.
+
+To do so, a template author must specify the required extensions in ``cookiecutter.json`` as follows:
+
+.. code-block:: json
+
+    {
+        "project_slug": "Foobar",
+        "year": "{% now 'utc', '%Y' %}",
+        "_local_extensions": ["local_extensions.FoobarExtension"]
+    }
+
+This example assumes that a ``local_extensions`` folder (python module) exists in the template root.
+It will contain a ``main.py`` file, containing the following (for instance):
+
+.. code-block:: python
+
+    # -*- coding: utf-8 -*-
+
+    from jinja2.ext import Extension
+
+
+    class FoobarExtension(Extension):
+        def __init__(self, environment):
+            super(FoobarExtension, self).__init__(environment)
+            environment.filters['foobar'] = lambda v: v * 2
+
+This will register the ``foobar`` filter for the template.
--- a/docs/advanced/local_extensions.rst
+++ b/docs/advanced/local_extensions.rst
@@ -15,7 +15,7 @@ To do so, a template author must specify the required extensions in ``cookiecutt
     {
         "project_slug": "Foobar",
         "year": "{% now 'utc', '%Y' %}",
-        "_local_extensions": ["local_extensions.FoobarExtension"]
+        "_extensions": ["local_extensions.FoobarExtension"]
     }
 
 This example assumes that a ``local_extensions`` folder (python module) exists in the template root.
@@ -43,7 +43,7 @@ as a filter. For this, we can use the ``simple_filter`` decorator:
     {
         "project_slug": "Foobar",
         "year": "{% now 'utc', '%Y' %}",
-        "_local_extensions": ["local_extensions.foobarextension"]
+        "_extensions": ["local_extensions.simplefilterextension"]
     }
 
 .. code-block:: python
--- a/docs/advanced/local_extensions.rst
+++ b/docs/advanced/local_extensions.rst
@@ -34,3 +34,27 @@ It will contain a ``main.py`` file, containing the following (for instance):
             environment.filters['foobar'] = lambda v: v * 2
 
 This will register the ``foobar`` filter for the template.
+
+For many cases, this will be unneccessarily complicated. It's likely that we'd only want to register a single function
+as a filter. For this, we can use the ``simple_filter`` decorator:
+
+.. code-block:: json
+
+    {
+        "project_slug": "Foobar",
+        "year": "{% now 'utc', '%Y' %}",
+        "_local_extensions": ["local_extensions.foobarextension"]
+    }
+
+.. code-block:: python
+
+    # -*- coding: utf-8 -*-
+
+    from cookiecutter.utils import simple_filter
+
+
+    @simple_filter
+    def simplefilterextension(v):
+        return v * 2
+
+This snippet will achieve the exact same result as the previous one.
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -9,6 +9,8 @@ from click.testing import CliRunner
 
 from cookiecutter import utils
 from cookiecutter.__main__ import main
+from cookiecutter.environment import StrictEnvironment
+from cookiecutter.exceptions import UnknownExtension
 from cookiecutter.main import cookiecutter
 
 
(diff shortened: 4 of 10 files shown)


croesnick:
I totally missed your PR, @mwesterhof, when quickly scanning the open PRs today for exactly this feature. While not being finally decided whether I like an implicit (silent) loading of local extensions or your more explicit approach via dedicated keyword, I definitely like your context-manager approach for prepending `repo_dir` to `sys.path`. 🙂

mwesterhof:
> I totally missed your PR, @mwesterhof, when quickly scanning the open PRs today for exactly this feature. While not being finally decided whether I like an implicit (silent) loading of local extensions or your more explicit approach via dedicated keyword, I definitely like your context-manager approach for prepending `repo_dir` to `sys.path`.

Thanks, it was a bit of a challenge to find the right way to implement this. I'd be very interested in this feature, though. When you decide on the way you want this to work, feel free to let me know. I'd be happy to make some changes

insspb:
@mwesterhof @croesnick thank you for contribution. 
As you both are active, we can speed up this feature implementation. 
First it should be well documented from point how user should it use.

mwesterhof:
@insspb @croesnick 
i added some docs for this feature to the "advanced" section. I haven't specified a version number yet, of course... should I leave that the way it is, or can i just replace the X.x with the next upcoming version?

con-f-use:
I'm sorry, it seems I've implemented a very similar thing in #1340

grahamalama:
Love this PR @mwesterhof! I was thinking about how to tackle the `modify config from a hook` problem that others were having, and basically came to the same idea that you did here. 

I had one idea that I thought might improve this implementation. I haven't fully thought it through, but I wanted to get your take:

As a `cookiecutter` user, I don't want to have to know / care about how to write a custom Jinja filter -- I just want to write a python function (or functions) that I can use in my `cookiecutter.json`.

What about instead of having users create a full-blown extension like:
```python
class FoobarExtension(Extension):
        super(FoobarExtension, self).__init__(environment)
         
        def foo():
             return "bar"
        
        environment.filters['foobar'] = foo
``` 

we had them create a module `filters.py` where they'd define a bunch of functions:
```python
def foo():
   return "bar"

def baz():
   return "qux"
```

and we could do something like:
```python
import inspect

import filters

functions = inspect.getmembers(filters, inspect.isfunction)
for function in functions:
    register_function_as_filter(function)
```

Thoughts on this?

mwesterhof:
it's an interesting take, @grahamalama 
I'm generally in favor of that kind of simplification, but i have a feeling it might be a bit too restrictive and too "magical".

I'm no expert on jinja, but i think we're ignoring a lot of the power of template filters by simplifying it this much. I'd propose a compromise:

```python
@simple_filter
def foo():
    return 'stuff'
```

1. creating fully fledged jinja-based filters is still possible
2. creating a simple function based one is possible
3. registering the function based filters is explicit
4. it's possible to add other (helper) functions to the module without them erroneously being picked up as filters

it actually mimics some ideas in the django framework regarding template filters and template tags, and it's an approach that i'm a big fan of.

Any opinions? I'd be happy to implement this behavior in my branch

con-f-use:
I still maintain, that just adding the extensions directory to the path and letting the user do whatever is the best, most simple and most generic solution.

mwesterhof:
@con-f-use i would agree. importing the decorator as an additional helper from cookiecutter.utils is exactly what i had in mind. As far as separating the local extensions from the "regular" ones goes... i'm sure this is somewhat of a personal thing, but I'm personally a fan of having a clear distinction between "extensions bundled with the template" and "installed extensions". That's just my view on it, but i'd be happy to change that. I'm not entirely sure who makes that call, but please let me know

mwesterhof:
@insspb @croesnick @con-f-use @grahamalama 
i just resolved a few conflicts, and I'm wondering if anyone has time to look at this MR. I'm more than happy to implement some changes, if necessary

gasbasd:
any news about this?

tests/test_cli.py
```diff
@@ -414,6 +414,19 @@ def test_echo_unknown_extension_error(tmpdir, cli_runner):
     assert 'Unable to load extension: ' in result.output
 
 
+def test_local_extension(tmpdir, cli_runner):
```
grahamalama, line 417: Would it be good to include tests for non-happy path? I can think of two cases, but maybe there are more:
- I include a filter in my template, but didn't include the extension in the `_local_extensions` list
- I include a filter in my template and extension in the `_local_extensions` list, but haven't defined that extension (or it's somehow not available to be imported)
  - It looks like there's a [similar test](https://github.com/cookiecutter/cookiecutter/blob/master/tests/test_environment.py#L8:L15) for built-in Jinja extensions

docs/advanced/local_extensions.rst
```diff
@@ -0,0 +1,36 @@
...
+.. code-block:: json
+
+    {
+        "project_slug": "Foobar",
+        "year": "{% now 'utc', '%Y' %}",
+        "_local_extensions": ["local_extensions.FoobarExtension"]
+    }
+
+This example assumes that a ``local_extensions`` folder (python module) exists in the template root.
+It will contain a ``main.py`` file, containing the following (for instance):
```
grahamalama, line 22: If I only need one or two extensions and they're not that complicated, I can just define them directly in a module `local_extensions.py`, correct? Maybe instead of having the documentation suggest a more-complicated package setup, it could suggest simply that you have a module `local_extensions` available in the template root with extensions defined, with no reference to a folder or `main.py`. That would leave it up to the user to use a standalone module or a package structure depending on their needs.

tests/test_cli.py
mwesterhof: that's a good idea. i'll pick that up ASAP

docs/advanced/local_extensions.rst
mwesterhof: that seems fair. i'll make the change

jonathan-s:
I was looking for exactly this. I'm glad that this in the works. My preference, would be to not introduce new naming such as the `_local_extensions` and instead use the already existing `_extensions`. From reading the code on how you are patching the python path that shouldn't be a problem.

docs/advanced/local_extensions.rst
jonathan-s: From what I gather reading the code you are adding the path of the local directory. So in effect you could create a folder that is named `extensions` that contains an `__init__.py` file. In that file you'll have the extensions and you'd refer to that in the cookiecutter.json file as `'extensions.MyExtension'` and it should all work.

jonathan-s:
@mwesterhof 
```
it actually mimics some ideas in the django framework regarding template filters and template tags, and it's an approach that i'm a big fan of.

Any opinions? I'd be happy to implement this behavior in my branch
```

I like the `@simplefilter` approach as well.

mwesterhof:
sorry guys, i've been quite sick recently and haven't gotten around to any work. I'll start working on this again though, so expect to see changes soon :muscle:

jonathan-s:
Mention me and I'll be happy to give feedback on any code changes. I hope
your illness isn't around any longer.

mwesterhof:
@grahamalama  @jonathan-s I updated my branch to include the simple_filter decorator, and rebased it. Can you guys take a look at it and let me know what you think about it? I'll try to check out your comment about a simple local_extensions.py asap, @grahamalama, but i think it should work just fine. 
also, the point about the naming scheme is a valid one, please weigh in on this. I'd be happy to change it so everything can be defined under '_extensions', without needing a separate '_local_extensions', i'd just like some more input on it first.

I finally have some time to work on this, and I hope we can do something cool with this MR soon. Please let me know what you guys think :muscle:

cookiecutter/utils.py
```diff
@@ -105,3 +106,15 @@ def prompt_and_delete(path, no_input=False):
             return False
 
         sys.exit()
+
+
+def simple_filter(filter_function):
```
grahamalama, line 111: 😍

jonathan-s:
@mwesterhof The simplefilter looks 👌. As mentioned by @con-f-use, I also think it'd be better to use `extensions` rather than `local_extensions`. It'll be one thing less to remember for the user of cookiecutter. I don't see any major advantages in having `extensions` and `local_extensions` separate. Are there any compelling reasons for that?

mwesterhof:
@jonathan-s i think it's a valid question. I initially wanted to leave it separate, but looking at the endresult, i doubt there's a particularly good reason to keep it that way. It shouldn't be a problem to change it. I'll wait for the feedback, but I'll make the change locally in the mean time so I can verify that it works fine in all scenarios

con-f-use:
My vote should be obvious... **no** `_local_extension` :grin:

mwesterhof:
@con-f-use @grahamalama @jonathan-s i made this change, we're now using exclusively _extensions
please let me know if you agree, i can always revert it if necessary

jonathan-s:
@insspb, it would be great if you could take a look at this. In my opinion this is ready to be merged.

con-f-use:
Please merge this already, it's super-useful!

mwesterhof:
@jonathan-s based on his activity, i think insspb might be unavailable. Do you know if anyone else might be able to take a look at this in the mean time?

jonathan-s:
@ssbarnea Would it be possible for you to take a look at this?

mwesterhof:
@jonathan-s is there a plan C? :sweat_smile:

jonathan-s:
@mwesterhof Sometimes it can take a while in the open source world. It's frustrating I know. Hacktober is coming up in a month, so let's wait for that. It'll likely get some more activity going.

ssbarnea:
I am afraid I cannot help with this PR, mainly because I personally do not see the use-case for it. If understand correctly this would mean that you will be able to use cc only from a specific location and the result would be dependent on your setup. To be this seems to make tool usage harder. Wouldn't a use of plugin mechanisms like pluggy allow it to be easier to extend. Installing an extension would be done using pip, same as cc itself.

mwesterhof:
@ssbarnea to clarify the use case; primarily, i've come across quite a few cases where we would like to have a template extension that does something relatively simple, yet very specific for a template that we're maintaining. Essentially, we want more power in creating and maintaining a CC template. Of course, installing extensions through pip is, and should remain an option, but this PR allows for much, much more flexibility by bundling custom python code with the template

jonathan-s:
@ssbarnea, As described in this issue -> https://github.com/pydanny/cookiecutter-djangopackage/issues/966 I would like to use a custom template filter `split` in the cookiecutter. Without this patch it's not possible. 

> If understand correctly this would mean that you will be able to use cc only from a specific location and the result would be dependent on your setup. 

I fail to see how this would happen.

mwesterhof:
> @ssbarnea, As described in this issue -> [pydanny/cookiecutter-djangopackage#966](https://github.com/pydanny/cookiecutter-djangopackage/issues/966) I would like to use a custom template filter `split` in the cookiecutter. Without this patch it's not possible.
> 
> > If understand correctly this would mean that you will be able to use cc only from a specific location and the result would be dependent on your setup.
> 
> I fail to see how this would happen.

actually, i think this might be an important misunderstanding. to clarify how this works... a "local" extension is an extension that's *bundled* with the template itself. So no, it's absolutely not dependent on an individual setup. The extension would be both used by, and *provided by* the template.

con-f-use:
Yes, it actually makes things a lot easier for the enduser. At the company I work for, we have a number of usecases for this and actually run on a fork, currently, that already has this.

xvr-hlt:
Please merge when possible, this is something that would be really useful for our projects.

ssbarnea (approved):

ssbarnea:
@insspb Can you please have another look at this and make a decision, eventually a merge if it looks ok to you. Thanks.

indigoviolet:
@insspb will this be merged?

brayanjuls:
Hey guys, do you have a workaround to enable this feature while we wait for @insspb to merge this feature ?

jonathan-s:
The most sensible workaround would be to ask the user of the cookie cutter
template to install a forked cookiecutter, where the fork contains this
fix.

mwesterhof:
@ssbarnea @jonathan-s apologies if I seem awfully pushy on this, but @insspb doesn't seem available for review, and hasn't been since he was first mentioned in august last year. Considering the fact that this regards a relatively small but welcome feature that doesn't impact any of the current features, I'm wondering if there's a way to short-track this a bit

jonathan-s:
I don't have any power here. It'd be up to ssbarnea in that case.

con-f-use:
Friendly reminder to @ssbarnea that we'd still very much like to have this.

cookiecutter/main.py
```diff
@@ -4,8 +4,11 @@
 The code in this module is also a good example of how to use Cookiecutter as a
 library rather than a script.
 """
+from __future__ import unicode_literals
```
samj1912, line 7: Any reason for adding this?

cookiecutter/main.py
```diff
@@ -72,15 +75,17 @@ def cookiecutter(
         password=password,
         directory=directory,
     )
+    import_patch = _patch_import_path_for_repo(repo_dir)
```
samj1912, line 78: This seems like it would add all the python files at the same level as the template directory to the import path. That might cause unexpected issues for people. Would it be better to limit the scope here to just the local extensions directory?

docs/advanced/local_extensions.rst
```diff
@@ -0,0 +1,60 @@
...
+.. code-block:: json
+
+    {
+        "project_slug": "Foobar",
+        "year": "{% now 'utc', '%Y' %}",
+        "_extensions": ["local_extensions.FoobarExtension"]
+    }
+
+This example assumes that a ``local_extensions`` folder (python module) exists in the template root.
+It will contain a ``main.py`` file, containing the following (for instance):
```
samj1912, line 22: Should this be updated to reflect the fact that we are adding the folder to sys path instead and the extension names would be derived from the modules/paths in the repo dir instead? The main.py may throw people off.

cookiecutter/main.py
mwesterhof: good catch, it seems I added this because I needed it for an initial change. It's been over 1.5 years, and I guess I forgot about that. I'll remove it

cookiecutter/main.py
mwesterhof: While it makes some sense, I'm not sure it would be the best way to go. 
The general idea with the current implementation is that template extension loading will essentially continue to work the same way it always has, except that they will also be loaded from local sources. While the provided example shows extensions in the "local_extensions" folder, I'm not suggesting that the extensions need to be contained in that specific folder, or even that extensions should all exist in the same folder.

I'm not convinced that it's necessary to limit the scope, but I'll make the change if I haven't changed your mind. I assume that adding another config keyword doesn't cause any problems? In that case I would suggest changing this:

```
    {
        "project_slug": "Foobar",
        "year": "{% now 'utc', '%Y' %}",
        "_extensions": ["local_extensions.FoobarExtension"]
    }
```
to 
```
    {
        "project_slug": "Foobar",
        "year": "{% now 'utc', '%Y' %}",
        "_local_extensions_folders": ["local_extensions"],
        "_extensions": ["FoobarExtension"]
    }
```
any thoughts?

docs/advanced/local_extensions.rst
mwesterhof: I'm honestly not sure what it is you're referring to. This is meant to be an example involving a python module as a "container" for extensions, and that's literally what is being explained here.

Maybe if we're indeed going with the "local extension folder config option" idea, then maybe this example could be simplified a bit, but the basic principle is kept as simple as possible while still being pythonic, and the changes in main.py are only to facilitate the extension loading and make it all possible

jonathan-s:
Can we please get this in the 2.0 release?

con-f-use:
@audreyfeldroy Pretty please? :grin:

frankhuurman:
2 days ago I've opened up an issue requesting to use some sort of custom .py script or function while generating the cookiecutter project to set a value in the cookiecutter.json, I believe this feature is exactly what I'm looking for!
Is there any update on this?

This is the one feature needed to automate all the things with cookiecutter :stuck_out_tongue:

con-f-use:
Yeah, please merge this asap, I'd prefer not to wait another 2.5 years.

matrixise:
Hi, I hope the best for @insspb but unfortunately, there is no Github activities for his account since December 2021.

Could we review/merge this PR without him?

Thank you,

mwesterhof:
@insspb indeed doesn't seem available long term. I'm sorry to once again have to bother @ssbarnea , but is there any way to proceed with this PR without his input?

ssbarnea:
I was planning to merge it but I wanted to see the CI result first.

mwesterhof:
@ssbarnea it seems that black linting was added somewhere along the line. I accounted for that, and it looks like everything passes now

con-f-use:
I'm overjoyed :grin: Thanks ssabernea! :1st_place_medal:

frankhuurman:
Awesome, can't wait to use it cause this will really make things a lot more extensive!

Same question as @con-f-use, is there a planned release date(with updated Docs as well?).

connerxyz:
I'm eager make use of this feature too! Thanks.

docs/advanced/local_extensions.rst
alkatar21, line 22: ``main.py`` must be ``__init__.py``

con-f-use:
Awesome, now we just have to wait for a new release. Any idea when that might be?

grahamalama:
@con-f-use looks like it made it into [this release](https://github.com/cookiecutter/cookiecutter/releases/tag/2.1.0).

PR #1242, opened 2019-10-14
alias "cc" to "cookiecutter" for templating and user's sanity

Granitosaurus:
Simply add alias to `cookiecutter` prefix for templating as `cc` short version. 

As per https://github.com/cookiecutter/cookiecutter/issues/1166


hackebrot:
@Granitosaurus please see https://github.com/cookiecutter/cookiecutter/issues/1166#issuecomment-542329850 ⚠️

PR #1243, opened 2019-10-15
Update README.md

RynHack:
My contribution to Open Source.


insspb:
I have no words, but ok. Thank you.

insspb:
Without mentions in doc. because mentions is bigger fix than this one :)

PR #1244, opened 2019-10-16
Alias "cc" to "cookiecutter"

Granitosaurus:
as per https://github.com/cookiecutter/cookiecutter/issues/1166

### DIFF
--- a/cookiecutter/find.py
+++ b/cookiecutter/find.py
@@ -4,6 +4,7 @@
 
 import logging
 import os
+import re
 
 from .exceptions import NonTemplatedInputDirException
 
@@ -22,7 +23,7 @@ def find_template(repo_dir):
 
     project_template = None
     for item in repo_dir_contents:
-        if ('cookiecutter' in item or 'cc' in item) and '{{' in item and '}}' in item:
+        if re.search(r'{{(cookiecutter|cc).*}}', item):
             project_template = item
             break
 
--- a/cookiecutter/find.py
+++ b/cookiecutter/find.py
@@ -22,7 +22,7 @@ def find_template(repo_dir):
 
     project_template = None
     for item in repo_dir_contents:
-        if 'cookiecutter' in item and '{{' in item and '}}' in item:
+        if ('cookiecutter' in item or 'cc' in item) and '{{' in item and '}}' in item:
             project_template = item
             break
 
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -83,6 +83,7 @@ def cookiecutter(
 
         # include template dir or url in the context dict
         context['cookiecutter']['_template'] = template
+        context['cc'] = context['cookiecutter']
 
         dump(config_dict['replay_dir'], template_name, context)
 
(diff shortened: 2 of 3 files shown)


cookiecutter/find.py
```diff
@@ -22,7 +22,7 @@ def find_template(repo_dir):
 
     project_template = None
     for item in repo_dir_contents:
-        if 'cookiecutter' in item and '{{' in item and '}}' in item:
+        if ('cookiecutter' in item or 'cc' in item) and '{{' in item and '}}' in item:
```
insspb: Maybe better to combine it to shorter form? 
Hard to read so lot of in item

insspb:
@Granitosaurus thank you for your contribution. I like you idea and ready to test it but please first fix tests and review comment.

cookiecutter/find.py
Granitosaurus: Do you mind if we introduce regex here? -> `re.search(r'{{.*(cookiecutter|cc).*}}', item)`

Granitosaurus:
@insspb I'm really not sure why the tests are failing. None of the changes I've made seem to be responsible. 

```
        context_str = json.dumps(context, indent=4, sort_keys=True)
>       assert context_str in result.output
E       assert '{\n    "cookiecutter": {\n        "_template": "tests/undefined-variable/file-name/",\n        "github_username": "hackebrot",\n        "project_slug": "testproject"\n    }\n}' in 'Unable to create file \'{{cookiecutter.foobar}}\'\nError message: \'collections.OrderedDict object\' has no attribute...ined-variable/file-name/",\n        "github_username": "hackebrot",\n        "project_slug": "testproject"\n    }\n}\n'
E        +  where 'Unable to create file \'{{cookiecutter.foobar}}\'\nError message: \'collections.OrderedDict object\' has no attribute...ined-variable/file-name/",\n        "github_username": "hackebrot",\n        "project_slug": "testproject"\n    }\n}\n' = <Result SystemExit(1)>.output
```

insspb:
@Granitosaurus 

I am thinking about this issue for a while, when was worked on 1.7.x

Many PRs related to namespace variables, and I am inviting all their contributors here for discussion.

Related issues and PRs: 

* #380 - Remove cookiecutter namespace discussion
* #822 - Custom context template holder name
* #1166 - Renaming cookiecutter dictionary
* #1233 - Parametrize namespace
* #1242 - Alias "cc" to "cookiecutter"
* #1244 - Alias "cc" to "cookiecutter"

My main concern is that namespace cli or class setting has no sense. This option will require user to write `cookiecutter --namespace=cc` or something like this, just to make some template work on their side. Or will require us to hardcode some value in our internals, as 'cookiecutter' now.

As I see it now: Correct solution is to implement namespace naming based on raw variable in template. Like it done with [`_copy_without_render`](https://cookiecutter.readthedocs.io/en/latest/advanced/copy_without_render.html) or [`_new_lines`](https://cookiecutter.readthedocs.io/en/latest/advanced/new_line_characters.html)

So we need to implement `_namespace` inside template, and check it in mentioned `cookiecutter/cookiecutter/generate.py` and pass it in next stages as independent variable.

I will do it myself before 2.0 if nobody will not be able to do it before, but this implementation is another way of hardcodding, that should be avoided.

I propose to make any further discussion in original issue #380

This PR is closed, to keep all discussion in #380. 

I am inviting everybody to join.

PR #1276, opened 2019-12-28
Add cookiecutter-pandoc-paper

Benjamin-Lee:


insspb:
Hi @Benjamin-Lee, we decided to stop including new templates in core documentation. 
Please use github tags to share your template. Thank you

PR #1277, opened 2020-01-04
New cookiecutter-aiohttp-uvloop template in docs

osminogin:
I'm glad to present to the community my template for python3 projects based on asyncio/aiohttp with uvloop (ultra fast event loop implementation). 

Read why [uvloop: Blazing fast Python networking](https://magic.io/blog/uvloop-blazing-fast-python-networking/), *TLDR* 
`In fact, it is at least 2x faster than nodejs, gevent, as well as any other Python asynchronous framework. The performance of uvloop-based asyncio is close to that of Go programs.`

Successfully used on several projects for high-load Public API, mobile app backends, but universally like aiohttp himself. I hope this will be useful to someone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -248,6 +248,7 @@ Please avoid creating PRs for listing additional templates. We outsourced their 
 * [python-project-template](https://github.com/Kwpolska/python-project-template): A template for Python projects with sophisticated release automation.
 * [cookiecutter-anyblok-project](https://github.com/AnyBlok/cookiecutter-anyblok-project): A template for Anyblok based projects.
 * [cookiecutter-python-cli](https://github.com/xuanluong/cookiecutter-python-cli): A cookiecutter template for creating a Python CLI application using click.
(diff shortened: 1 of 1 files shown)


ssbarnea:
We are no longer listing templates in the readme, read https://github.com/cookiecutter/cookiecutter/blob/master/README.md#a-pantry-full-of-cookiecutters and just add the right topic to your repository on github.

osminogin:
I think the project is really dead.

osminogin:
Remove other templates from readme so segmented community members be in equal conditions. As developer I personally recomends that all just use yoman there project is alive and where projects templates can be useful.

osminogin:
Yoman have ecosystem - cookiecutter is not.

ssbarnea:
@osminogin With all due respect for alternatives, which I looked upon myself,  cookiecutter is not dead, is only in a transitional phase. Cookiecutter is python based and most its users like that, yoman is from JS world. Cookiecutter is a dependency of many python packages, which yoman will never be (unless there is a python port).

In number cookiecutter still has >11K start on github, while yoman does not even reach 1K.

Please read at least the project README when preparing a PR, the template listing was clearly documented there.

PS. About community..., we can do better, we messed few things like moving between chat platforms too often.

Issue #1263, opened 2019-12-23
PEP257 docstrings for file "./tests/conftest.py"


insspb:
Cover `./tests/conftest.py` file with docstrings and follow [PEP257](https://www.python.org/dev/peps/pep-0257/). We use [pydocstyle](https://pypi.org/project/pydocstyle/) for validation.

Current validation log:

```
./tests/conftest.py:3 at module level:
        D205: 1 blank line required between summary line and description (found 0)
./tests/conftest.py:3 at module level:
        D400: First line should end with a period (not 't')
./tests/conftest.py:25 in public function `backup_dir`:
        D103: Missing docstring in public function
./tests/conftest.py:38 in public function `restore_backup_dir`:
        D103: Missing docstring in public function
./tests/conftest.py:62 in public function `clean_system`:
        D202: No blank lines allowed after function docstring (found 1)
./tests/conftest.py:154 in public function `user_dir`:
        D400: First line should end with a period (not 'y')
./tests/conftest.py:160 in public function `user_config_data`:
        D205: 1 blank line required between summary line and description (found 0)
./tests/conftest.py:160 in public function `user_config_data`:
        D400: First line should end with a period (not 'e')
./tests/conftest.py:178 in public function `user_config_file`:
        D205: 1 blank line required between summary line and description (found 0)
./tests/conftest.py:178 in public function `user_config_file`:
        D400: First line should end with a period (not 'e')
./tests/conftest.py:193 in public function `disable_poyo_logging`:
        D103: Missing docstring in public function
```

Subtask for #742

PR #1278, opened 2020-01-08
Adopt pycodestyle

ssbarnea:
This is a first step towards full PEP257 compliance. It includes a
series of excludes in order to keep the change size manageable.

Please raise follow-ups which remove excludes one-by-one.

Related: #1263


ssbarnea:
Please review, I cannot do it myself.

PR #1291, opened 2020-01-20
Make template-local extensions available to cookiecutter executions

croesnick:
As I am currently writing and refactoring my first cookiecutter template, chances are quite good I am missing on some already existing functionality. 🙂 So please tell me if what I am trying to achieve is already possible.

### Situation

While refactoring my template, I tried extracting common functionality (like formatting and conversion functions) into custom extensions. An example of such a conversion would be to reformat a given project name into a valid Python module name (no leading non-word characters, no leading numbers, non-word characters replaced by underscores). I created an `extensions` below my template's root, made it a module (added `__init__.py`), and wrote said extensions. Finally, I added the extensions to my `cookiecutter.json`'s `_extensions` list; e.g., `"extensions": ["extensions.naming.ModuleNameExtension"]`.

### Issue

As `extensions` is not within my PYTHONPATH, cookiecutter fails to find my modules (custom extensions).

### Proposal

Auto-prepend `repo_dir` to `sys.path` in order to make custom Python modules (e.g., extensions) available to cookiecutter.

Downside: Custom modules might shadow other identically-named modules.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -8,12 +8,14 @@ library rather than a script.
 """
 
 from __future__ import unicode_literals
+
 import logging
 import os
+import sys
 
 from cookiecutter.config import get_user_config
-from cookiecutter.generate import generate_context, generate_files
 from cookiecutter.exceptions import InvalidModeException
+from cookiecutter.generate import generate_context, generate_files
 from cookiecutter.prompt import prompt_for_config
 from cookiecutter.replay import dump, load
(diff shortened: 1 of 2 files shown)


croesnick:
Duplicate of #1240

insspb:
@croesnick thank you for contribution, but as you mentioned above it is duplicate. 
Original issue already in plan.
Not sure about time forsing for original issue, but will look on it again, after dealing with current repo status.

PR #1293, opened 2020-01-30
Update replay.py json.dump indent for easy viewing

nicain:
This change will make the dumped json file in replay easier on the eyes, and grep/find-able


insspb:
@nicain nice. Approved.

PR #1308, opened 2020-02-05
Fixed the link to the Django similar projects page, there was an extr…

joshbangle:
…a hyphen that had broken the current link.


joshbangle:
Not sure why the percentage is at 99.87%, all I changed was a line in the Readme.

insspb:
Hi @joshbangle, thank you for contribution. I just found this pull request, but I already fixed same issue in #1342

PR #1309, opened 2020-02-10
Using XDG_CACHE_HOME

ligurio:
[XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) defines is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable $XDG_CACHE_HOME.

This commit change directories `cookiecutter_replay` and `cookiecutters` where cookiecutter stores its own data.

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,12 +101,12 @@ default_context:
 full_name: "Audrey Roy"
 email: "audreyr@gmail.com"
 github_username: "audreyr"
-cookiecutters_dir: "~/.cookiecutters/"
+cookiecutters_dir: "~/$XDG_CACHE_HOME/cookiecutters/"
 ```
 
-* Cookiecutters (cloned Cookiecutter project templates) are put into ``~/.cookiecutters/`` by default, or cookiecutters_dir if specified.
+* Cookiecutters (cloned Cookiecutter project templates) are put into ``$XDG_CACHE_HOME`` by default, or cookiecutters_dir if specified.
 
-* If you have already cloned a cookiecutter into ``~/.cookiecutters/``, you can reference it by directory name:
+* If you have already cloned a cookiecutter into ``$XDG_CACHE_HOME``, you can reference it by directory name:
 
 ```bash
 # Clone cookiecutter-pypackage
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -26,8 +26,8 @@ BUILTIN_ABBREVIATIONS = {
 }
 
 DEFAULT_CONFIG = {
-    'cookiecutters_dir': os.path.expanduser('~/.cookiecutters/'),
-    'replay_dir': os.path.expanduser('~/.cookiecutter_replay/'),
+    'cookiecutters_dir': os.path.expanduser('$XDG_CACHE_HOME/cookiecutters/'),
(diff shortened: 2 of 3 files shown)


ssbarnea:
Very good idea! I am waiting to see the CI green.

ssbarnea:
Can you please resolve conflicts?

hackebrot:
Hi @ligurio! 👋 

The environment variable `XDG_CACHE_HOME` does not seem to be set on my mac OS devices. I don't think we can merge this pull-request. 😟

luzfcb:
pip has implemented a module only to discover some user-related directories, like cache dir:

https://github.com/pypa/pip/blob/master/src/pip/_vendor/appdirs.py#L264-L322

ssbarnea:
A big red **warning**: never ever consider importing anything from pip module because the maintenance team stated multiple times that their own API is the command line. They do not support any kind import from pip. 

There is also a module doing the same on https://pypi.org/project/appdirs/ but I did not see it updated in long time and I do not want to go nodejs path where there is a package for everything. 

I used the `os.getenv("XDG_CACHE_HOME", os.path.expanduser("~/.cache"))` approach in lots of cross platform projects without any issues. My main development machine are MacOS based so I care a lot about not breaking on them.

insspb:
I do not have such variable on Ubuntu. What I do wrong? 
As i see it is archlinux stuff. 
I do not think that we should merge something, that will not work on different systems.

ssbarnea:
Did you read the specification paragraph? It does not state anywhere that the variable must be defined, it does clearly specify what default value should be used when not defined. In fact 99% of systems do not have it defined but programs will respect it if defined. In fact there is a 9/10 chance that any program creating files there will already use the variable. The exceptions are those the did not bother to read the spec. XDG sped got a lot of popularity in the last years and that part of the spec is already used by lots of tools that have nothing to do with X/GUI.

con-f-use:
Ehrm, it's a pretty widely adopted thing, not just  "achlinux stuff". Even Ubuntu implements part of the XDG specification, i.e. you will probably have `$XDG_DATA_DIRS` defined on your Ubuntu.

README.md
```diff
@@ -101,12 +101,12 @@ default_context:
 full_name: "Audrey Roy"
 email: "audreyr@gmail.com"
 github_username: "audreyr"
-cookiecutters_dir: "~/.cookiecutters/"
+cookiecutters_dir: "~/$XDG_CACHE_HOME/cookiecutters/"
 ```
 
-* Cookiecutters (cloned Cookiecutter project templates) are put into ``~/.cookiecutters/`` by default, or cookiecutters_dir if specified.
```
ssbarnea, line 107: Correct it to `$XDG_CACHE_HOME/cookiecutters/` and optionally include link to the XDG spec to sort two issues: what is the default when variable is undefined and also to allow reader to learn avout XDG spec.

ssbarnea:
Reopening it as I think it was closed due to a confusion, clearly it does not break CI covered systems for sure.

insspb:
Hi @ligurio, this taken into the work by core commiter @ssbarnea . his PR includes your changes, and you will be mentioned in contributors list. 
I will close this PR in flavor to #1355 
Please keep all discussion there.
Please review @ssbarnea changes, as you more involved in this topic.

ligurio:
@insspb Thanks!

PR #1322, opened 2020-03-04
Fix indentation of .cookiecutterrc in README.md

mhsekhavat:
Indent fields in `default_context` so that copy-pasting the snippet into `.cookiecutterrc` works.
Yaml is sensitive to indentation and it doesn't work as is now.


insspb:
@mhsekhavat thank you for contribution. Contribution mention restored in #1345

PR #1329, opened 2020-03-18
cpp-template

kassane:
cmake w/ conan


ssbarnea:
Read the docs, we no longer add new entries.

ssbarnea:
@kassane Please look at https://github.com/cookiecutter/cookiecutter/pull/1331/files -- as long you add the right topics your template will be visible.

Issue #1256, opened 2019-12-21
News. Development and release plan. 


insspb:
Two weeks ago I was given full permissions to this repository, as well as to readthedocs and pypy. 
I see that there is a very small interest in this project from developers, so several things will be changed soon. This is an announcement about these changes.

1. At the Sunday version `1.7.0` will be released to pypy. It will be the current master, with maybe some not-code related changes.
2. This will be the last release with full code reviews from several core committers. Just because it delays development.
3. The final release of `1.7.x` expected on December 31. Only bugfixes for  `1.7.0` will be reviews at this time. 

New year plan.

The history of this project shows two things.
1. It is self-full at this moment. I.e. it does the thing that supposes to do at this moment. All problems have some kinds of workarounds.
2. Waiting for a sponsor is non-effective.

So the decision:

Let's make something new. With honor and respect to the past.

1. From January 1st and later master branch will include new changes that can be non-compatible with the old version but should introduce new functions.
2. Fixes for the old version will be included in new releases of `1.7.x`, but only bugfixes for confirmed bugs will be included. All new development should be made in the new version `2.0.x`
3. Contributors list for versions 1.x and 2.x will be separated in the documentation.
4. Templates links will be removed from the main master page, as we cannot control templates live and security. I will try to create issue at every template branch, how to implement correct keywords for GitHub search.
5. Most of the issues and pull requests will be closed or merged to 2.0.0 without waiting for contributors. Only live work should be at issues/pr page.

Feel free to comment on this issue. Every thought will be included in the development plan.

insspb:
* Tag for 1.8.0 for next version where some not deep changes will be made, with still support of Python 2.7 (not sure about 2.7)
* Tag 2.0.0 is wide list of improvements and changes, some will be moved to 1.9 or to 2.1... The goal to make releases more frequently. Not one time in several years :)

ljvmiranda921:
Hello @insspb , should we expect a final release of 1.7.X today? 

Happy New Year and Happy Holidays!!

insspb:
Hi @ljvmiranda921, an excuse for missed message, there was a lot of work right before NY (I had a release to production on Jan 7)
There was no significant bugs in 1.7.0 and significant changes, so it was released to pip at Dec without pre-release stage.

insspb:
Version 1.7.1 released.
## 1.7.1 (2019-04-21)

[...]

PR #1339, opened 2020-04-08
Render variables starting with an underscore.

smoothml:
Currently, variables starting with an underscore (which are not shown to the user on project creation) are not rendered with Jinja2. If you want to create one of these variables based on other variables you cannot. For example, if you are creating a Python package repository and want to base the package name on the project name you might have a configuration like this

```python
{
    "project_name": "Project Name",
    "project_slug": "{{ cookiecutter.project_name|lower|replace(' ', '-') }}",
    "package_name": "{{ cookiecutter.project_name|lower|replace(' ', '_') }}",
}
```

Which could create a structure like this

```
project-name
├── Makefile
├── README.md
├── requirements.txt
└── src
    ├── project_name
    │   └── __init__.py
    ├── setup.py
    └── tests
        └── __init__.py
```

However, the user can change `project_slug` and `package_slug`. This may be desirable, but (as in my current situation) it may not be. You _could_ create post-generation hooks to fail project creation if these are changed, but this seems inelegant. This PR allows for the configuration to be specified like this

```python
{
    "project_name": "Project Name",
    "_project_slug": "{{ cookiecutter.project_name|lower|replace(' ', '-') }}",
    "_package_name": "{{ cookiecutter.project_name|lower|replace(' ', '_') }}",
}
```

The user then cannot change these variables, enforcing consistent naming.


con-f-use:
I like this very much, as I ran into that as a relatively new user. Perhaps you should add a test for that as well.

smoothml:
@con-f-use I'm glad you like it! You're right, I should add a test and some docs, though as far as I could see the underscore variables aren't explicitely tested for or documented (I didn't spend too long looking so may have missed it).

>have you checked how that interacts with the replay function / repo updater packages?

I haven't, but all the tests pass so I assume this is fine.

I'm not sure this will be merged given #1256 Perhaps this could go in the 2.0 release @insspb?

insspb:
Hi @smoothml 
First of all - thank you for contribution. I really like this idea. Why I do not done it myself? Haha :)
Then here is my thoughts about it.

1. Remaking how current underscore works can break current templates. Maybe somebody read this code before and used it as feature for example with numbers. In your case numbers will be converted to string.
Here is code for you, to reproduce what I am talking about:
```python
    def test_should_render_dict(self):
        """

        """
        context = {
            'cookiecutter': {
                'project_name': 'Slartibartfast',
                '_project_name': 1123,
                'details': {'other_name': '{{cookiecutter.project_name}}'},
            }
        }

        cookiecutter_dict = prompt.prompt_for_config(context, no_input=True)
        assert cookiecutter_dict == {
            'project_name': 'Slartibartfast',
            '_project_name': 1123,
            'details': {'other_name': u'Slartibartfast',},
        }
```
This code will be good with current implementation. Will fail with your proposal. 

So...

Let do double underscore for user rendered private variables + docs + test. And it can be included in nearest release after this will be done.

But I really like this idea. At least things like this will be shorter: https://github.com/ionelmc/cookiecutter-pylibrary/blob/master/cookiecutter.json

Currently included in 1.7.1, but as we are plan to make a release in few days maybe it will not be ready for that time. At this case i will move it to 2.0.

smoothml:
Thanks for your comments @insspb. Good point regarding numbers vs. strings. Double underscores work 👍 I'm not sure I'll be able to get this ready in the next few days so getting it in 2.0 is a good aim I think.

ssbarnea:
I like this change but please add some tests for it and we should merge it in next major release 1.8.0, not the patch one (1.7.x) due to the risk of breaking existing templates.

The risk of breaking templates is small and template authors should already have CI/CD configured to test their templates with cookiecutter from master, running jobs at least weekly. We will always wait at least one week before merging a breaking change and tagging a release or pre-release.

tests/test_prompt.py
```diff
@@ -260,6 +260,29 @@ def test_dont_prompt_for_private_context_var(self, monkeypatch):
         cookiecutter_dict = prompt.prompt_for_config(context)
         assert cookiecutter_dict == {'_copy_without_render': ['*.html']}
 
+    def test_render_hidden_variables(self):
```
insspb, line 263: Docstrings missed.

smoothml: Yup, thanks. I haven't finished yet!

smoothml: I should have made it a draft PR actually. I'll do that.

insspb:
@smoothml still draft?

smoothml:
Yeah, sorry. Been slammed the last couple of weeks. Just docs to go. Will try and get it done this weekend. Sorry for the delay!

smoothml:
@insspb @ssbarnea apologies for the delay. This is ready for review now.

con-f-use:
We seem to have a [regression here](https://github.com/cookiecutter/cookiecutter/issues/1831). This happens on the latest 2.1.1 version. @ssbarnea

Issue #1211, opened 2019-08-15
Where to put Jinja extension code?


pouledodue:
Hi
I put extension code in `cookiecutter.json`
```
{
"_extensions": ["super.SuperExtension"]
}
```

and I created the `super/super.py` and `super/__init__.py` files with:
```
from jinja2 import nodes
from jinja2.ext import Extension


class SuperExtension(Extension):
    # a set of names that trigger the extension.
    # tags = {'cache'}

    def __init__(self, environment):
        super(SuperExtension, self).__init__(environment)

        # add the defaults to the environment
        environment.extend(
            keep_trailing_newline=True
        )
```

However when I run the cookiecutter command it keeps telling me 
`cookiecutter.exceptions.UnknownExtension: Unable to load extension: No module named 'super'`

I copied the `super` directory everywhere I can think and it don't find it.
So my question is where do I need to put the `super` dir?

pydanny:
Super is a reserved python keyword. Try a different module name.

pouledodue:
Hi
I renamed all to `hello` and `hello.HelloExtension` yet I have the same error:
`ModuleNotFoundError: No module named 'hello'`

Where do I put the `hello` dir?

michaeljoseph:
@pouledodue you need to put the `hello` directory on your `PYTHONPATH`.

<img width="799" alt="image" src="https://user-images.githubusercontent.com/169933/63210202-bcdda280-c0eb-11e9-825f-e12deda415d1.png">

michaeljoseph:
An improvement to the [docs](https://cookiecutter.readthedocs.io/en/latest/advanced/template_extensions.html) would be appreciated 📖 ✍️ 🙏

pouledodue:
thanks

con-f-use:
Just a quick related question

Why is it, that the cookiecutter template directory is not added to path by default or at least an `extensions` directory therein? Lets say we have this structure and I have a custom extension in `my_custom_ext.py`, that I'd like to use.

```
$ tree .
.
├── cookiecutter.json
├── {{cookiecutter.package_name}}
│   └── ...
├── extensions
│   └── my_cumstom_ext.py
└── hooks
    ├── post_gen_project.py
    └── pre_gen_project.py

$ cat cookiecutter.json
{
    "_extensions": ["extensions.my_custom_ext"]
}
```

I don't want to tell my user to `export PYTHONPATH=/path/to/./extensions` or some such.
[...]

Issue #547, opened 2015-10-04
Define and use custom Jinja2 filters


chekunkov:
Custom filters can make templating even more flexible. E.g. if I want to create nicely formatted .rst file in new project I should do something like this:

```
{% for _ in cookiecutter.project_name %}={% endfor %}
{{ cookiecutter.project_name }}
{% for _ in cookiecutter.project_name %}={% endfor %}
```

to produce

```
====================
My cool project name
====================
```

Looks cumbersome, right? How cool would it be to do something like this instead:

```
{{ cookiecutter.project_name|rst_h1 }}
```

Unfortunately there's no way to access Jinja2 environment before rendering a template - thus no way to add custom filters. 

I was thinking about following approach - add `filters` directory, where developer can place python modules with custom filters, import functions from those modules after environment instantiation and update `env.filters`. This should be simple to implement and if you give me a green light I'll submit a PR.

P.S. Actually I would love to have some way to access environment directly - then it would be unnecessary to do stuff described above. But I'm not sure how it could be implemented - maybe there could be some convention, like `environment.py` file in template root with `update_environment(env)` function... Seems clumsy, don't like this idea too much, so please share your thoughts.

hackebrot:
Hi @chekunkov :wave: 

I like the idea of writing custom filters for **Cookiecutter**, yet I am not sure what would be the best way to integrate them. Having said that, I will definitely put some thought into it. 

On a side note: You could use [Jinja2's macros](http://jinja.pocoo.org/docs/dev/templates/#macros)  for the mentioned use case.

```
{%- macro rst_h1(text) -%}
{% for _ in text %}={% endfor %}
{{text}}
{% for _ in text %}={% endfor %}
{%- endmacro %}
```

```
{{ rst_h1('pytest-' + cookiecutter.plugin_name) }}
```

I created the a [branch](https://github.com/pytest-dev/cookiecutter-pytest-plugin/tree/template-macro-in-readme) that features a dedicated folder for macros, a hook that removes the folder after rendering the project and an import for the macros in a rst file.

```
{%- from "macros/rst" import doc_title, doc_subtitle -%}
```

chekunkov:
Hi @hackebrot 

Yeah, macros can do the trick for the case I've mentioned, and helps to follow DRY, yet it's not as powerful and clean as plain Python, and I don't see any reason to hide this power from users :wink: 

[...]

PR #1340, opened 2020-04-10
Support of `./extensions/` directory in template

con-f-use:
In `cookiecutter.json`, you can now use the `_extensions` keyword to import extensions for a new `./extensions` directory in the template (similar to `./hooks/`).

image

For some related discussion see e.g.: #1211 and #547

The [updated documentation](https://github.com/con-f-use/cookiecutter/blob/2bbea175558bf4b7fdb83630ec0d08c029d840b1/docs/advanced/template_extensions.rst) has a more detailed description.

I would be willing to write additional tests to test this functionality.

Also. the "new in version ???" from my documentation should be updated before merging and releasing, same as release notes obviously.


con-f-use:
As mentioned, this is similar to #1240 and #944 .

@ssbarnea @insspb 
I'm biased, but I'd gravitate toward my more simple solution that involves less code. For the same reason I'd also vote for using the _extensions keyword in the json for both local and installed extensions.

If anything, make the folder name in #1240 just extensions instead of local_extensions to go along with ./hooks/.

cookiecutter/config.py
```diff
@@ -62,7 +62,7 @@ def merge_configs(default, overwrite):
 def get_config(config_path):
     """Retrieve the config from the specified path, returning a config dict."""
     if not os.path.exists(config_path):
-        raise ConfigDoesNotExistException
+        raise ConfigDoesNotExistException(config_path)
```
insspb, line 65: Out of this issue and already changed.

docs/advanced/template_extensions.rst
```diff
@@ -15,14 +15,34 @@ To do so, a template author must specify the required extensions in ``cookiecutt
...
+        "_extensions": ["jinja2_time.TimeExtension", "myext.RegexReplace"]
     }
 
 On invocation Cookiecutter tries to import the extensions and add them to its environment respectively.
 
 In the above example, Cookiecutter provides the additional tag `now`_, after
 installing the `jinja2_time.TimeExtension`_ and enabling it in ``cookiecutter.json``.
 
+You can place small custom extensions in your cookiecutter's ``./extensions`` directory
+(*New in Cookiecutter ???*).
```
insspb, line 27: Please remove this line.

docs/advanced/template_extensions.rst
```diff
@@ -15,14 +15,34 @@ To do so, a template author must specify the required extensions in ``cookiecutt
...
-        "_extensions": ["jinja2_time.TimeExtension"]
+        "_extensions": ["jinja2_time.TimeExtension", "myext.RegexReplace"]
     }
 
 On invocation Cookiecutter tries to import the extensions and add them to its environment respectively.
 
 In the above example, Cookiecutter provides the additional tag `now`_, after
 installing the `jinja2_time.TimeExtension`_ and enabling it in ``cookiecutter.json``.
 
+You can place small custom extensions in your cookiecutter's ``./extensions`` directory
```
insspb, line 26:
```suggestion
You can place custom Jinja2 extensions in your cookiecutter's ``./extensions`` directory
```

docs/advanced/template_extensions.rst
insspb: (*New in Cookiecutter ???*).

cookiecutter/config.py
con-f-use: Yeah, it was useful for debugging during development. Can I just leave it in or does it conflict?

insspb:
@con-f-use  As I mention earlier some tests required. 
At least these behavior cases should be covered:
User requested an custom extension, extension exist and work.
User requested an custom extension, but extension not exist in `extension` folder files.
User requested an custom extension, file/class exist in folder, but is nor an extension.

This feature can be useful for some part of users, who want to deliver own extensions without pip packages. But it is not just path extension.

con-f-use:
> User requested an custom extension, but extension not exist in `extension` folder files.
> User requested an custom extension, file/class exist in folder, but is nor an extension

I wouldn't worry about these two cases. The python import system will catch  the first one. 

For the second IMHO it is not necessary to check if it is an extension. If Jinja cannot find an extention, it will complain. If it is a python module or package instead of an extension than I'd not explicitly forbid that.

insspb:
@con-f-use I just found #1240 and seems it is more correct implementation of same thing and almost done. 
I will close this one and #944  as duplicates of #1240

PR #1343, opened 2020-04-13
Changed format of loggers invocation

### LINKED ISSUE #1307: Reformat log messages with variables
All logger messages, that contain variables should be reformatted to:
```
logger.level("message %s end of message", variable)
```
Why? :: https://reinout.vanrees.org/weblog/2015/06/05/logging-formatting.html

insspb:
Closes #1307


ssbarnea:
There are some serious performance impact related to use of string formatters on logging because previously the strings where rendered only when the logging level was appropiate. Now they are rendered all the time.

Still, that should not impact a tool like cookiecutter, which is not running as a service or doing lots of logging, so not really a big deal. I only wanted to mention it because on other projects it could be a real issue.

PR #1336, opened 2020-03-30
Add slugify template filter

ppanero:
Closes https://github.com/cookiecutter/cookiecutter/issues/1333

Added tests and docs.

PR #1344, opened 2020-04-13
Fix readme for #1336

insspb:
Restore contribution note for @ppanero


ppanero:
@insspb Thanks a lot!

PR #1355, opened 2020-04-14
Use XDG_CACHE_HOME for caching

ssbarnea:
[XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) defines is a single base directory relative to which user-specific non-essential (cached) data should be written. This directory is defined by the environment variable $XDG_CACHE_HOME.

This commit change directories `cookiecutter_replay` and `cookiecutters` where cookiecutter stores its own data.


insspb:
Related #104

ssbarnea:
@insspb There is a problem with python2.7 32bit from appveyor which failed twice without any explanation. When I added the `-v` to pytest to increase verbosity, it worked. 

I propose we keep the -v to see what happens in the future.

luzfcb:
> I propose we keep the -v to see what happens in the future.

+1 to this idea.

insspb:
@ssbarnea @luzfcb Agree with -v, about pull request, will review a little bit later.

ssbarnea:
I removed this feature from 1.7.1 release because it is introduding a breaking change: the
new caching directory will be different.

insspb:
* Related to #1309 
* Do not forget to mention @ligurio in project contributors.

luzfcb:
Any manual test was made on Windows?

ssbarnea:
@luzfcb I am not a big fun of either manual-testing or Windows platform, so you are welcomed to take this challenge. The good part is that we do hae CI/CD running on Windows, and it did not fail.

insspb:
@ssbarnea please check #971 for same ideas, but with external windows support

ssbarnea:
Closing as dead, I hope someone will implement it.

PR #1361, opened 2020-04-17
Reformatted readme file and removed outdated information

insspb:


insspb:
> Same problem as with the docstring change, too many unrelated changes squeezed inside a single PR, even if they are against a single file.
> 
> Unrelated changes should never be part of the same PR. If you do reformatting, you do not remove outdated project in the same change.

Here is only 10 lines really changed. All other is reformatting or ordering. Changed lines are dark.

PR #1365, opened 2020-04-17
Added: GitHub Action to build Sphinx docs and store artefacts

insspb:


.github/workflows/docs.yml
```diff
@@ -0,0 +1,26 @@
...
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: 'recursive'
+      - uses: actions/setup-python@v1
+        with:
+          python-version: "3.8"
```
ssbarnea, line 16: Use py36 for all non-unit jobs as this is the most stable and widlely supported version. py38 is known to keep causing problems with various platforms and libraries.

ssbarnea: My plan was to propose a single change that enable all currenlty travis jobs using github workdlows but I had not time to do it yet.

insspb: Switch to the same version of python, like on read the docs.

Issue #1305, opened 2020-02-02
Remove flake-black from pre-commit


insspb:
Black is black and after implementation can report its own changes. What for we need plugin before black implementation?

kishan3:
@insspb Should it be done or it is still open for discussion?

ssbarnea:
If I remember well, flake8-black change flake8 settings to make it black compatible but I think that at some point it will no longer be needed as I know that flake8 guys were working actively to achieve 100% compliance between the two. If this happened, ok to remove it.

PR #1368, opened 2020-04-17
Added: Black formatting verification for whole project

ssbarnea:
Adopts black formatting and runs it as part of the linting process. This should assure that code style is assured by black and we do not have to care about it anymore.

Fixes:  #1305


cookiecutter/generate.py
```diff
@@ -263,17 +267,10 @@ def generate_files(repo_dir, context=None, output_dir='.',
 
     unrendered_dir = os.path.split(template_dir)[1]
     ensure_dir_is_templated(unrendered_dir)
-    env = StrictEnvironment(
-        context=context,
-        keep_trailing_newline=True,
-    )
+    env = StrictEnvironment(context=context, keep_trailing_newline=True,)
```
insspb, line 270: Trailing commas in such places should be removed

ssbarnea: @inglesp I tried to remove the but some of them are added back by black itself.

inglesp: I think you meant @insspb...

ssbarnea:
@inglesp Can you please comment again? Github reports a comment from you which seems impossible to find.

inglesp:
> @inglesp Can you please comment again? Github reports a comment from you which seems impossible to find.

image

ssbarnea:
Thanks! That GH but is really annoying, the "hidden comments". I think last version got the commas sorted.

PR #1369, opened 2020-04-17
Tests refactoring: test_abbreviation_expansion tests

insspb:
Partial-Fix: #742


pfmoore:
@insspb Please don't keep requesting my review here, I no longer have sufficient time to work on this project.

PR #1380, opened 2020-04-17
tests: test_prompt Exception tests refactoring

insspb:


insspb:
Hi @pfmoore, no problem. Did not want to interrupt your current work. 
I just removed you write access to this project, so you will not pop-up in reviewers. 
Hope you ok with it. Msg me if you want access back.

PR #1382, opened 2020-04-18
readme: document specific topics

ssbarnea:
Documents how template authors can use github topics to make
their templates easily discoverable by others.

Note: many of these topics do not have any entries because template authors did not know about this. Still, to avoid a chick-n-egg issue, we start by documenting them first. Later we can ping some of them to ask them to update their github topics based on the documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,20 +138,44 @@ hope you find a cookiecutter that is just right for your needs.
 
 ## A Pantry Full of Cookiecutters
 
-The best place to start searching for specific and ready to use cookiecutter
-template is [Github search](https://github.com/search?q=cookiecutter&type=Repositories).
-Just type `cookiecutter` and you will discover over 4000 related repositories.
-
 We also recommend you to check related GitHub topics. For general search use
 [cookiecutter-template](https://github.com/topics/cookiecutter-template).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -138,20 +138,44 @@ hope you find a cookiecutter that is just right for your needs.
...
+If you fail to find any, you can always rely to standard
+[Github search](https://github.com/search?q=cookiecutter&type=Repositories).
 
 If you are template developer please add related
 [topics](https://help.github.com/en/github/administering-a-repository/classifying-your-repository-with-topics)
-with `cookiecutter` prefix to you repository. We believe it will make it more
-discoverable. You are almost not limited in topics amount, use it!
+with `cookiecutter` prefix to you repository. Assign at least `cookiecutter-template`
+topics and a more specific one from the list below. `cookiecutter-other` should also be
+added if your template does not fit any of the categories below:
```
insspb, line 152: After test it on myself I really dislike this Github topic introduction at all. This feature should not be first recommended approach. And i do not see any reason to have any links list in our readme.

I respect users, and sure that they can use both methods like they want. And I respect developers, and I am sure that they can specify topics like they want, we should only recommend some pattern, not complete/limited list.

PR #1395, opened 2020-05-20
Fixed: Replaced no longer maintained pipsi by pipx

ndclt:
Pipsi is no longer maintained (from the [pipsi github page](https://github.com/mitsuhiko/pipsi/)).


insspb:
@ndclt Thank you for contribution. Merged.

PR #1396, opened 2020-05-23
feat: don't call Popen for python hooks

paulpach:
I want to run cookiecutter with untrusted templates.

One way to do it is run cookiecutter in a sandbox environment such as the the ones in pypy, jython and ironpython.   In such sandbox, I can prevent the untrusted template from modifying files outside the generated project.

The challenge is that the hooks are invoked using subprocess.Popen.  Sandboxes cannot allow Popen, because the child process would escape the sandbox. 

This PR uses python's exec to run the hook code within the same process and remain sandboxed.

Note the tests don't pass because some of the tests invoke sys.exit(1),  which just exits cookiecutter itself.  Looking for feedback for ways to solve this.

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -64,6 +64,16 @@ def find_hook(hook_name, hooks_dir='hooks'):
 
     return None
 
+def run_python_script(script_path, cwd='.'):
+    """Execute a python script from a working directory.
+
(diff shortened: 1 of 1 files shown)


insspb:
Hi @paulpach, will you continue to work on this one? It is still in draft.

PR #1397, opened 2020-05-25
Added: Example for conditional files / directories in hooks

xyb:
fix https://github.com/cookiecutter/cookiecutter/issues/127#issuecomment-223635938


insspb:
@xyb Thank you for contribution.

PR #1401, opened 2020-05-26
py3: remove futures, six and encoding

insspb:


ssbarnea:
@insspb If you could split it by change-type it would be much easier to review and merge.

insspb:
@ssbarnea It is already splitted. for #1401, #1402 and many tests-related PR that will be done one by one. :)

I will merge this one in minutes, as it is already approved by you ;)

Issue #248, opened 2014-07-31
Encoding model cleanup


audreyfeldroy:
As @pfmoore said in #227:

> it looks like the sources make a lot of assumptions that files are encoded in utf-8. I guess that's the norm on Unix? It's typically not common on Windows, as the system default file encoding is not utf-8. I'm not entirely sure what to say as regards portability, though. "You should always use utf-8" is possibly a bit strong.
> 
> So far, I've been doing fixes in the code on a case by case basis, but it might actually be better to clean up the whole encoding model properly. Maybe let a template declare what encoding it uses and have cookiecutter read using that encoding and write using the user's preferred encoding (defaulting to the OS standard, but settable in .cookiecutterrc). That's harder to do, as the changes will affect a number of parts of the code, but it would likely be cleaner conceptually. I'll think about this a bit once all the changes I've been bombarding you with have settled down...

+1 on exploring this and cleaning up the encoding model. I'm starting a new ticket to track this. It may be good to prioritize this ticket early (0.7.3 or 0.8.1, depending on when @michaeljoseph's extra context work is ready), before we get too deep in new feature work. 

@pfmoore you have a stronger understanding of this than I, so feel free to take the lead here.

kurtmckee:
I'm closing this ticket due to its age.

It may be valuable to allow for custom encodings, but that can currently be accomplished using post-gen hooks that rewrite files.

PR #1414, opened 2020-05-29
Fixed: Values encoding on Windows

agateau:
When using `cookiecutter` on the cookiecutter I am working on (https://github.com/agateau/cookiecutter-qt-app) and using the default values except for selecting the MIT license, the author name appeared wrongly encoded both in the prompt default value and in the generated LICENSE file.

Explicitly setting the encoding to utf-8 when reading the context file fixes this.

This only happens on Windows for me: the added test pass with or without the fix on Linux, but fails without the fix on Windows.


insspb:
@agateau thank you for contribution.

You commit merged to master branch. You can test it there, as well as many features that will be available in cookiecutter 2.0

With cookiecutter 2.0 we want to extend windows support, so if you find any windows related bugs please report. 
Encoding problems discussion began in 2014 in #248 Please leave you thoughts there.

Also you can make a great help with review of existing windows related issues, and maybe introducing some bugfixes.

agateau:
Thanks for merging this fix. Unfortunately I am not a Windows expert: I do most of my work on Linux, I just want it to work on Windows. Getting that one-liner done took a long time (and much cursing!).

My 2 cents regarding encoding (again as a mostly Linux user, sometimes dabbling on other platforms) would be: standardize on every file being utf-8, but do not assume it is the default encoding on the OS, so always specify it.

PR #1419, opened 2020-05-30
Added: Jinja2 Environment extension on files generation stage

insspb:
This is code update for #946, as we do not have access to original branch. 

This PR implements option to change default Jinja2 settings by templates developers on template development stage.


bittner:
IIUC, this allows us to set (static) [Jinja2 environment](https://jinja.palletsprojects.com/en/2.11.x/api/#jinja2.Environment) values.

Non-simple values, such as `filters` and `tests`, which are dicts referring to Python code objects, won't be able to set like this. Hence, it still won't be possible to add custom filters "the Jinja way", by simply adding a function reference to the `filters` dict.

Is there a chance to add that feature before v2.0.0 is released?

PR #1445, opened 2020-08-21
Implement cloning before deleting existing template

ivanovmg:
This PR attempts to address https://github.com/cookiecutter/cookiecutter/issues/1444

When the template is downloaded and the user is asked whether the template should be re-downloaded, then the following steps are carried out.
1. The existing template is backed up in a temporary directory.
2. Try to download the new template
3. If failed to download the new template, then restore the backup.

Changes
---------

1. Function ``cookiecutter.utils.prompt_and_delete`` did both prompting the answer and deleting the existing template. I split function into ``prompt_ok_to_delete`` and ``prompt_ok_to_reuse``. Meanwhile the deletion of the existing template is carried out in ``cookiecutter.vcs``.
2. Refactor ``cookiecutter.vcs`` by extracting functions ``_need_to_clone``, ``_clone_repo``, ``_handle_clone_error``, ``_backup_and_delete_repo``, ``_restore_old_repo``.

Known issues
--------------

Code coverage dropped because I did not manage to implement proper tests with correct mocking of reading user input.
I appreciate if someone helps with that. The tests were created but marked to be skipped.

### DIFF
--- a/cookiecutter/utils.py
+++ b/cookiecutter/utils.py
@@ -5,7 +5,6 @@ import logging
 import os
 import shutil
 import stat
-import sys
 
 from cookiecutter.prompt import read_user_yes_no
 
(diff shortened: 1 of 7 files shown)


kurtmckee:
I'm closing this PR due to its age, and to avoid having cookiecutter increasingly manage the many VCS systems and corner cases that exist.

PR #1452, opened 2020-09-11
Retrieve project templates from S3

mdippery:
This commit adds optional support for allowing .zip files containing Cookiecutter project templates to be stored and accessed via S3. In short, project templates can be stored in a private S3 bucket instead of having to be publically available at an HTTP(S) endpoint. This support is achieved by using pre-signed S3 URLs, so that the mechanism for retrieving files from S3 is not dramatically different from downloading a file via HTTP(S), as is currently done.

Support for this feature is optional, since it requires the boto3 library, so this commit also adds an "s3" extra as an installation option. Support can be enabled by installing Cookiecutter using

    $ pip install cookiecutter[s3]

### DIFF
--- a/tests/zipfile/test_unzip.py
+++ b/tests/zipfile/test_unzip.py
@@ -1,6 +1,8 @@
 """Tests for function unzip() from zipfile module."""
 import tempfile
 
+import boto3
+import moto
 import pytest
 
 from cookiecutter import zipfile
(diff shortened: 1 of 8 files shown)


tests/zipfile/test_unzip.py
```diff
@@ -165,8 +167,18 @@ def test_bad_zip_file(mocker, tmpdir):
...
+    'url',
+    [
+        'https://example.com/path/to/fake-repo-tmpl.zip',
+        's3://example-bucket/path/to/fake-repo-tmpl.zip',
+    ],
+)
+@moto.mock_s3
+def test_unzip_url(mocker, tmpdir, url):
     """In `unzip()`, a url will be downloaded and unzipped."""
+    boto3.resource('s3').create_bucket(Bucket='example-bucket')
```
insspb, line 180: I remember from own projects that boto3 testing is something terrible, but it is not clear here what and why it is for https case. 
Please split normal and s3 tests

insspb (changes requested):
@mdippery 

1. Thank you for commit.
2. Excuse me long answer.
3. Documentation missing. Not clear on auth and what types supported. 
4. I am not completely sure that this option will be wide used.

kurtmckee:

PR #1459, opened 2020-10-02
Moved cli options to the top level menu

ghostcodekc:
Moved cli options to the top level menu as suggested in https://github.com/cookiecutter/cookiecutter/issues/1413#issuecomment-636306175


insspb:
@ghostcodekc please test merge requests before review request. This MR will not work.

PR #1470, opened 2020-10-28
Add slugify in the default extensions list

oncleben31:
To avoid new users to think they have to add it manually.


insspb (approved):

insspb:
@oncleben31 thanks. Merged.

PR #1481, opened 2020-12-14
Allow specifying a different hooks directory

roiavidan:
Currently, existing hooks are only sourced from within the root cookiecutter location under the hooks/ folder.
This commit adds a new CLI option to allow overriding that folder, if needed.
The rationale behind it is that one might want to have a generic or common set of pre/post hooks usable by all cookiecutters and sitting outside each cookiecutter template.


insspb:
Rebased to master

insspb:
@roiavidan thank you for contribution. Can you add small documentation note on this?

PR #1483, opened 2021-01-01
feat(config): allows user to read the cookiecutterrc file from various locations

pgierz:


cookiecutter/config.py
```diff
@@ -14,7 +14,34 @@
...
+    for path in paths:
+        if path:
+            path = _expand_path(path)
+            logger.debug("Trying to load: %s", path)
+            if os.path.exists(path):
+                return path
+    return None
+
+
+USER_CONFIG_PATH = _find_first_path(
```
jensens, line 37: Personally, I do not like execution of slightly more complex code on initialization time. This  function should be called in `merge_configs` below.

jensens (changes requested):
This is useful!
- See my comment 
- add documentation to https://github.com/cookiecutter/cookiecutter/blob/main/docs/advanced/user_config.rst
- test for `_find_first_path` would be good as well.

PR #1492, opened 2021-01-18
Accept a YAML replay file if the user specifically provides one.

dHannasch:
This is built atop https://github.com/cookiecutter/cookiecutter/pull/1490. Look there first. Once that is subtracted out, all this is doing is adding a conditional to `load_replay_file` that allows loading YAML. This could, if we wanted, actually be a one-line change, if we simply used a YAML parser always, since the YAML parser will also parse JSON.

The replay file is a useful feature, and it enables users to save and share the replay files. YAML files can be more human-readable than JSON files, since they can include comments.

This PR is very unambitious. This branch simply accepts a YAML replay file if the user specifically provides one; it does nothing else. It does not go looking for a YAML file if the user does not specifically provide one. It certainly does not go looking for a cookiecutter.yaml or any such thing. It just accepts a YAML replay file if the user provides one.


kurtmckee:
I'm closing this PR due to its age, and because replay files are generated by cookiecutter as JSON. If users are generating their own replay files they need to use JSON.

PR #1493, opened 2021-01-21
Added uuid extension to be able to generate uuids

### LINKED ISSUE #1101: Need help with generating GUID/UUID values for context variables
* Cookiecutter version: 1.6
* Template project url: none
* Python version: 3.7 (virtual env created using win python 3.7 x64)
* Operating System: Windows 10, 64 Bit

### Description:

First off many thanks for creating this project !
Here is some context of what I am trying to do and where I need some guidance
* I am trying to use CC to generate new a Visual Studio 2017 solution and project files with a particular folder/file organization that I like
* I was able to most of it working but for the below:
* Parts of the above project, solution files involves generating several unique GUIDs
* my first approach was creating a `pre_gen_project.py` inside the `hooks` folder and update/create new variables that could be added to the ones loaded from `cookiecutter.json` or entered by the user
* I was however blocked as I could not figure out how to access the context being used by CC and the jinja2 engine 
* I proceeded to go over the many issues on github and found some related ones like the following: #60, #102, #180, #288 but no clear answer on how to achieve what I'd like
* I also followed some others issues that suggested creating custom jinja2 extension/filter (#944) but I couldnt figure out how or where to put them in the template folder so the cookiecutter.exe can identify them and pick them up
* Lastly, I also tried going over the CC source code and tried to create a new executable from my script (similar to `cli.py`) that passes the guids via the `extra_context` to `cookiecutter.main(...)` but ran into some other problems that I am still trying to figure out

Appreciate any pointers on how I can inject GUID values for the context variables

jonaswre:
Fixes #1101 

I aiming to generate uuids from directly from jinja2. Ansible has an extension exactly like this.

These are only a few lines in non tighly coupled part. So this will not interfere with other parts of the application. 

I guess you cloud add a custom extension for this as well but to be honest this is such a basic functionality, far more basic then the slugify extension. 
A custom extension will always mean the cookiecutters are not very portable. Since the extension must be install as a python package on every client. (correct me if im wrong). 

Thanks for the consideration.


RobRoseKnows:
This fixes #1101 which is a big issue for C# cookiecutters. @cookiecutter could you get this reviewed?

jonaswre:
Any news? I kind of need this feature. 
Can a reviewer please review this. I wouldn't  mind paying for this pull request. If funding is the issue. 

Thanks

samj1912:
I think it might be better to review/merge https://github.com/cookiecutter/cookiecutter/pull/1240 instead which should provide this functionality and more.

ssbarnea (approved):

Garulf:
This extension is not included in the latest release `1.7.3` but is mentioned in the latest documentation leading to very confusing errors.

PR #1508, opened 2021-03-11
Added SVN support

Theta-Dev:
I have added SVN support for cookiecutter.

```
cookiecutter svn+https://mysvnrepo.com/path
```

Also refactored the VCS code to a object-oriented model, which should allow the addition of other VCS software if needed.

### DIFF
--- a/cookiecutter/repository.py
+++ b/cookiecutter/repository.py
@@ -3,13 +3,13 @@ import os
 import re
 
 from cookiecutter.exceptions import RepositoryNotFound
-from cookiecutter.vcs import clone
+from cookiecutter.vcs import clone, REPO_TYPES
 from cookiecutter.zipfile import unzip
 
 REPO_REGEX = re.compile(
     r"""
 # something like git:// ssh:// file:// etc.
-((((git|hg)\+)?(git|ssh|file|https?):(//)?)
+((((""" + '|'.join(REPO_TYPES.keys()) + r""")\+)?(git|ssh|file|https?):(//)?)
  |                                      # or
  (\w+@[\w\.]+)                          # something like user@...
 )
--- a/cookiecutter/repository.py
+++ b/cookiecutter/repository.py
@@ -11,7 +11,7 @@ REPO_REGEX = re.compile(
 # something like git:// ssh:// file:// etc.
 (((("""
     + '|'.join(REPO_TYPES.keys())
-    + r""")\+)?(git|ssh|file|https?):(//)?)
+    + r""")\+)?(git|svn|ssh|file|https?):(//)?)
  |                                      # or
  (\w+@[\w\.]+)                          # something like user@...
 )
--- a/cookiecutter/repository.py
+++ b/cookiecutter/repository.py
@@ -9,7 +9,9 @@ from cookiecutter.zipfile import unzip
 REPO_REGEX = re.compile(
     r"""
 # something like git:// ssh:// file:// etc.
-((((""" + '|'.join(REPO_TYPES.keys()) + r""")\+)?(git|ssh|file|https?):(//)?)
+(((("""
+    + '|'.join(REPO_TYPES.keys())
+    + r""")\+)?(git|ssh|file|https?):(//)?)
  |                                      # or
  (\w+@[\w\.]+)                          # something like user@...
 )
--- a/cookiecutter/repository.py
+++ b/cookiecutter/repository.py
@@ -3,26 +3,15 @@ import os
 import re
 
 from cookiecutter.exceptions import RepositoryNotFound
-from cookiecutter.vcs import clone, REPO_TYPES
+from cookiecutter.vcs import is_repo_url, clone
 from cookiecutter.zipfile import unzip
 
-REPO_REGEX = re.compile(
-    r"""
-# something like git:// ssh:// file:// etc.
-(((("""
-    + '|'.join(REPO_TYPES.keys())
-    + r""")\+)?(git|svn|ssh|file|https?):(//)?)
- |                                      # or
- (\w+@[\w\.]+)                          # something like user@...
-)
-""",
-    re.VERBOSE,
-)
+URL_REGEX = re.compile(r'''https?://''')
 
 
-def is_repo_url(value):
-    """Return True if value is a repository URL."""
-    return bool(REPO_REGEX.match(value))
+def is_url(value):
+    """Return True if value is a http/https URL."""
+    return bool(URL_REGEX.match(value))
 
 
 def is_zip_file(value):
@@ -97,7 +86,7 @@ def determine_repo_dir(
     if is_zip_file(template):
         unzipped_dir = unzip(
             zip_uri=template,
-            is_url=is_repo_url(template),
+            is_url=is_url(template),
             clone_to_dir=clone_to_dir,
             no_input=no_input,
             password=password,
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -3,6 +3,7 @@ import logging
 import os
 import subprocess  # nosec
 from shutil import which
+import re
 
 from cookiecutter.exceptions import (
     RepositoryCloneFailed,
@@ -99,6 +100,9 @@ class SVN(_VCS):
         :param repo_dir: Path of the cloned repository
         :raise subprocess.CalledProcessError: if the VCS returned an error
         """
+        # SVN SSH syntax: svn+ssh://private.com/myrepo
+        repo_url = re.sub(r'''^ssh://''', 'svn+ssh://', repo_url)
+
         command = [cls.cmd, 'export', repo_url]
 
         if checkout:
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -60,7 +60,7 @@ class VCS:
     @classmethod
     def match_repo_url(cls, url):
         """
-        Check if the raw URL input from the matches this VCS.
+        Check if the raw URL input matches this VCS.
 
         Return match level:
 
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -89,7 +89,7 @@ class VCS:
         if protocol and protocol not in cls.allowed_protocols + cls.protocols:
             return VCS.MatchLevel.NONE, ''
 
-        # If a VCS prefix is stated, it has to equal the VCS command
+        # If a VCS prefix is stated, it has to be equal to the VCS command
         # -> explicit match
         if prefix:
             if prefix == cls.cmd:
--- a/cookiecutter/vcs.py
+++ b/cookiecutter/vcs.py
@@ -18,6 +18,7 @@ logger = logging.getLogger(__name__)
 BRANCH_ERRORS = [
     'error: pathspec',
     'unknown revision',
+    'No such revision'
 ]
 
 
@@ -34,31 +35,17 @@ class _VCS:
 
     @classmethod
     def clone(cls, repo_url, checkout, clone_to_dir, repo_dir):
-        try:
+        subprocess.check_output(
+            [cls.cmd, 'clone', repo_url],
+            cwd=clone_to_dir,
+            stderr=subprocess.STDOUT,
+        )
+        if checkout is not None:
             subprocess.check_output(
-                [cls.cmd, 'clone', repo_url],
-                cwd=clone_to_dir,
+                [cls.cmd, 'checkout', checkout],
+                cwd=repo_dir,
                 stderr=subprocess.STDOUT,
             )
-            if checkout is not None:
-                subprocess.check_output(
-                    [cls.cmd, 'checkout', checkout],
-                    cwd=repo_dir,
-                    stderr=subprocess.STDOUT,
-                )
-        except subprocess.CalledProcessError as clone_error:
-            output = clone_error.output.decode('utf-8')
-            if 'not found' in output.lower():
-                raise RepositoryNotFound(
-                    'The repository {} could not be found, '
-                    'have you made a typo?'.format(repo_url)
-                )
-            if any(error in output for error in BRANCH_ERRORS):
-                raise RepositoryCloneFailed(
-                    'The {} branch of repository {} could not found, '
-                    'have you made a typo?'.format(checkout, repo_url)
-                )
-            raise
 
 
 class Git(_VCS):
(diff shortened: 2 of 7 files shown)


cookiecutter/repository.py
```diff
@@ -3,13 +3,15 @@
...
-from cookiecutter.vcs import clone
+from cookiecutter.vcs import clone, REPO_TYPES
 from cookiecutter.zipfile import unzip
 
 REPO_REGEX = re.compile(
     r"""
 # something like git:// ssh:// file:// etc.
-((((git|hg)\+)?(git|ssh|file|https?):(//)?)
+(((("""
+    + '|'.join(REPO_TYPES.keys())
```
simobasso, line 13: Hey @Theta-Dev, with this type of expansion, you are catching correctly `svn+<schema>` like :

```
svn+https://private.com/gitrepo
```

but not:

```
svn://private.com/gitrepo
```

you should consider to add `svn` also in the second regex group like `(git|svn|ssh|file|https?)`

and add some test case on 

https://github.com/cookiecutter/cookiecutter/blob/52dd18513bbab7f0fbfcb2938c9644d9092247cf/tests/repository/test_is_repo_url.py#L25-L38

simobasso:
Hey, @Theta-Dev, thanks for this!

I just wrote down a small doubt on the regex, and it seems that `black` fails on `cookiecutter/cookiecutter/vcs.py`.

Theta-Dev:
Hi @simobasso,

i'm glad you like my new feature.

But I'm not sure what you think the expected behavior when running `cookiecutter svn://private.com/svnrepo` should be.

svn:// is not a valid protocol like git://
You can do `git clone git://github.com/audreyfeldroy/cookiecutter-pypackage`
but you cant do `svn export svn://private.com/svnrepo`

The correct command for copying a local SVN repo is `svn export file:///home/thetadev/svnrepo` and for a remote one `svn export https://private.com/svnrepo`.

Thus the cookiecutter commands have to be `cookiecutter svn+file:///home/thetadev/svnrepo` and `cookiecutter svn+https://private.com/svnrepo`

hg isn't included in the protocol regex either. Probably for the same reason: it would result in an invalid `hg clone hg://private.com/hgrepo` command. I've never used hg though, so I can be 100% sure in that regard

I'll push a new commit where I fix the Black formatting issue and add `git://private.com/gitrepo` to the repo url test case.
But I dont think there is anything wrong with the protocol regex.

Theta-Dev:
YES. Finally the linter passes. I think I made the mistake to run black itself instead of the `tox -e lint` command before.

cjolowicz:
Just a quick note, hth :)

Here's the official list of supported URLs for Subversion repositories: http://svnbook.red-bean.com/en/1.7/svn.basic.in-action.html

Two points to note:

- `svn://` is the URL to a repository served via `svnserve`
- `svn+ssh://` is the URL for `svnserve` over SSH; note that the `svn+` must *not* be stripped off the URL

Theta-Dev:
Thank you, @cjolowicz.
I didnt know about svnserve, because I've only used svn with local repos and my company's repo, which is available via https.

Doing some research before coding probably isnt a bad idea.

**Expected behavior:**
cookiecutter svn+https:// => svn export https://
cookiecutter svn+file:// => svn export file://
cookiecutter svn:// => svn export svn://
cookiecutter svn+ssh:// => svn export svn+ssh://

Guess I'll have to do some more changes.

simobasso:
@cjolowicz, thanks for your notes!

@Theta-Dev looking forward to reviewing your pr!

Theta-Dev:
Okay, I think I'm done.
Cookiecutter SVN support now works with all 3 protocols. If you use SSH, you have to include your username followed by an @ and get prompted for your password.

```
$ cookiecutter svn+https://svn.code.sf.net/p/cookiecutter-pypackage/code/trunk/cookiecutter-pypackage
$ cookiecutter svn://svn.code.sf.net/p/cookiecutter-pypackage/code/trunk/cookiecutter-pypackage
$ cookiecutter svn+ssh://username@svn.code.sf.net/p/cookiecutter-pypackage/code/trunk/cookiecutter-pypackage
```

To test everything, I created a public SVN repo on SourceForge and put audreyfeldroy's pypackage cookiecutter there.
https://sourceforge.net/p/cookiecutter-pypackage/code/HEAD/tree/trunk/

You can see everything in action here:
https://asciinema.org/a/0OFMg7IFmLlj5vIdMUBku3CEB

cookiecutter/repository.py
simobasso: thx

simobasso (approved):
LGTM

simobasso:
@Theta-Dev, thank you, I just reviewed your pull request, and it seemed all good to me.

I don't have the merge permission, and there is the `codecov/project` check failed, but IMHO the culprit it's not your pr.

cookiecutter/vcs.py
```diff
@@ -115,6 +187,10 @@ def clone(repo_url, checkout=None, clone_to_dir='.', no_input=False):
                     'The {} branch of repository {} could not found, '
                     'have you made a typo?'.format(checkout, repo_url)
                 )
-            raise
+            raise RepositoryCloneFailed(
```
simobasso, line 190: Hey @Theta-Dev, I've just found out that can be a breaking change because no one traps the bare `subprocess.CalledProcessError`. This is useful to get the full traceback on an unhandled exception. What do you think about it?

Theta-Dev: What do you mean with "no one traps the bare CalledProcessError"?

If a known error is encountered, it outputs the respective error message. Otherwise it outputs a basic error message together with the raw error output of the VCS command.

```python
try:
    vcs.clone(repo_url, checkout, clone_to_dir, repo_dir)
except subprocess.CalledProcessError as clone_error:
    output = clone_error.output.decode('utf-8')
    if 'not found' in output.lower():
        raise RepositoryNotFound(
            'The repository {} could not be found, '
            'have you made a typo?'.format(repo_url)
        )
    if any(error in output for error in BRANCH_ERRORS):
        raise RepositoryCloneFailed(
            'The {} branch of repository {} could not found, '
            'have you made a typo?'.format(checkout, repo_url)
        )
    raise RepositoryCloneFailed(
        'Cloning of Repository {} returned an error:\n{}'.format(
            repo_url, output
        )
    )
```

Just like this:

```
(venv) thetadev@ThetaBook:~/Documents/Python/cookiecutter$ cookiecutter svn+https://xsvn.code.sf.net/p/cookiecutter-pypackage/code/trunk/cookiecutter-pypackages
Cloning of Repository https://xsvn.code.sf.net/p/cookiecutter-pypackage/code/trunk/cookiecutter-pypackages returned an error:
svn: E170013: Unable to connect to a repository at URL 'https://xsvn.code.sf.net/p/cookiecutter-pypackage/code/trunk/cookiecutter-pypackages'
svn: E670005: No address associated with hostname
```
Yes, the original version did output the full stack trace in this case. But is that really more debug and user friendly?

```
thetadev@ThetaBook:~$ cookiecutter kit@github.com:audreyfeldroy/cookiecutter-pypackage
  File "/usr/local/bin/cookiecutter", line 8, in <module>
    sys.exit(main())
  File "/usr/lib/python3/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/local/lib/python3.8/dist-packages/cookiecutter/cli.py", line 140, in main
    cookiecutter(
  File "/usr/local/lib/python3.8/dist-packages/cookiecutter/main.py", line 67, in cookiecutter
    repo_dir, cleanup = determine_repo_dir(
  File "/usr/local/lib/python3.8/dist-packages/cookiecutter/repository.py", line 110, in determine_repo_dir
    cloned_repo = clone(
  File "/usr/local/lib/python3.8/dist-packages/cookiecutter/vcs.py", line 101, in clone
    subprocess.check_output(
  File "/usr/lib/python3.8/subprocess.py", line 411, in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
  File "/usr/lib/python3.8/subprocess.py", line 512, in run
    raise CalledProcessError(retcode, process.args,
subprocess.CalledProcessError: Command '['git', 'clone', 'kit@github.com:audreyfeldroy/cookiecutter-pypackage']' returned non-zero exit status 128.
```
By raising the bare exception you wont see the error message from the VCS and have a hard time debugging it (if it's not obviouse, like this typo).
Plus it makes it seem like there is a problem with Cookiecutter itself, which is not the case.

simobasso: @Theta-Dev, thanks for your answer.

> Yes, the original version did output the full stack trace in this case. But is that really more debug and user friendly?

Of course not, but you should consider that a cookiecutter is a high used package; this means that someone can already expect the `subprocess.CalledProcessError` exception and do something when it happens. If we change this behaviour, we can break things.

What do you think about it?

cookiecutter/repository.py
```diff
@@ -3,13 +3,15 @@
...
+    + r""")\+)?(git|svn|ssh|file|https?):(//)?)
```
glumia, line 14: I would say to just write:
```python
REPO_REGEX = re.compile(
    r"((((git|hg|svn)\+)?(git|svn|ssh|file|https?):(//)?)"  # git:// git+https:// ...
    r"|(\w+@[\w\.]+))",  # or something like user@...
    re.VERBOSE,
)
```
It's simpler and more readable. @Theta-Dev what do you think?

(we should rework this regex in the future anyway, as it is right now it's overly permissive and allows no sense patterns like `git+svn://` etc)

tests/vcs/test_clone.py
```diff
@@ -67,55 +67,87 @@ def test_clone_should_abort_if_user_does_not_want_to_reclone(mocker, tmpdir):
...
+    )
 
     mock_subprocess = mocker.patch(
         'cookiecutter.vcs.subprocess.check_output', autospec=True,
     )
     expected_repo_dir = os.path.normpath(os.path.join(clone_dir, repo_name))
 
+    # repo_url defaults to input_url
+    if repo_url is None:
+        repo_url = input_url
```
glumia, line 116: I think it would be better to just write two times the same url in the `@pytest.mark.parametrize(` statement in those cases instead of this logic. In this way it's less clear that this is due to the particularity of svn that doesn't wants the `svn+` in the repo url.

glumia (changes requested):
Thanks for you work, that's great!

Apart from the things I wrote in the review, I have a small concern about the use of classes: it goes a bit against the philosophy of functions and modules pointed out in `CONTRIBUTING.md`.

I have nothing against them (even more in this case where they are used just as stateless entities to encapsulate related logic) but I don't feel confident to merge those changes without approval by one of the older mantainers.

Anyway, on my side it's a +1!

Theta-Dev:
I applied the requested changes.

In case of an unknown VCS error, Cookiecutter now prints the raw output while still throwing the subprocess error.
That should be more user-friendly while still backwards-compatible.

I got rid of the giant regex in the repository.py, too.

It got replaced with the VCS-specific ``VCS.match_repo_url`` function. It first checks if the URL and protocol is valid. Then it tries to match the ``git+``, ``svn+``, etc. prefixes, protocol and repo path suffix to the VCS specifications.
This can result in an explicit match -> the respective VCS gets chosen immediately.

Otherwise it tries to find an implicit match (VCS identifier like `git` or `bitbucket` present in repository path). An implicit match becomes the second choice if no explicit match was found.

Theta-Dev:
Oh, nevermind. Just went through the docs and found out that the gh syntax does not work with my new code.
Guess I'll have to do some more work tomorrow.

```
$ cookiecutter gh:audreyr/cookiecutter-pypackage
```

cookiecutter/vcs.py
```diff
@@ -15,47 +18,268 @@
...
+
+        :param url: Raw url parameter
+        :return: VCS.VCS.MatchLevel, repo_url (or None)
+        """
+        match = re.match(r'''(?:([\w\d]+)\+)?((\w+)://|\w+@)(\S+)''', url)
+
+        if not match:
+            return VCS.MatchLevel.NONE, ''
+
+        prefix, protocol_raw, protocol, repo_path = match.groups()
```
glumia, line 83: This naming doesn't fit the case of ssh urls, check for example `git@gitorious.org:cookiecutter-gitorious/cookiecutter-gitorious.git`, the result would be:
```python
prefix, protocol_raw, protocol, repo_path = (
    None,
    'git@',
    None,
    'gitorious.org:cookiecutter-gitorious/cookiecutter-gitorious.git',
)
```

Even if everything works (`Git.match_repo_url` returns the expected result) it's only a coincidence and this could lead to bugs if someone changes the logic of `VCS.match_repo_url`.

What do you think about a separate regex for each protocol (file, http, ssh, vcs specific)? 
(I'm looking at https://git-scm.com/book/en/v2/Git-on-the-Server-The-Protocols while I write this, if I'm correct on the other VCSs it should work in the same way.)

It would allow us to write simpler regexes (less probability to be exposed to ReDoS) and more explicit code.

glumia (commented):
Hey sorry for the big delay in the review, I've been a little busy in the last month. Check the last comments, great work by the way!

glumia:
Added the `breaking-change` label because of changes in the interface of `cookiecutter.vcs` module.

cookiecutter/vcs.py
Theta-Dev: Seems like a good idea. I'll look into rewriting that.

Another question: Cookiecutter matches Bitbucket urls to Hg automatically. As far as I know, Bitbucket switched from Hg to Git years ago. The current go-to for Hg hosting would be https://foss.heptapod.net. Shouldn't that be updated as well?

glumia: Uh, good point, I was totally unaware of that.

Yes, the `elif 'bitbucket' in repo_url:` in `cookiecutter/vcs.py:42` is wrong nowadays. Hmm I don't know if it's a good idea to keep this behaviour and replace it with `foss.heptapod`, I think this pattern of matching a vcs depending on the hosting domain is just wrong (like bitbucket the underlying vcs could change over time and we can't add and mantain such logic for every different hosting solution).

insspb:
@Theta-Dev Hi and thank you for contribution. 
I like object oriented approach in general, but unfortunately I had no time in 2021 to review this and several releases already was made. 
Can you rebase to current version and is you interested in continuing this feature?

cookiecutter/vcs.py
```diff
@@ -15,34 +18,263 @@
 logger = logging.getLogger(__name__)
 
 
-BRANCH_ERRORS = [
-    'error: pathspec',
-    'unknown revision',
+class VCS:
+    """Abstract base VCS class. Handles the cloning of VCS repositories."""
+
+    cmd = 'xyz'
```
att14, line 24: It seems more obvious to make this `None` and then raise an `AttributeError` in `__init__` or `__new__` if it is not set.

cookiecutter/vcs.py
```diff
@@ -15,34 +18,263 @@
...
+    """VCS command"""
```
att14, line 25: Are comments after the attribute definition standard for this repository?

alkatar21: It is at least standard Python and is used by editors and sphinx.

kurtmckee:

Issue #1514, opened 2021-04-10
Key error when using nested configuration files.


acturner:
* Cookiecutter version: 1.7.2
* Template project url: https://github.com/acturner/cookiecutter-key-error
* Python version: 3.8.9
* Operating System: macOS

### Description:

When using a nested configuration file, we get a key error because the key can't be found in the default `cookiecutter.json` config file (even when it exists). I will open a pull request with a one-line change that doesn't require the key to exist in the `cookiecutter.json` config, but I don't know why the key doesn't exist in the default.

### What I've run:

1. `git clone https://github.com/acturner/cookiecutter-key-error`
2. `cookiecutter cookiecutter-key-error --config-file=cookiecutter-key-error/init.yaml --no-input`

simobasso:
Hey @acturner, thanks for spotted out. Can you please check https://github.com/cookiecutter/cookiecutter/pull/1516 if fix your issue?

acturner:
@simobasso #1516 does fix this issue.

simobasso:
@acturner, thanks for the update; we should wait for a maintainer to merge #1516

simobasso:
@acturner just merged https://github.com/cookiecutter/cookiecutter/pull/1516 👍  let me know!

Issue #1513, opened 2021-04-10
Error reading yaml file using `poyo` - `ValueError: Parent of ChildMixin instance needs to be a Container.`


acturner:
* Cookiecutter version: 1.7.2
* Template project url: https://github.com/acturner/cookiecutter-value-error
* Python version: 3.8.9
* Operating System: macOS

### Description:

When using a nested configuration which includes a list (see the list of authors in [example template](https://github.com/acturner/cookiecutter-value-error)), we obtain an error from `poyo` simply trying to read the yaml file:

```
ValueError: Parent of ChildMixin instance needs to be a Container.
```

### What I've run:

1. `git clone https://github.com/acturner/cookiecutter-value-error`
3. `cookiecutter cookiecutter-value-error --config-file=cookiecutter-value-error/init.yaml --no-input`

simobasso:
Hey @acturner, thanks for spotted out. Can you please check https://github.com/cookiecutter/cookiecutter/pull/1516 if fix your issue?

acturner:
@simobasso #1516 does not fix this issue, unfortunately. The error comes from `poyo` and I don't know why (drop-in replacement of `pyyaml` for `poyo` here fixes the problem but I know we can't use `pyyaml`).

simobasso:
hey, @acturner thanks for the follow-up.

Yes, #1516 doesn't fix this issue.

it seems we have a lot of duplicated issue of the same bug, I'll try to link all of this issue so we can track better how to solve the issue at the root.

glumia:
I looked into it, it's a problem with poyo not being able to correctly parse the list in the yaml file. There's even an open issue for that (hackebrot/poyo#12) dating back to 2016.

+1 for switching to PyYAML, as @ samj1912 said in https://github.com/cookiecutter/cookiecutter/pull/1515#issuecomment-825285176 they now provide pre-compiled binaries for windows so cross-platform compatibility shouldn't be a problem anymore.

glumia:
Fixed with https://github.com/cookiecutter/cookiecutter/pull/1489.

PR #1515, opened 2021-04-10
`poyo` -> `pyyaml` (Issue #1513)

acturner:
I understand it may not be desirable to add `pyyaml` as a dependency, but these few changes do address both Issue #1513 and Issue #1514.


simobasso:
Hey @acturner acturner please check https://github.com/cookiecutter/cookiecutter/pull/1516

acturner:
@simobasso - Understood - unfortunately #1516 doesn't seem to fix these issues.

simobasso:
@acturner, thanks for reaching out. I just tried with your fork, and it works. Did you change anything?

acturner:
@simobasso Sorry I was mistaken, #1516 fixes #1514 - I had a similar solution, which doesn't require calling `merge_configs` when the key is missing (you can see in the diff). As far as fixing #1513, I just replaced `poyo` with `pyyaml`, which I know you said we can't use. It appears to be a problem with `poyo` ...

samj1912:
> Hey, @acturner, please check #1516.
> 
> Sadly `pyyaml` can't be introduced without dropping windows users, please have a look at [#1498 (comment)](https://github.com/cookiecutter/cookiecutter/pull/1498#issuecomment-771973584)

Hmm, AFAIK this should not be the case anymore. These yaml libraries now ship with wheels for windows so it should be possible to switch over. It is worth exploring moving to a better supported yaml library IMHO.

glumia:
I looked into it, it's a problem with `poyo` not being able to correctly parse the list in the yaml file. There's even an open issue for that (https://github.com/hackebrot/poyo/issues/12) dating back to 2016.

+1 for switching to `PyYAML`, as @ samj1912 said they now provide pre-compiled binaries for windows so cross-platform compatibility shouldn't be a problem anymore.

cookiecutter/config.py
```diff
@@ -44,7 +44,7 @@ def merge_configs(default, overwrite):
     for k, v in overwrite.items():
         # Make sure to preserve existing items in
         # nested dicts, for example `abbreviations`
-        if isinstance(v, dict):
+        if isinstance(v, dict) and k in default:
```
glumia, line 47: This is already being addressed with #1516, remove it please.

cookiecutter/config.py
```diff
@@ -60,13 +60,8 @@ def get_config(config_path):
...
 
     logger.debug('config_path is %s', config_path)
-    with open(config_path, encoding='utf-8') as file_handle:
-        try:
-            yaml_dict = poyo.parse_string(file_handle.read())
-        except poyo.exceptions.PoyoException as e:
-            raise InvalidConfiguration(
-                'Unable to parse YAML file {}. Error: {}'.format(config_path, e)
-            )
+    with open(config_path, 'r') as yaml_file:
```
glumia, line 63: Please specify the encoding, this is needed to avoid issues on some platforms where `utf-8` is not the default.

setup.py
```diff
@@ -11,6 +11,7 @@
     'binaryornot>=0.4.4',
     'Jinja2<3.0.0',
     'click>=7.0',
+    'pyyaml>=5.4',
     'poyo>=0.5.0',
```
glumia, line 15: We can remove this dependency if we switch to pyyaml.

cookiecutter/config.py
```diff
@@ -60,13 +60,8 @@ def get_config(config_path):
...
+        yaml_dict = yaml.load(yaml_file, Loader=yaml.FullLoader)
```
glumia, line 64: We should catch the `yaml.YAMLError` exception and raise a `InvalidConfiguration` exception in case of a malformed yaml file. 

Also, we should use the `yaml.SafeLoader` loader: we don't need the yaml features that it doesn't support (they weren't supported by poyo either).

glumia (changes requested):
Please remove also lines 179:183 in `tests/conftest.py`, they are not needed anymore if we don't use poyo.

acturner:
@glumia I believe I made all of your changes - let me know if anything doesn't look right. Cheers!

glumia (approved):
Top, lgtm!

acturner:
Updated title to reflect that this PR no longer fixes Issue #1514.

cookiecutter/config.py
```diff
@@ -60,10 +60,10 @@ def get_config(config_path):
         )
 
     logger.debug('config_path is %s', config_path)
-    with open(config_path, encoding='utf-8') as file_handle:
+    with open(config_path, encoding='utf-8') as yaml_file:
         try:
-            yaml_dict = poyo.parse_string(file_handle.read())
-        except poyo.exceptions.PoyoException as e:
+            yaml_dict = yaml.load(yaml_file, Loader=yaml.SafeLoader)
```
samj1912, line 65:
```suggestion
            yaml_dict = yaml.safe_load(yaml_file)
```

samj1912 (approved):

simobasso:
Hey @acturner, thanks for this pr; I just enabled the ci.

Could you please check windows failures? Also, have you seen https://github.com/cookiecutter/cookiecutter/pull/1489 opened by @dHannasch?

samj1912:
#1489 does seem like a similar implementation. It also fixes the test issues related to Windows builds.

acturner:
@simobasso @samj1912 I will check out #1489 soon and see what's going on with the Windows builds - also should probably add some tests for the bug this fixes (in case anyone makes a PR to use `poyo` again or something that breaks similarly)

ssbarnea:
I do support this change but I fixing windows builds is required before being able to merge it.

ssbarnea:
Merged https://github.com/cookiecutter/cookiecutter/pull/1489 which did pass the tests.

Issue #1149, opened 2019-02-19
[user config] exception with nested structures in default context


lbrack:
* Cookiecutter version: 1.6.0
* Template project url: https://github.com/audreyr/cookiecutter-pypackage.git (but irrelevant in this case)
* Python version: 3.6.2 (but also irrelevant)
* Operating System: OSX (but also irrelevant)

### Description:

After reading the section on [user config](https://cookiecutter.readthedocs.io/en/latest/advanced/user_config.html) and [dictionary variables](https://cookiecutter.readthedocs.io/en/latest/advanced/dict_variables.html) I was under the impression that defining nested YAML data structures under default_context would work (but it fails with a key error exception as is it attempting to access a key that doesn't exist in DEFAULT_CONFIG when iterating recursively through the dictionary resulting from the user config).

After looking at the code the root cause is pretty obvious (and so is the fix) but I am wondering if this is intended behavior (still it shouldn't fail with an exception). 

Just for the sake of clarity here is a minimalist user config that makes cookiecutter fail
```
default_context:
  project_name: my-project
  will_fail:
    on_this: pyops
```

### What I've run:

To reproduce this, copy the YAML example given above in user_config.yml and run
```cookiecutter https://github.com/audreyr/cookiecutter-pypackage.git --config-file=user_config.yml ```
The actual template is irrelevant since the failure occurs before it is even fetched. Here is the backtrace
```
Traceback (most recent call last):
  File "/usr/local/var/pyenv/versions/user_cfg_problem_3.6/bin/cookiecutter", line 11, in <module>
    sys.exit(main())
  File "/usr/local/var/pyenv/versions/user_cfg_problem_3.6/lib/python3.6/site-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/local/var/pyenv/versions/user_cfg_problem_3.6/lib/python3.6/site-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/local/var/pyenv/versions/user_cfg_problem_3.6/lib/python3.6/site-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/local/var/pyenv/versions/user_cfg_problem_3.6/lib/python3.6/site-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
[...]

PR #1516, opened 2021-04-11
config: set default on missing default_context key

simobasso:
Closes: https://github.com/cookiecutter/cookiecutter/issues/1514
Closes: https://github.com/cookiecutter/cookiecutter/issues/1149

### DIFF
--- a/cookiecutter/config.py
+++ b/cookiecutter/config.py
@@ -45,7 +45,7 @@ def merge_configs(default, overwrite):
         # Make sure to preserve existing items in
         # nested dicts, for example `abbreviations`
         if isinstance(v, dict):
-            new_config[k] = merge_configs(default[k], v)
+            new_config[k] = merge_configs(default.get(k, {}), v)
         else:
             new_config[k] = v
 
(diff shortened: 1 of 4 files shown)


acturner:
Does not yet fix #1513 or #1514 as far as I can tell.

acturner:
For testing, you can clone [`cookiecutter-key-error`](https://github.com/acturner/cookiecutter-key-error) and [`cookiecutter-value-error`](https://github.com/acturner/cookiecutter-value-error) and then run:
```shell
cookiecutter cookiecutter-key-error --config-file=cookiecutter-key-error/init.yaml --no-input
```
and
```shell
cookiecutter cookiecutter-value-error --config-file=cookiecutter-value-error/init.yaml --no-input
```
respectively.

simobasso:
@acturner thanks, just updated the description and also, I've found #1149 that seems the same issue.

glumia:
lgtm! 

The fact you had to update both `tests/test_get_config.py::test_get_config` and `tests/test_get_user_config.py::custom_config` after editing `tests/test-config/valid-config.yaml` is a clear sign of code duplication, we should have a single fixture as source of truth of file's content.  @simobasso do you think it would be worth to open a refactor issue for that?

simobasso:
Sure @glumia! Feel free to open an issue on this!

glumia (approved):

PR #1540, opened 2021-04-28
Added refresh option to CLI to always re-download template

justinessert:
This PR adds the feature described in [1201](https://github.com/cookiecutter/cookiecutter/issues/1201) and would help solve other issues like [1082](https://github.com/cookiecutter/cookiecutter/issues/1082).

I added a refresh option to the CLI so that there is a way to avoid the following prompt (and effectively just say yes):

```
You've downloaded ... before. Is it okay to delete and re-download it? [yes]:
```

I'm happy to add tests and update more docstrings, but I wanted to get an OK on the feature prior to making further changes.


justinessert:
Hello admins,

Can you please allow the pipeline to run on this PR?

Also, please let me know if I should fix the CodeClimate errors. There are errors for lines of code that I did not touch and the suggestions they give for the code that I did change seem like they would do more harm than good. That being said, I'm happy to take another look if desired.

justinessert:
@simobasso or other members,

Do you want me to fix the CodeClimate stuff? It seems like it would require some refactoring (for little, or negative, gain IMO). 

I'm willing to try refactoring but given the current simplicity of this PR, and that I don't necessarily agree with CodeClimate, I wanted to get a second opinion.

glumia:
Hi @justinessert, thanks for taking the time to work on this! Unfortunately, I'm going to close this pull request for the reasons explained in https://github.com/cookiecutter/cookiecutter/issues/1201#issuecomment-830597644. Sorry for the delay, we're trying to dispose of the pile of accumulated issues/pull requests, the situation should improve once finished :see_no_evil: 


Regarding CodeClimate it's nothing wrong with your code, the failed check you see is an issue we are already discussing in https://github.com/cookiecutter/cookiecutter/issues/1523#issuecomment-827721569.

Issue #1405, opened 2020-05-26
Tests refactoring: writing outside of temp dir


insspb:
* Change `os.path` usage to `pathlib`.
* Change `tmpdir` usage to `tmp_path`.
* Restrict write operations outside of `tmp_path`.

1 file, 1 PR

Related issues: #517 
Related PRs: #1355 (tests should be updated)

insspb:
@ssbarnea  Please do not close this one. 
There almost all tests should be checked and refactored, as many of them does not use tmp folder at all. I am traccking related PRs here.

ssbarnea:
@insspb I did not close this manually, it was closed by github due to a merge I did on another PR (magic keywords work on PRs too not only issues).

My goal was only to help you merge stuff that was ready. If something is WIP, be sure that you mark it as "Draft" (see convert-to-draft link top-right, bit hidden.)

SharpEdgeMarshall:
I'm trying to remove `os.path` from the codebase but I found that `os.path.expandvars` is not implemented in `pathlib`.

Do you know how we can handle it apart from keeping `os.path.expandvars`?

simobasso:
Hey @SharpEdgeMarshall 👋 ,

please have a look at https://bugs.python.org/issue21301#msg290510

> expanduser() is path handling function, but not expandvars(). expandvars() is defined in posixpath and ntpath because different OSes use different syntax for interpolated strings, and we sometimes need to use a foreign syntax. It doesn't have any special relations to paths, it handles them just as strings. Actually expandvars() functions are the fifth and the sixth ways of string formatting in Python.

what the drawback of keeping `os.path.expandvars`?

SharpEdgeMarshall:
@simobasso at the moment I'm keeping `os.path.expandvars`, I read it already and that's why I asked. 
I thought that the goal here was to remove `os.path` completely but i think that we could keep it in this case.

simobasso:
@SharpEdgeMarshall yes I think we can keep it.

Thanks

PR #1547, opened 2021-05-10
Convert os.path to pathlib

SharpEdgeMarshall:
Partially solves #1405


cookiecutter/main.py
```diff
@@ -77,17 +77,17 @@ def cookiecutter(
...
 
     if replay:
         with import_patch:
             if isinstance(replay, bool):
                 context = load(config_dict['replay_dir'], template_name)
             else:
-                path, template_name = os.path.split(os.path.splitext(replay)[0])
-                context = load(path, template_name)
+                replay = Path(replay)
+                context = load(str(path.parent), template_name)
```
jensens, line 88:
```suggestion
                context = load(str(replay.parent), template_name)
```
should solve failing test after merge

jensens:
Actually I would like to get this one in. But the conflict-merge seems to break tests. Probably I did something wrong. I need to look into it later, or anyone else to step in?

insspb:
@jensens I really do not like idea to merge it as is. So please do not merge this.

Such big changes I want at least in 2 merge requests. In 1 codebase only. (Tests should be stable, as we do not change app behavior).

In second - tests only.

jensens:
> Such big changes I want at least in 2 merge requests. In 1 codebase only. (Tests should be stable, as we do not change app behavior).
> 
> In second - tests only.

Indeed, this would be favorable.

kurtmckee:

PR #1549, opened 2021-05-12
Remove direct dependency on markupsafe

ssbarnea:
This sorts potential dependency conflicts with jinja2 which is the only place where markupsafe is used (since version 2.7)


simobasso (approved):
LGTM

simobasso:
Thanks @ssbarnea 🙏 !

PR #1548, opened 2021-05-12
allow jinja 3.0.0

wouterdb:
Jinja 3.0.0 has just been released 
https://jinja.palletsprojects.com/en/3.0.x/changes/#version-3-0-1

We depend on both cookiecutter and Jinja2, but cookiecutter is preventing us from updating Jinja2.
(see https://github.com/inmanta/inmanta-core/pull/2921)

I understand if you consider it too soon to merge this in.

PR #1550, opened 2021-05-14
Fix dependency conflicts

ssbarnea:
This backports changes already made to master via: #1549 and #1548

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
       - "*"
   pull_request:
     branches:
-      - "*"
+      - "**"
 
 jobs:
   build:
--- a/setup.py
+++ b/setup.py
@@ -26,14 +26,13 @@ with io.open('README.md', 'r', encoding='utf-8') as readme_file:
 
 requirements = [
     'binaryornot>=0.4.4',
-    'Jinja2<3.0.0',
+    'Jinja2>=2.7,<4.0.0',
     'click>=7.0',
     'poyo>=0.5.0',
     'jinja2-time>=0.2.0',
     'python-slugify>=4.0.0',
     'requests>=2.23.0',
     'six>=1.10',
-    'MarkupSafe<2.0.0',
 ]
 
 if sys.argv[-1] == 'readme':


SharpEdgeMarshall:
We should backport CI updates too

ssbarnea:
> We should backport CI updates too

Ohh 💩...., that may require an amount of work that nobody will be able to do. The hidden cost of not making any release in more than an year.

@simobasso What can we do to unblock this? The jinja2 and MarkupSafe new releases are causing lots of failures for cookiecutter users. I already found CI/CD on two projects I use and another user made a comment on my fix aking about the backport.

I am wondering if it would not make sense to perform this hotfix manually, including releasing it. It should buy us enough time to make the 2.0.0 release as I doubt we will see many similar problems very soon. It work nothing that these were the only dependencies that were pinned down.

SharpEdgeMarshall:
We could try to backport CI updates, it should not be a big deal.

ssbarnea:
@SharpEdgeMarshall Apparently GHA refuses to start running any workflows on this branch, even after I updated the conditions. It is supposed to run old jobs names like linting, ubuntu-py27,... but it triggers nothing, the only one listed below are those that I marked as required in branch rules but if you look at actions tab, you will not find anything triggered.

ssbarnea:
After manually adding the branch config I was able to trigger the workflow but only on push (after the merge) see https://github.com/cookiecutter/cookiecutter/actions/runs/841530404 

That PR made from my fork towards this the stable branch still does not trigger a thing. I wonder what else I am missing.

ssbarnea:
Found! Apparently putting `pull_request: {}` works. Somehow branches: * bit was preventing it from triggering the workflows... another badly documented area on the workflows.

SharpEdgeMarshall:
Actually is documented [here](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#patterns-to-match-branches-and-tags):
>
'*' | Matches all branch and tag names that don't contain a slash (/). The * character is a special character in YAML. When you start a pattern with *, you must use quotes. | mainreleases
-- | -- | --
'**' | Matches all branch and tag names. This is the default behavior when you don't use a branches or tags filter. | all/the/branchesevery/tag

.github/workflows/main.yml
```diff
@@ -8,8 +8,6 @@ on:
     tags:
       - "*"
   pull_request:
-    branches:
```
SharpEdgeMarshall, line 11: Keep the old version, just try with `**`

ssbarnea:
@SharpEdgeMarshall Do not forget to approve the PR, so I can merge it w/o using admin rights.

PR #1568, opened 2021-05-31
Make release-drafter diff only between master releases

SharpEdgeMarshall:
This should restore release-drafter for the 2.0.0 release

Based on this comment on release-drafter: https://github.com/release-drafter/release-drafter/issues/656#issuecomment-806184887

This PR makes release-drafter considering only releases made on master branch.
If we want to support also hotfixes on stable branches we should make a release-drafter config for each `stable/x` branch

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,3 +1,5 @@
+filter-by-commitish: true
+commitish: master
 categories:
   - title: 'Breaking Changes'
     labels:


ssbarnea (changes requested):
IMHO that PR makes no sense as we already run release drafter only on master, see https://github.com/cookiecutter/cookiecutter/blob/master/.github/workflows/release-drafter.yml#L6

Keep in mind that release drafter must run on any push made to master branch because
it does update the draft release notes (something visible only to project members).

Did I miss something?

SharpEdgeMarshall:
> IMHO that PR makes no sense as we already run release drafter only on master, see https://github.com/cookiecutter/cookiecutter/blob/master/.github/workflows/release-drafter.yml#L6
> 
> Keep in mind that release drafter must run on any push made to master branch because
> it does update the draft release notes (something visible only to project members).
> 
> Did I miss something?

Yes if you take a look art the current release draft, after you did the patch release 1.7.3, we lost all old PR changes.
This PR configure release-drafter to look only at releases made on master branch so we should get back all PR from 1.7.2 to master HEAD

ssbarnea (approved):

PR #1569, opened 2021-06-01
Add support for click 8.x

### LINKED ISSUE #1563: Add support for click 8.0.0
### Description:

We should add support to click 8.0.0, there are some breaking changes like #1558

cjolowicz:
This PR makes `read_user_dict` compatible with click 8.x, and widens the version constraint to include the new release.

Closes #1563 

Based on @davidism's comment in https://github.com/cookiecutter/cookiecutter/issues/1558#issuecomment-852182383

**Additional info**

The `read_user_dict` function uses a `"default"` sentinel instead of the actual default value. Being a JSON dict, the latter would often be hard to read.

Under click 8.x, the default value for `click.prompt` is passed to the `read_proc` callback. We use this callback to load JSON from the user input, and this would choke on an input like `"default"` (without quotes). Therefore, we change the callback to return the actual default value when it receives this input.

Under click 7.x (which remains our minimum version), the default value for `click.prompt` is returned as-is. Therefore, we continue to handle the case where `click.prompt` returns `"default"` instead of the actual default value. (In this last case, we explicitly check for click 7.x. This is only done for clarity, though. Under click 8.x, `click.prompt` would never return `"default"`, even if a user entered it as a valid JSON string. This is because our callback requires a dict, not a string.)

The PR also updates three unit tests in `test_read_user_dict`, which broke due to mocks being tied to implementation details. The commit messages have more detail on this.


ssbarnea (approved):

ssbarnea:
@cjolowicz can you please rebase this? Basically dependencies moved to setup.cfg. That is indeed a change that we really want in next release.

cjolowicz:
@ssbarnea done :)

PR #1566, opened 2021-05-28
context 2.0 - Advanced features

louisdegaste:
This is a continuation of eruber's PR #1008 to implement a new cookiecutter context format with the following features:
* Use of [JSON Schema](https://json-schema.org/) to define the requirements of cookiecutter.json (also works with cookiecutter.yml)
* New Variables structure with:
  * Different types: string, boolean, yes_no, int, float, uuid, json
  * Regex validation: checks if a prompt entry matches a user defined regex expression and asks again on failure
  * Skipping options: if a certain value has been entered, skip to the next or to a specific question
  * Variable description: extended description in verbose mode
  * Hidden input prompt: to enter sensitive information like access tokens
* Versioning for the schema to facilitate future changes
* Requirements in cookiecutter.json (python version, cookiecutter version)
This implementation is compatible with the old cookiecutter context format (we call it schema 1.0). 
It contains a single breaking change: If we have a field named "version" in cookiecutter.json, it will be interpreted as the schema version if it corresponds to an existing schema version number. Because schema 1.0 had no pre-defined structure, it is impossible to avoid this kind of collision, but the issue should be quite rare in practice.
<details>
  <summary>Sample cookiecutter.yml</summary>
  
  ```yaml
  version: '2.0'
  
  requires:
    cookiecutter: '2.0.0'
    python: '>=3.8'
  
  template:
    name: awesome-template
    version: '0.1'
    description: this is the best cookiecutter template ever, you should totally use it
    authors:
      - Jane Doe <jane@example.org>
      - John Doe <jane@example.org>
    license: Apache 2.0
    keywords:
      - awesome
      - python
    url: https://github.com/...
    variables:
      - name: full_name
        default: Jane Doe
        prompt: What's your name?
        description: Please enter your full name. It will be displayed in the README file
          and used for the PyPI package definition.
        type: string
      - name: email
        default: jane@example.org
        prompt: What's your email?
        description: Please enter an email address for the meta information in setup.py.
        type: string
      - name: project_name
        default: my-project
        prompt: What's the project name?
        description: Please enter a name for your project.
        type: string
      - name: module_name
        default: "{{ cookiecutter.project_name | slugify | replace('-','_') }}"
        prompt: Please enter a name for your base python module
        type: string
        validation: "^[_a-zA-Z][_a-zA-Z0-9]+$"
      - name: license
        default: MIT
        prompt: Please choose a license!
        description: Cookiecutter will add an according LICENSE file for you and set the
          according classifier in setup.py.
        type: string
[...]

PR #1571, opened 2021-06-07
context 2.0 - JSON schema

louisdegaste:
This PR is a follow-up to PR #1566. It contains only the features of context V2 related to JSON Schema implementation. 
Because the context V2 enables significantly more complexity in the cookiecutter.json schema, it becomes necessary to have a more robust and unified way of defining the structure of the cookiecutter.json file. This is why we want to use the python package for [JSON schema](https://json-schema.org/) to do so. This PR includes :

- Definition of the V1.0 Schema (pass all schema)
- Base definition of the V2.0 Schema:
- Tools for inferring schema version and validating the structure of a  cookiecutter.json file


louisdegaste:
replacing this PR with a cleaner one

PR #1572, opened 2021-06-07
context 2.0 - classes and base structure

louisdegaste:
This PR is a follow-up to PR #1566. It contains the implementation of the context V2 classes and run-flow. This includes :

- Variable class: a class for implementing the dynamical features of a context variable
- Template class: a class for implementing the dynamical features of a cookiecutter context (contains a list of **Variable** instances)
- load_context(): An iterative method for prompting the variables to handle more complex behavior
- Integration of the V2 run flow into the V1 framework in a non-interfering way.


louisdegaste:
We're closing this PR as we are going to improve the workflow for reviewing #1566

Issue #1479, opened 2020-12-02
Extend cookiecutter templates


isik-kaplan:
* Cookiecutter version: 2.0.0

### Description:

Hello, reading the docs I couldn't find anything relevant, pardon me if it is in the docs and I couldn't find it.

I was wondering if there was any way to orchestrate multiple cookiecutters from yet another cookiecutter. 

```
├── {{ cookiecutter.project_slug }}
│   ├──  {% extend_cookiecutter "cookiecutter-django" %}
│   ├── {% extend_cookiecutter "cookiecutter-data-scientece" %}
```

If it is not possible, consider this a feature request. 

Also, given [this](https://github.com/cookiecutter/cookiecutter/issues/1467) is implemented, this feature would allow even more beautiful cookiecutters. 

 ```
├── {% for cookiecutter.custom_app_count }
│   ├──  {% extend_cookiecutter "cookiecutter-djangopackage" %}
```

Although as I was writing the above I realized that we can't do for loops in cookiecutter's directories (I assume), which yet would be another feature request I'd guess.[1]  But the second is just my hyped imagination while I was writing the first part of this issue.

[1]: Given that this is technically another feature request, do let me know and I'll edit this and move this part to another issue

Sean-McVeigh:
One of the primary attributes of cookiecutter over other scaffolding tools is its conservativeness. A cookiecutter template should do exactly what it's supposed to do and no more. To that end, you wouldn't want to extend both the `django` and `data-science` cookiecutters since there would be a lot of overlap and could possibly conflict with one another. 

For example, suppose one cookiecutter used a Docker container in conjunction with `pipenv` and another had no Docker but used `poetry`? Or different linting tools (flake8, black, isort, etc.). Therefore, the best solution is to begin with a cookiecutter that closely resembles what you want and then fork it.

Now, that said, I suggest you voice your support for #1484 (and its PR) which IMO is the best compromise in enforcing conservativeness and providing extendibility. One could add, extend, or supersede templates from a fork or include templates from other cookiecutters using git submodules.

PR #1469, opened 2020-10-24
ISSUE-1441: add support to git clone recursive submodules

jgrasser:

PR #1048, opened 2018-03-02
Add support for recursing git submodules

jgrasser:

Issue #1041, opened 2018-02-15
Git submodule within template are not cloned from Git repo


szelenka:
* Cookiecutter version: 1.6.0
* Template project url: https://github.com/szelenka/cookiecutter_bug
* Python version: 3.6
* Operating System: macOS 10.12.6

### Description:
Trying to create a cookiecutter with a submodule. When working with the git repo locally, the cookiecutter command can duplicate the structure of the submodule. 

However, when trying to do this through git, it fails to pull the structure of the submodule.

This is likely not performing a clone with the **--recurse-submodules** flag when cloning from Git. I couldn't find an option in the documentation if this is supported or not.

### What I've run:
When performing locally:
```
$ cookiecutter cookiecutter_bug/
project_name [project_name]: 
repo_name [project_name]: 
$ tree project_name/
project_name/
├── cookiecutter_bug_submodule
│   ├── dir1
│   └── dir2
└── internal_dir

4 directories, 0 files
```

Compare to Git:
```
$ cookiecutter https://github.com/szelenka/cookiecutter_bug.git
project_name [project_name]: 
repo_name [project_name]: 
$ tree project_name/
project_name/
├── cookiecutter_bug_submodule
└── internal_dir

2 directories, 0 files
```

I would expect the Git option to clone the submodule and create the structure of the submodule, similar to how it operates when using a local template.

jgrasser:
This code for cloning/checking out a repository exists in the [vcs.py](https://github.com/audreyr/cookiecutter/blob/master/cookiecutter/vcs.py) module.

As you can see, you were correct in your assertion that they were not cloning with the --recurse-submodules flag.

This looks like a fairly simple change.

jgrasser:
After some more review, this changes look  a little more complex since the clone method is built to be usable with both git and mercurial. I'll play around with the code and see if I can't make something a bit cleaner.

jgrasser:
#1048 should do it.

yuvaldolev:
Hi @szelenka @jgrasser @cagonza6

We started working on a new community fork of the Cookiecutter project, named Cookieninja.
The fork is found at: https://github.com/cookieninja-generator/cookieninja

We have re-opened your issue at: https://github.com/cookieninja-generator/cookieninja/issues/48

We're trying to bring the life back to this amazing project.
We are committed to addressing each issue in up to 1 week from the day it's opened.
We'd really like you to join the fork, star it and contribute to it!

Thank you :)

PR #1573, opened 2021-06-08
Added support for git submodules.

cagonza6:
This PR is based in previous work by @jgrasser see: #1041, #1048 and #1469; it has been discussed also in other context as in #1479.
The code was base on those requests which are not of my authory, I recicled what @jgrasser made and make it compatible with the current code base.

It provides:
 * Support for git submodules
 * A bit of documentation on the feature
 * Unit test

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -81,6 +81,11 @@ def list_installed_templates(default_config, passed_config_file):
     '-c', '--checkout', help='branch, tag or commit to checkout after git clone',
 )
 @click.option(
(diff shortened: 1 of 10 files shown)


cagonza6:
Updated PR to consider latest master: d6037b7

fmfreeze:
What is the problem here?
I'd love to see this feature and - fortunately - cookiecutter isn't dead yet, as there is some activity again :)

kurtmckee:
I'm closing this PR due to its age, and to avoid cookiecutter taking on increasing responsibility for interacting with VCS systems.

Issue #1555, opened 2021-05-15
2.0.1 🚀 


simobasso:
Hey @cookiecutter 👋 ,
I think we should start to talks about 2.0.1.

We have more than a year of unreleased contributions with some breaking changes (python 2, pyyaml, etc) so the next release will be a major 🎉  and will be huge!

Let's start with sweet things...

# Changes

## Breaking Changes

* Replace poyo with pyyaml. (#1489) @dHannasch
* Added: Path templates will be rendered when copy_without_render used (#839) @noirbizarre
* Added: End of line detection and configuration. (#1407) @insspb
* Remove support for python2.7 (#1386) @ssbarnea
* allow jinja 3.0.0 (#1548) @wouterdb

## Minor Changes

* Log the error message when git clone fails, not just the return code (#1505) @logworthy
* Added uuid extension to be able to generate uuids (#1493) @jonaswre
* Alert user if choice is invalid (#1496) @dHannasch
* update AUTHOR lead (#1532) @HosamAlmoghraby
* Add Python 3.9 (#1478) @gliptak
* Added: --list-installed cli option, listing already downloaded cookiecutter packages (#1096) @chrisbrake
* Added: Jinja2 Environment extension on files generation stage (#1419) @insspb
* Added: --replay-file cli option, for replay file distributing (#906) @Cadair
* Added: _output_dir to cookiecutter context (#1034) @Casyfill
* Added: CLI option to ignore hooks (#992) @rgreinho
* Changed: Generated projects can use multiple type hooks at same time. (sh + py) (#974) @milonimrod
* Added: Path templates will be rendered when copy_without_render used (#839) @noirbizarre
* Making code python 3 only: Remove python2 u' sign, fix some strings (#1402) @insspb
* py3: remove futures, six and encoding (#1401) @insspb
* Render variables starting with an underscore. (#1339) @smoothml
* Tests refactoring: test_utils write issues fixed #1405 (#1406) @insspb

## CI/CD and QA changes

* Replace tmpdir in favour of tmp_path (#1545) @SharpEdgeMarshall
* Fix linting in CI (#1546) @SharpEdgeMarshall
* Coverage 100% (#1526) @SharpEdgeMarshall
* Run coverage with matrix (#1521) @SharpEdgeMarshall
* Lint rst files (#1443) @ssbarnea
* Python3: Changed io.open to build-in open (PEP3116) (#1408) @insspb
[...]

PR #1577, opened 2021-06-14
Adopt setuptools-scm packaging

ssbarnea:
One direct benefit of this change is that package version is no longer hardcoded and is generated based on last tag.

Related: #1555

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,6 +24,8 @@ jobs:
           pip install tox virtualenv
       - name: Lint
         run: "tox -e lint"
+      - name: Packaging
+        run: "tox -e packaging"
       - name: Safety
         run: "tox -e safety"
   build:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,17 @@
+[build-system]
+requires = [
+  "setuptools >= 42.0.0",  # required by pyproject+setuptools_scm integration
+  "setuptools_scm[toml] >= 3.5.0",  # required for "no-local-version" scheme
+  "setuptools_scm_git_archive >= 1.0",
+  "wheel",
+]
+build-backend = "setuptools.build_meta"
+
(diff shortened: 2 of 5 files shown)


luzfcb:
The only thing I don't like about `setuptools_scm` is that apparently, it doesn't work well with PEP 508

package@https://github.com/repository/project/archive/branch_or_commit_hash.zip#sha1=HASH

I discovered this behavior yesterday so I don't have a solution or workaround

```bash
pip install pip -U
pip django-user-sessions@https://github.com/jazzband/django-user-sessions/archive/master.zip
```

ssbarnea:
It is expected not to work with git archives because it does not have any clue about which version to use. Git clones work. Is not the only thing not working with git archives. If I remember well microsoft was considering adding some extra metadata to these archives in order to enable consumers to make use of tag information.

simobasso (commented):
Hey @ssbarnea 👋 ,

Does this configuration get only the last tag? What about backports?

ssbarnea:
In fact it will work for backports as long they are made from inside branches named stable/*

simobasso (approved):

audreyfeldroy:
Hi @simobasso @ssbarnea @luzfcb it looks like you have done a huge amount of work on attempting to improve packaging of Cookiecutter. The current team that is trying to put out the Cookiecutter 2 release (including myself) is having trouble getting setuptools_scm and build to work properly. We want to use the improved setup that you created here in this pull request, but we just cant get it to work on my computer. Can any of you help us sometime over the next 24 hours over a voice screen share in discord the Cookiecutter [Discord](https://discord.gg/ZvaJt9cAbj)? The ideal situation would be for us to use the current setup as-is to the new release, but if we cant figure it out, we will revert this pull request.

NOTE: This message was typed by @ozer550  as Audrey is unable to use her hands.

Issue #1543, opened 2021-05-01
Improve `--no-input` documentation


glumia:
Check https://github.com/cookiecutter/cookiecutter/issues/1201#issuecomment-830597644.

Currently it isn't clear from the documentation what exactly cookiecutter's behaviour is in case the `--no-input` flag is provided.

We should add the information about template's cache refresh and check if there's something else missing (by looking at what happens when `no_input` is `True` in the source code).

jeremyswerdlow:
Hi there, was looking at starting to contribute and thought this would be a good issue with which to begin. However, I see that in HISTORY.md there is an accreditation to @pokoli for their documentation on this flag.

Is this issue still valid?

glumia:
Hi @jeremyswerdlow, yes it is still valid! 

If you take a look at https://github.com/cookiecutter/cookiecutter/pull/701 you can see that it doesn't explain exactly what happens when you provide the `--no-input` flag on the command line (I remember noticing that template's cache refresh was only one of its undocumented side effects).

jeremyswerdlow:
Great! If it's alright I'll take over this issue then!

jeremyswerdlow:
Looks like the CR needs to be approved for testing by a maintainer, if someone has a chance

insspb:
@jeremyswerdlow @glumia Thank you for contribution, will be in next release.

PR #1587, opened 2021-08-06
Expand cli documentation relating to the no-input flag (#1543)

jeremyswerdlow:
Addresses the concerns in #1543 that not all side effects are clearly documented

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -168,7 +168,7 @@ def render_variable(env, raw, cookiecutter_dict):
 def prompt_choice_for_config(cookiecutter_dict, env, key, options, no_input):
     """Prompt user with a set of options to choose from.
 
-    :no_input: Do not prompt for user input and return the first available option.
+    :param no_input: Do not prompt for user input and return the first available option.
     """
     rendered_options = [render_variable(env, raw, cookiecutter_dict) for raw in options]
     if no_input:
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -168,10 +168,9 @@ def render_variable(env, raw, cookiecutter_dict):
 def prompt_choice_for_config(cookiecutter_dict, env, key, options, no_input):
     """Prompt user with a set of options to choose from.
 
-    Each of the possible choices is rendered beforehand.
+    :no_input: Do not prompt for user input and return the first available option.
     """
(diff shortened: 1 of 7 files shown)


glumia:
Hey @jeremyswerdlow thanks for taking time for this, and sorry for the delay on the review!

While we're at it, I'd like to be able to fix other problems with flag's documentation within the codebase.  

I tracked down all the places where `no_input` appears, and this is what I noticed:

- cookiecutter/main.py::cookiecutter
- cookiecutter/prompt.py::prompt_for_config
- cookiecutter/prompt.py::prompt_choice_for_config
- cookiecutter/repository.py::determine_repo_dir

In all those places the docstring just says *"Prompt the user at command line for
manual configuration?"*. But as a user of this function, I ask myself: what
happens if the end-user isn't prompted for manual configuration? (I
should be able to understand it by looking only at the docstring) 


- cookiecutter/zipfile.py::unzip

*"Suppress any prompts"* doesn't say anything about how the function
behaves in this case. Take a look at cookiecutter/utils.py::prompt_and_delete, 
this is an example of good documentation: if I read the docstring I
don't have to guess what happens.


- cookiecutter/vcs.py::clone

Same as above, how does the function behave if the flag is `True`? (there's no mention
about the cache being refreshed).


- docs/advanced/suppressing_prompts.rst:6 

We should document also here the fact that `no_input` causes a refresh of the cached resources. 
I would also add a comment on line 6 to say that `no_input` corresponds to `--no-input` flag on the CLI. 


Would you mind taking care of them?

glumia:
Hey @jeremyswerdlow are you still interested on working on this? 

I don't want to put pressure, it's just to know if I should wait or add those changes by myself and merge the pr.

cookiecutter/prompt.py
```diff
@@ -168,10 +168,9 @@ def render_variable(env, raw, cookiecutter_dict):
 def prompt_choice_for_config(cookiecutter_dict, env, key, options, no_input):
     """Prompt user with a set of options to choose from.
 
-    Each of the possible choices is rendered beforehand.
+    :no_input: Do not prompt for user input and return the first available option.
```
insspb, line 171:
```suggestion
    :param no_input: Do not prompt for user input and return the first available option.
```

insspb (approved):

insspb:
@jeremyswerdlow @glumia Thank you for contribution, will be in next release.

PR #1590, opened 2021-08-20
Sourcehut abbreviation [WIP]

jgarte:
Hi, Sourcehut has mercurial repositories also. 

This pr does not support those.

Any ideas how to cover that in a simple way?

It would be cool if the user could just give the `sr` abbreviation and either a git or hg repo as arg and cookiecutter will just dispatch to either `git.sr.ht/~` or `hg.sr.ht/~`


audreyfeldroy (approved):
I reviewed this and it looks great.

ssbarnea:
Closing as I do not think that this service is popular enough to deserve an abbreviation. The fact that we did not see other supporters in so long is a proof. If situation changes, obviously that we will reconsider it.

PR #1609, opened 2021-10-31
Correct broken links to jinja docs for `if` and `for` statements

lmmentel:
The current links for jinja statements `if` and `for` are resolving to `Page not found`, updated them to correct pages.


insspb:
@lmmentel thank you for contribution, unfortunately it is on outdated documentation base. I will replace it in #1691

PR #1608, opened 2021-10-30
Release preparation for 2.0.1rc1

audreyfeldroy:
This PR includes all changes by @audreyfeldroy and @pydanny in preparation for releasing 2.0.1 Release Candidate 1.

PR #1610, opened 2021-11-04
Improve local development step 3

ozer550:
- Use developer install of package
- Don't invoke `python setup.py` directly

See https://stackoverflow.com/questions/19048732/python-setup-py-develop-vs-install
for info

Co-authored with @audreyfeldroy


luzfcb (approved):
LGTM

michaeljoseph:
👌 #1608 fixes the build, i think ✅

michaeljoseph (approved):

ssbarnea (approved):

PR #1621, opened 2021-12-12
Fix Python version number in `cookiecutter --version` and test on Python 3.10

### LINKED ISSUE #1619: `cookiecutter --version` prints out the wrong Python version on 3.10
``` console
$ cookiecutter --version
Cookiecutter 1.7.3 from /usr/lib/python3.10/site-packages (Python 3.1)
```

It should end with `(Python 3.10)` not `(Python 3.1)`. This seems to be the problematic code:

https://github.com/cookiecutter/cookiecutter/blob/f821895a0cb536a907ac2937471d7d4ac36fec75/cookiecutter/cli.py#L26-L31

ozer550:
Closes #1619

- On Python 3.10 `cookiecutter --version` now prints Python 3.10 instead of Python 3.1
- Added test to verify this behavior 
- Added cross-platform Python 3.10 CI environments

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -25,7 +25,7 @@ from cookiecutter.config import get_user_config
 
 def version_msg():
     """Return the Cookiecutter version, location and Python powering it."""
-    python_version = str(sys.version_info[0]) + '.' + str(sys.version_info[1])
+    python_version = f'{sys.version_info.major}.{sys.version_info.minor}'
     location = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
     message = 'Cookiecutter %(version)s from {} (Python {})'
     return message.format(location, python_version)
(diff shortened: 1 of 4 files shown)


audreyfeldroy (approved):
We reviewed this together over screenshare and I think this looks good, but I welcome others to review this.

gotmax23:
I thought f-strings were not supported on Python 2.7 and 3.5, which this project purports it supports. Am I missing something?

michaeljoseph:
@gotmax23 where did you find references to those legacy python versions?
We don't support them any longer (as you can see from the CI runs).

cookiecutter/cli.py
```diff
@@ -25,7 +25,7 @@
 
 def version_msg():
     """Return the Cookiecutter version, location and Python powering it."""
-    python_version = sys.version[:3]
+    python_version = f'{sys.version_info.major}.{sys.version_info.minor}'
```
michaeljoseph, line 28: Any reason we wouldn't want to include the full version here?

gotmax23:
> @gotmax23 where did you find references to those legacy python versions? We don't support them any longer (as you can see from the CI runs).

It's listed in one of the shields.io badges in the README. I guess that whatever data source it pulls from is outdated.

michaeljoseph (approved):
🙌

PR #1634, opened 2022-01-04
Publish with poetry

HosamAlmoghraby:
In order to test the poetry configuration, we need to publish the package to the test.pypi.org repository.
I can push the package to the test repository if you add my test.pypi.org username "hosam" https://test.pypi.org/user/hosam/ to the contributors' list.


HosamAlmoghraby:
closed as there is another pull request #1638

PR #1639, opened 2022-01-10
Docs: Advanced Usage "New in version" unified

alkatar21:
So far, the versions are written differently and sometimes twice. For a new feature they were missing completely.
This is to be made more consistent now.


jensens:
Resolved now in the opposite direction (like it is in Python documentation) within #1679

Issue #1650, opened 2022-03-02
override doesn't work if you have a copy only directory


zhongdai:
* Cookiecutter version: 1.7.3
* Template project url: https://github.com/zhongdai/issue-cookiecutter.git
* Python version: 3.7
* Operating System: Darwin Kernel Version 21.3.0

### Description:

There is an option for the `cookiecutter` command, which is called [override folder](https://cookiecutter.readthedocs.io/en/1.7.2/advanced/cli_options.html#cmdoption-f). However, it does not work if you have defined a 'copy only directory' in your `cookiecutter.json`. I checked the code, looks like a check needs to be done before the copy in the `generate.py`. 

I am happy to fix that, and I got a running version in my local, just need to get familiar with the testing requirements and contributing rules.

### What I've run:

```
~/projects/temp
❯ cookiecutter --version
Cookiecutter 1.7.3 from /Users/zhongdai/.local/pipx/venvs/cookiecutter/lib/python3.7/site-packages (Python 3.7)

~/projects/temp
❯ uname -a
Darwin Zhongs-iMac.local 21.3.0 Darwin Kernel Version 21.3.0: Wed Jan  5 21:37:58 PST 2022; root:xnu-8019.80.24~20/RELEASE_X86_64 x86_64

~/projects/temp
❯ cookiecutter -v https://github.com/zhongdai/issue-cookiecutter.git
DEBUG cookiecutter.config: Loading config from /Users/zhongdai/.cookiecutterrc.
DEBUG cookiecutter.config: config_path is /Users/zhongdai/.cookiecutterrc
DEBUG cookiecutter.utils: Making sure path exists: /Users/zhongdai/.cookiecutters/
DEBUG cookiecutter.vcs: repo_dir is /Users/zhongdai/.cookiecutters/issue-cookiecutter
DEBUG cookiecutter.main: context_file is /Users/zhongdai/.cookiecutters/issue-cookiecutter/cookiecutter.json
DEBUG cookiecutter.generate: Context generated is OrderedDict([('cookiecutter', OrderedDict([('project_name', 'reproduce-issue-project'), ('_copy_without_render', ['copy_only_dir'])]))])
project_name [reproduce-issue-project]:
DEBUG cookiecutter.utils: Making sure path exists: /Users/zhongdai/.cookiecutter_replay/
DEBUG cookiecutter.find: Searching /Users/zhongdai/.cookiecutters/issue-cookiecutter for the project template.
[...]

PR #1651, opened 2022-03-02
fixed the override not working with copy only dir #1650

zhongdai:
Fixed the issue of failure execution when
- the `overwrite_if_exists=True`
- there are copy only directories in the context

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -333,6 +333,7 @@ def generate_files(
                 # specified in the ``_copy_without_render`` setting, but
                 # we store just the dir name
                 if is_copy_only_path(d_, context):
+                    logger.debug('Found copy only path %s', d)
                     copy_dirs.append(d)
                 else:
                     render_dirs.append(d)
@@ -342,6 +343,10 @@ def generate_files(
                 outdir = os.path.normpath(os.path.join(project_dir, indir))
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
+
+                # if the outdir is there, it must be a overwrite execution
+                if os.path.isdir(outdir):
+                    shutil.rmtree(outdir)
                 shutil.copytree(indir, outdir)
 
             # We mutate ``dirs``, because we only want to go through these dirs
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -346,8 +346,10 @@ def generate_files(
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
 
-                # if the outdir is there, it must be a overwrite execution
-                if os.path.isdir(outdir) and overwrite_if_exists:
+                # The outdir is not the root dir, it is the dir which marked as copy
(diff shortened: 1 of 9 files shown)


zhongdai:
This bug is really annoying when you create your own cookiecutter package and need to generate the files again and again. You have to manual `rm -rf the-dir` before you can re-generate it. Is there anyone who can have a look?

cookiecutter/generate.py
```diff
@@ -339,6 +340,10 @@ def generate_files(
                 outdir = os.path.normpath(os.path.join(project_dir, indir))
                 outdir = env.from_string(outdir).render(**context)
                 logger.debug('Copying dir %s to %s without rendering', indir, outdir)
+
+                # if the outdir is there, it must be a overwrite execution
+                if os.path.isdir(outdir):
+                    shutil.rmtree(outdir)
```
att14, line 346: I don't think this is the correct solution. I think #1661 is.

att14: As discussed in #1661. I think this is the correct solution for all versions of python.

```suggestion
                # if the outdir is there, it must be a overwrite execution
                if os.path.isdir(outdir) and overwrite_if_exists:
                    shutil.rmtree(outdir)
```

att14:
Here was my attempt to fix the failing test: https://github.com/cookiecutter/cookiecutter/pull/1661/commits/19a8f2fce6f50810dd558f34d32b085f46a67a55.

It might not work on Windows, though.

zhongdai:
@att14 

Let step back to think about the requirement of this `-f` flag. I had a look for the document, and I noticed [doc](https://cookiecutter.readthedocs.io/en/1.7.2/advanced/cli_options.html#cmdoption-f). It says `Overwrite the contents of the output directory if it already exists`. 

Here is my understanding, if you decide to use `-f` flag, you give up any changes to the existing clone, and you re-start a new clone. So my initial fix was straightforward, deleting the existing clone.

The `-f` is not a frequently used option, because most of the time, you clone once, and you start to modify the clone. However, this flag is useful for template developers, they need to clone the template again and again to check the output content. During the testing/verification, they don't want to pick different project names, so the `-f` flag is very handy to let them use the same project name.


I am happy to hear your thought or anyone else from the committer group.

att14:
I don't disagree with you. I'm confused what the problem is. The change I made is to make your delete respect the `-f` option. You want to avoid deleting all the copy only directories if they haven't supplied `-f`.

zhongdai:
I don't think there is a case the copy-on-dir exists and the `overwrite_if_exists=False`. Please check this [block of code](https://github.com/cookiecutter/cookiecutter/blob/master/cookiecutter/generate.py#L216), if the cloned dir exists and the `overwrite_if_exists=False`, the program will raise an exception, but not copying/rendering anything. This means, if the code hits this line `if os.path.isdir(outdir) and overwrite_if_exists:`, the `overwrite_if_exists` must be `True`, and we don't need to add that check. 

I also noticed there is any flag `skip_if_file_exists`, and I believe that is the one to control not overriding any modified content.

att14:
You are correct, but I see no reason to avoid being defensive since the check is in another function.

att14 (approved):

insspb (approved):

Issue #1655, opened 2022-03-28
test_generate_file_verbose_template_syntax_error fails with ./ difference in path 


sebix:
* Cookiecutter version: 2.0.2
* Template project url: none
* Python version: 3.9
* Operating System: openSUSE Tumbleweed

### Description:

Running the tests in the openSUSE build environment yields an error in the test function `test_generate_file_verbose_template_syntax_error` which is just a mismatch between `tests/files/syntax_error.txt` and `./tests/files/syntax_error.txt`.

### What I've run:

```
pytest-3.9 -v tests
[...]
[   10s] =================================== FAILURES ===================================
[   10s] _______________ test_generate_file_verbose_template_syntax_error _______________
[   10s] 
[   10s] env = <cookiecutter.environment.StrictEnvironment object at 0x7fa6e5baf040>
[   10s] expected_msg = 'Missing end of comment tag\n  File "./tests/files/syntax_error.txt", line 1\n    I eat {{ syntax_error }} {# this comment is not closed}'
[   10s] 
[   10s]     def test_generate_file_verbose_template_syntax_error(env, expected_msg):
[   10s]         """Verify correct exception raised on syntax error in file before generation."""
[   10s]         with pytest.raises(TemplateSyntaxError) as exception:
[   10s]             generate.generate_file(
[   10s]                 project_dir=".",
[   10s]                 infile='tests/files/syntax_error.txt',
[   10s]                 context={'syntax_error': 'syntax_error'},
[   10s]                 env=env,
[   10s]             )
[   10s] >       assert str(exception.value) == expected_msg
[   10s] E       assert 'Missing end ...s not closed}' == 'Missing end ...s not closed}'
[   10s] E           Missing end of comment tag
[   10s] E         -   File "./tests/files/syntax_error.txt", line 1
[   10s] E         ?         --
[   10s] E         +   File "tests/files/syntax_error.txt", line 1
[   10s] E               I eat {{ syntax_error }} {# this comment is not closed}
[   10s] 
[   10s] tests/test_generate_file.py:136: AssertionError
[   10s] ------------------------------ Captured log call -------------------------------
[   10s] DEBUG    cookiecutter.generate:generate.py:144 Processing file tests/files/syntax_error.txt
[   10s] DEBUG    cookiecutter.generate:generate.py:159 Created file at ./tests/files/syntax_error.txt
[...]

PR #1629, opened 2021-12-27
Removed changes related to setuptools_scm

ozer550:
As part of our group programming session today led by @audreyfeldroy, we are reverting #1577 together.

PR #1656, opened 2022-03-28
setup: fix version number

sebix:
minor change to fix the version number in setup.py. that mistake leads
to wrong version numbers on the local installation

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -2,7 +2,7 @@
 """cookiecutter distutils configuration."""
 from setuptools import setup
 
-version = "2.0.0"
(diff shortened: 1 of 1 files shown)


sebix:
The test fails are due to #1655

bnavigator:
The root cause is #1629 which reverted a bit more than it should have. Including the click <8 upper pin.

ericof:
This should be solved with [release 2.1.0](https://github.com/cookiecutter/cookiecutter/releases/tag/2.1.0)

PR #1679, opened 2022-05-30
housekeeping: run pyupgrade to find remaining py27 artifacts

jensens:
Looked for Python 2.7 artifacts by running ``find . -name "*.py" -exec pyupgrade  --py3-only --py37-plus {} +``.

Primary changes:
- remove inheritence from ``object``
- a bunch conversion to f-strings
- super calls w/o parameters
- open without ``'r'``or ``rt``  (its by default)
- IOError -> OSError
- binary with `b'...'` instead of `'...'.encode('utf8')`


ericof (approved):
LGTM

Issue #1695, opened 2022-06-06
Please, inform which are the valid values in docs advanced/boolean_variables.html


italomaia:
### Description:

Could you please inform which are the valid values for a boolean configuration variable? Will "y" coerce to True? What about "yes"? What about "n" and "no"? What about "true"?

insspb:
Hi @italomaia, are you asking about this feature?
https://cookiecutter.readthedocs.io/en/latest/advanced/boolean_variables.html

In json specification it is only ``true`` or ``false``. this feature is not released to pip (2.2+)

alkatar21:
I think the question is not what you can specify in json, but what users can specify on the command line for true/false during generation. The question I have also asked myself, what all works there?

italomaia:
@insspb I refer to the user input. Documentation informs how to create an boolean configuration but not what counts as a true or false user input. Is there any pre-procession? Does the user have to write specifically True or False when prompted?

italomaia:
@luzfcb which is reasonable. We need clarity regarding how cookiecutter does it. Can the ticket be rebranded, as it is **not a question**?

luzfcb:
@italomaia I guess it is only necessary to update the documentation.
In theory, these are the supported values: 

https://github.com/cookiecutter/cookiecutter/blob/5ea952204024180c2fca383e0410b3451f603ead/cookiecutter/prompt.py#L23-L33

italomaia:
Indeed. MR available: https://github.com/cookiecutter/cookiecutter/pull/1705

insspb:
@italomaia @italomaia Thank you for contribution and asking this question. Updated docstrings and documentation. Check as 'latest' hope now everything clear.

PR #1705, opened 2022-06-06
Update boolean_variables.rst

italomaia:
Adding explanation on what is considered valid input for boolean configuration.


insspb:
Closes #1695

Issue #1403, opened 2020-05-26
Windows compatibility. Change all os.path invocations to pathlib.Path.


insspb:
* Cookiecutter version:
* Template project url:
* Python version:
* Operating System:

### Description:

// REPLACE ME: What are you trying to get done, what has happened, what went wrong, and what did you expect?

### What I've run:

```
// REPLACE ME: Paste a log of command(s) you ran and cookiecutter's output, tracebacks, etc, here
```

SharpEdgeMarshall:
I would like to take care of this issue, could you explain why `os.path` should be replaced for compatibility reasons on Windows?

luzfcb:
> I would like to take care of this issue, could you explain why os.path should be replaced for compatibility reasons on Windows?

The `Path` class from pathlib normalizes path separators automatically.

On Linux:

```python
>>> from pathlib import Path
>>> str(Path('src/foobar'))
'src/foobar'
>>> 
```

On Windows

```python
>>> from pathlib import Path
>>> str(Path('src/foobar'))
'src\\foobar'
>>> 
```

https://treyhunner.com/2019/01/no-really-pathlib-is-great/
https://treyhunner.com/2018/12/why-you-should-be-using-pathlib/

SharpEdgeMarshall:
Ok let's do this

maspling:
I get liking pathlib, but is it really a compatibility issue?

Standard os.path lib:

On Linux:
```python
Python 3.6.9 (default, Jan 26 2021, 15:33:00)
[GCC 8.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from os import path
>>> path.join('src', 'foobar')
'src/foobar'
>>>
```

On Windows:
```python
Python 3.9.10 (tags/v3.9.10:f2f3f53, Jan 17 2022, 15:14:21) [MSC v.1929 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from os import path
>>> path.join('src', 'foobar')
'src\\foobar'
>>>
```

kurtmckee:
`pathlib.Path` doesn't increase Windows compatibility.

The `pathlib` module makes many interactions easier, and its cross-platform path normalization is nice when writing logs or printing files. However, Python on Windows is just fine with paths like `"c:/Users/Kurt/path/whatever"`.

I'm closing this issue.

kurtmckee:
@pydanny Please unpin this issue.

PR #1721, opened 2022-06-08
Code quality: find.py refactored and type annotated

insspb:
Related #1403 

PR #1547 converting os to pathlib silently and implementing a lot of str method calls. It is not what we want, we want to work with Path when it is Path and work with str when str required. 

find module not related to actual work, but require update in test and in module in one time, otherwise it has no sense.


jensens (approved):
LGTM

PR #1722, opened 2022-06-09
Code quality: utils.make_sure_path_exists refactored and type annotated

insspb:
Related #1403

I checked usage of ``make_sure_path_exists`` function, as one of the function related to files modification. First thought was to remove it, as it is almost wrapper over basic functions. But I decided to leave it with correct exception text, but without return.

Type annotate for functions that use ``make_sure_path_exists`` for correct values passing.

### DIFF
--- a/cookiecutter/utils.py
+++ b/cookiecutter/utils.py
@@ -41,7 +41,7 @@ def make_sure_path_exists(path: "os.PathLike[str]") -> None:
     try:
         Path(path).mkdir(parents=True, exist_ok=True)
     except OSError as error:
-        raise OSError(f'Unable to create replay directory at {path}') from error
+        raise OSError(f'Unable to create directory at {path}') from error
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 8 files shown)


cookiecutter/utils.py
```diff
@@ -31,19 +32,16 @@ def rmtree(path):
...
 
-def make_sure_path_exists(path):
+def make_sure_path_exists(path: "os.PathLike[str]") -> None:
     """Ensure that a directory exists.
 
-    :param path: A directory path.
+    :param path: A directory tree path for creation.
     """
-    logger.debug('Making sure path exists: %s', path)
+    logger.debug('Making sure path exists (Create tree if not exist): %s', path)
```
jensens, line 40:
```suggestion
    logger.debug('Making sure path exists (creates tree if not exist): %s', path)
```

jensens (dismissed):
LGTM, just a minor idea how to phrase the log better.

PR #1726, opened 2022-06-09
Enabled black to force string formatting: core files

insspb:
Related #1724

Please force merge it with default **merge commit**, to check changes from #1723 behavior.

PR #1725, opened 2022-06-09
Enabled black to force string formatting: tests files

insspb:
Related #1724

Please merge it with squash and push, to check changes from #1723 behavior.

Issue #1723, opened 2022-06-09
Use "squash and merge" for merging PRs


sisp:
Just a simple suggestion towards a cleaner Git history:

I think many projects these days have adopted the [squash-merging method](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests) for merging PRs. Typically, a PR consists of a single logical change of the target branch. This means, when a PR is merged, the Git history of the target branch receives one additional commit, even though the PR might have several commits due to code review and/or iterative development. Currently, it seems PRs are merged with an additional merge commit which clutters the Git history and makes it harder to read. If squash-merging was used, the Git history would become cleaner.

What do you think?

jensens:
After the latest commits and digging through the history I had similar thoughts. 
I switched the settings from merge commits to squash commits. 
This may lead to smaller PRs, which is favorable.

PR #1727, opened 2022-06-09
Enabled black to force string formatting: final configuration

### LINKED ISSUE #1724: Enable black force string formatting to exclude future discussion on strings
Two year ago I rolled back changes of black for string normalization (#1298). The purpose of revert was to downgrade used python version and for easy PR merges (as PRs based on versions with different strings). 

At this time the time is come to implement such changes, because in several future refactoring we will anyway touch strings and if we implement black after refactoring it will be harder to find real last changes in git blame.

insspb:
This PR should be merged after #1725 and #1726, before them pipeline will be failed.

Closes #1724

Please merge it with squash and push, to check changes from #1723 behavior.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -15,7 +15,7 @@ repos:
     hooks:
       - id: black
         language_version: python3
-        exclude: ^(tests\/hooks-abort-render\/hooks|docs\/HelloCookieCutter1)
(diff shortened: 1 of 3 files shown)


kurtmckee (commented):
@jensens @pydanny @ericof Please close this PR.

This is likely to cause lint failures for other PRs currently in the pipeline.

Issue #1765, opened 2022-08-15
Feature request - print directory tree


liortct:
* Cookiecutter version: latest
* Template project url: https://github.com/liortct/SimplePythonApp.git
* Python version: 3.8
* Operating System: linux

### Description:
After generating a project via cookiecutter there is no indication for the user which files/directories were created.
I want the the directory tree to be printed to the console afterwards.

supcik:
Hello I would like to work on this issue. Is it OK for you ?

PR #1766, opened 2022-08-15
Added: print directory tree

liortct:
Addresses #1765 .

Added print directory tree after generating project.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,7 @@ requirements = [
     'jinja2-time>=0.2.0',
     'python-slugify>=4.0.0',
     'requests>=2.23.0',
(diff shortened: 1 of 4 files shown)


kurtmckee (commented):
@jensens @ericof @browniebroke Please close this PR.

setup.py
```diff
@@ -14,6 +14,7 @@
     'jinja2-time>=0.2.0',
     'python-slugify>=4.0.0',
     'requests>=2.23.0',
+    'directory-tree >= 0.0.2',
```
jensens, line 17: Adding an external dependency is not accepted.

Issue #1744, opened 2022-06-23
Dynamic creation of choice variable


alexandersm:
### Description:

I have the following variables in **cookiecutter.json** file
```
  "_site_locations": {
    "zone1": [
      "SiteA",
      "SiteB",
      "SiteC"
    ],
    "zone2": [
      "SiteD",
      "SiteE"
    ]
  },

  "zone": ["zone1", "zone2"],
  "site" : [ "{{ cookiecutter._site_locations.get(cookiecutter.zone) }}" ],


```

My idea was to show the list of choices for **site** which is dynamically generated according to the selection made in **zone**

But in fact I'm getting the following results:
```
Select zone:
1 - zone1
2 - zone2
Choose from 1, 2 [1]: 1
Select site:
1 - ['SiteA', 'SiteB', 'SiteC']
Choose from 1 [1]:
```
Is there a way to fix that and show 3 choice items in **site** ?

liortct:
I fix this issue in PR #1768.
Now the prompt will locks like that:

![image](https://user-images.githubusercontent.com/38353773/187023534-ecfd6243-3a15-465a-81d5-41e1fe2da24b.png)
But, notice that the `cookiecutter.json` should looks like that:
```json
{
 "_site_locations": {
    "zone1": [
      "SiteA",
      "SiteB",
      "SiteC"
    ],
    "zone2": [
      "SiteD",
      "SiteE"
    ]
  },

  "zone": ["zone1", "zone2"],
  "site" : "{{ cookiecutter._site_locations.get(cookiecutter.zone) }}"
}
```

I advice to close this issue after the PR will merge.

GreasyAvocado:
Anyone knows when this will be merged?

liortct:
@GreasyAvocado , @alexandersm 
Hi!
We started working on a new community fork of the Cookiecutter project, named Cookieninja.
The fork is found at: https://github.com/cookieninja-generator/cookieninja

We have re-opened your issue at https://github.com/cookieninja-generator/cookieninja/issues/22

We're trying to bring the life back to this amazing project.
We are committed to addressing each issue in up to 1 week from the day it's opened.
We'd really like you to join the fork, star it and contribute to it!

Thank you :)

liortct:
@ericof , I solve this issue in #1768 . Lets work on this PR

PR #1768, opened 2022-08-27
Dynamic choice variable

liortct:
Addresses #1744 .

Fixed bug where dynamic non string variable rendered as string.
That's not allowed dynamic creation of choice variable.
Also, added a unit-test for recreate this scenario.

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -2,6 +2,7 @@
 import functools
 import json
 from collections import OrderedDict
+import ast
 
 import click
 from jinja2.exceptions import UndefinedError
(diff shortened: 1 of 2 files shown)


liortct:
Hi!
We started working on a new community fork of the Cookiecutter project, named Cookieninja.
The fork is found at: https://github.com/cookieninja-generator/cookieninja

We have re-opened and merged your PR in the Cookieninja project at https://github.com/cookieninja-generator/cookieninja/pull/23

We're trying to bring the life back to this amazing project.
We are committed to reviewing and merging each PR in up to 2 weeks from the day it's opened (if there are no issues with it).
We'd really like you to join the fork, star it and contribute to it!

Thank you :)

cookiecutter/prompt.py
```diff
@@ -159,7 +160,18 @@ def render_variable(env, raw, cookiecutter_dict):
 
     template = env.from_string(raw)
 
-    return template.render(cookiecutter=cookiecutter_dict)
+    rendered_variable_str = template.render(cookiecutter=cookiecutter_dict)
+    try:
+        rendered_variable = ast.literal_eval(rendered_variable_str)
```
ericof, line 165: @jensens Should we worry about this?

jensens: I am not a fan of even `literal_eval` in code which is meant to run on cookiecutters fetched from Git.
 
see https://docs.python.org/3/library/ast.html#ast.literal_eval

> This function had been documented as “safe” in the past without defining what that meant. That was misleading. This is specifically designed not to execute Python code, unlike the more general [eval()](https://docs.python.org/3/library/functions.html#eval). There is no namespace, no name lookups, or ability to call out. But it is not free from attack: A relatively small input can lead to memory exhaustion or to C stack exhaustion, crashing the process. There is also the possibility for excessive CPU consumption denial of service on some inputs. Calling it on untrusted data is thus not recommended.

liortct:
@ericof . I can work on this PR again.
Do you want I will fix the conflicts?

PR #1773, opened 2022-09-09
fix so dict val parsing isn't deferred

aryeh:
Currently, you cannot reference another value within the cookiecutter.json file if the referenced value is a dict, because those values are deferred to a second pass for parsing.  The cookiecutter.json is parsed as an OrderedDict, therefore there is no benefit to do multiple passes as it only prevents the use of some vars (dicts) in later values.  If a val depends on another value within the json file, it only needs to be put earlier/higher within the file and then can be used anywhere later in the file.

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -208,6 +208,14 @@ def prompt_for_config(context, no_input=False):
                     )
                 else:
                     cookiecutter_dict[key] = read_user_yes_no(key, raw)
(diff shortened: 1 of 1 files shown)


kurtmckee (changes requested):
@aryeh It appears that this change has caused the test suite to fail.

Please rebase on `main` (or merge `main` into this branch), verify that the test suite passes, and add tests that demonstrate that this feature is working. Thanks!

PR #1775, opened 2022-09-13
Add verbose values option to choice options

aminabedi-bentley:
Requirement:
I need to have verbose human-readable values options being shown to the user while having different values being set to the variable underneath. Example question for FooBarFlags variable:
Choices shown: 1-Foo 2- Bar 3- Both Foo and Bar
Actual values: 1- Foo 2- Bar 3-FooBar

Usage:
{
"FooBarFlag": ["Foo", "Bar", ("Both Foo and Bar", "FooBar")] }

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -59,7 +59,8 @@ def read_user_choice(var_name, options):
     if not options:
         raise ValueError
 
-    choice_map = OrderedDict((f'{i}', value) for i, value in enumerate(options, 1))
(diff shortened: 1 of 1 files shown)


kurtmckee (changes requested):
@aminabedi-bentley This seems like it could be a useful feature!

I'm not a maintainer but want to suggest these changes to make this ready for review:

* Add tests that demonstrate that this feature is working
* Add documentation so others can take advantage of this feature

Then, mark this PR as ready for review. Thanks!

henryiii:
This was added in https://github.com/cookiecutter/cookiecutter/pull/1881!

PR #1783, opened 2022-09-26
add req rich for better prompt

copdips:
Hello,

[rich](https://github.com/Textualize/rich) is a very beautiful and popular Python text formatting in the terminal, and as cookiecutter displays text in terminal, it might be useful to add it in the requirements, this makes users no more need to install `rick` package after the installation of `cookiecutter`, if the authors of the template want to use `rich text` in the **pre and post hook scripts**.

If we dont add `rick` in the requirements, the authors must guide users to install it seperately, and sometimes it might be not easy as expected.

And as we just add `rich` in the requirements, this PR won't force people to use it, but ease people who want to use it.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,6 +14,7 @@ requirements = [
     'jinja2-time>=0.2.0',
     'python-slugify>=4.0.0',
     'requests>=2.23.0',
+    'rich>=12.5.0 ; python_version>="3.6"',
 ]
 
 setup(


copdips:
hello,
someone @audreyfeldroy, @browniebroke, @ericof, @hackebrot, @insspb, @pydanny, @simobasso, @ssbarnea could give a feedback to this request please ?

pydanny:
While I'm a big fan of rich, I don't think this belongs in Cookiecutter. Certainly not in this capacity of supporting a subset of users. Adding Rich add [5 other dependencies](https://github.com/Textualize/rich/blob/master/pyproject.toml#L31-L35) on Cookiecutter. We've always tried to keep the dependency tree as light as possible.

Also if we add Rich as a utility dependency, where do we draw the line at adding these kinds of dependencies that aren't affecting code? 

Unless convinced otherwise, I'm 👎🏻 on this PR.

copdips:
> While I'm a big fan of rich, I don't think this belongs in Cookiecutter. Certainly not in this capacity of supporting a subset of users. Adding Rich add [5 other dependencies](https://github.com/Textualize/rich/blob/master/pyproject.toml#L31-L35) on Cookiecutter. We've always tried to keep the dependency tree as light as possible.
> 
> Also if we add Rich as a utility dependency, where do we draw the line at adding these kinds of dependencies that aren't affecting code?
> 
> I'm open to arguments to change my mind, but right now I'm 👎🏻 on this PR.

Hi @pydanny 

Thanks for sharing your thoughts, in fact I totally agree with you about not adding third-party non-important dependencies.
My goal was to add colorful prompt during the pre and post hooks, because most of the user interactions are from there, but unfortunately Python doesn't provide such ability natively, and I wouldn't like to write such things inside of the hooks neither.

I looked around the solutions implemented elsewhere, what about the one used by poetry's script [install-poetry.py](https://github.com/python-poetry/poetry/blob/8a500f053b756616be99454939bcbf7ad42c7e3b/install-poetry.py#L122) ?, they wrote their own color functions, pure code, no pip install.

willmcgugan:
There's 4 dependancies for Rich (one is optional). 2 of which are backports of stdlib modules for older Pythons. The other two are Pygments and Commonmark, both pure-python and mature.

Beyond that I couldn't really say. I understand the desire to keep dependancies light, and I can see @pydanny 's POV that you probably don't want to add all modules that template authors *may* want.

Perhaps cookie cutter templates could expose their own dependancies?

ssbarnea:
I am +0.5 in favour of rich. Still, I would more likely want to see setup.py being ditches and PEP-517 implemented as more important.

jensens:
As @pydanny opposed here, I close the PR.

pydanny:
I don't remember opposing this. Did I? 🤔

PR #1785, opened 2022-10-03
Update documentation to mention that output directory is configurable

WilliamDEdwards:
The documentation says that the output directory is the current directory. However, with PRs https://github.com/cookiecutter/cookiecutter/pull/452 and https://github.com/cookiecutter/cookiecutter/pull/452, it is configurable.

With this commit, the documentation simply links to CLI options page, not to the [CLI option entry on it](https://cookiecutter.readthedocs.io/en/stable/cli_options.html#cmdoption-cookiecutter-o). How do I do that with `:doc`?

### DIFF
--- a/docs/overview.rst
+++ b/docs/overview.rst
@@ -38,7 +38,8 @@ of this.
 Output
 ------
 
-This is what will be generated locally, in your current directory::
+This is what will be generated locally, in your output directory. The output directory defaults to your current directory.
+The location is configurable: see :doc:`cli_options` for details.
 
     mysomething/  <---------- Value corresponding to what you enter at the
     │                         project_name prompt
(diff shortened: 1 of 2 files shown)


WilliamDEdwards:
I'm fixing the failing workflow when the comment in the PR description has been processed. I'm awaiting feedback.

jensens:
The command line options block is generated in code (Sphinx plugin `sphinx_click.ext`). I am not sure if the [generated option](https://github.com/click-contrib/sphinx-click/blob/09a71dcc5052b9e8882eaa606b06826b950ee05b/sphinx_click/ext.py#L150) can be referenced. But if an option can be referenced, then this should work for an generated one too, because in the end both are parsed the same.

Issue #1792, opened 2022-10-30
Working with line-ends: Inverted attributes


Ikari33:
* Cookiecutter version: 2.1.1
* Template project url:
* Python version: 3.9.13
* Operating System: Windows 10

### Description:

In the documentation (link below) there is a code sample with "\n\r", but I think it is the opposite :)
For me it works with "\r\n". 
\r represents cr
\n represents lf

[docs/advanced/new_line_characters.rst](https://github.com/cookiecutter/cookiecutter/blob/2.1.1/docs/advanced/new_line_characters.rst)

Lahiry:
I guess they did a mistake in the docs, indeed the correct order is "\r\n".

"The traditional order, when both control characters are used, is Carriage Return, then Line Feed.

The reason for this goes back to the old ASR-33 Teletype.

When a Carriage Return is issued to an ASR-33, the print head, if it is near the right margin, takes over a tenth of a second to return to the left margin, plus there is a bit of "bounce" when the left margin is hit.

If the order were Line Feed, then Carriage Return, the first printed character might occur a tenth of a second after the Carriage Return, and thus might end up printing (as a smear) halfway across the page. But if Line Feed comes after Carriage Return then the time taken by the Line Feed provides extra time for the print head to complete it's trip."

[StackOverflow](https://stackoverflow.com/questions/1885900/order-of-carriage-return-and-new-line-feed)

I'm going to make a PR to fix this. Please assign this issue to me! :)

yuvaldolev:
Hi @Ikari33 & @Lahiry
We started working on a new community fork of the Cookiecutter project, named Cookieninja.
The fork is found at: https://github.com/cookieninja-generator/cookieninja

We have re-opened your issue at: https://github.com/cookieninja-generator/cookieninja/issues/28

We're trying to bring the life back to this amazing project.
We are committed to addressing each issue in up to 1 week from the day it's opened.
We'd really like you to join the fork, star it and contribute to it!

Thank you :)

PR #1793, opened 2022-11-07
Fixing Carriage Return Line Feed (CRLF) order in docs #1792

Lahiry:
In the "Working with line-ends special symbols LF/CRLF" section of the documentation we can see:

The special template variable `_new_lines` enforces a specific line ending. Acceptable variables: `'\n\r'` for CRLF and `'\n'` for POSIX.

Here is example how to force line endings to CRLF on any deployment:

```
{
    "project_slug": "sample",
    "_new_lines": "\n\r"
}
```
But this order is mistaken, since:

"The traditional order, when both control characters are used, is Carriage Return, then Line Feed.

The reason for this goes back to the old ASR-33 Teletype.

When a Carriage Return is issued to an ASR-33, the print head, if it is near the right margin, takes over a tenth of a second to return to the left margin, plus there is a bit of "bounce" when the left margin is hit.

If the order were Line Feed, then Carriage Return, the first printed character might occur a tenth of a second after the Carriage Return, and thus might end up printing (as a smear) halfway across the page. But if Line Feed comes after Carriage Return then the time taken by the Line Feed provides extra time for the print head to complete it's trip."

[StackOverflow](https://stackoverflow.com/questions/1885900/order-of-carriage-return-and-new-line-feed)

So this PR aims at fixing the CRLF order to the traditional `'\r\n'`

### DIFF
--- a/docs/advanced/new_line_characters.rst
+++ b/docs/advanced/new_line_characters.rst
@@ -16,7 +16,7 @@ This allow template developers to have both types of files in the same template.
 Developers should correctly configure their ``.gitattributes`` file to avoid line-end character overwrite by git.
 
 The special template variable ``_new_lines`` enforces a specific line ending.
-Acceptable variables: ``'\n\r'`` for CRLF and ``'\n'`` for POSIX.
+Acceptable variables: ``'\r\n'`` for CRLF and ``'\n'`` for POSIX.
 
 Here is example how to force line endings to CRLF on any deployment:
 
@@ -24,5 +24,5 @@ Here is example how to force line endings to CRLF on any deployment:
 
     {
         "project_slug": "sample",
-        "_new_lines": "\n\r"
+        "_new_lines": "\r\n"
     }


yuvaldolev:
Hi @Lahiry 

We started working on a new community fork of the Cookiecutter project, named Cookieninja.
The fork is found at: https://github.com/cookieninja-generator/cookieninja

We have re-opened and merged your PR in the Cookieninja project at: https://github.com/cookieninja-generator/cookieninja/pull/29

We're trying to bring the life back to this amazing project.
We are committed to reviewing and merging each PR in up to 2 weeks from the day it's opened (if there are no issues with it).
We'd really like you to join the fork, star it and contribute to it!

Thank you :)

jensens (approved):

kurtmckee:
@jensens When this is merged, #1792 can be closed. (It is not linked and will not close automatically.)

Issue #1759, opened 2022-08-05
Add checkout details to context


JonZeolla:
* Cookiecutter version: 2.1.1
* Template project url: N/A
* Python version: 3.9.13
* Operating System: macOS

### Description:

I would like the context output available in `{{cookiecutter | pprint }}` to include the `--checkout` argument passed to the CLI, under a '_checkout' field. This will allow me to perform different actions in my post generation hooks if a non-default branch, tag, or commit was passed.

### What I've run:

```
from collections import OrderedDict
cookiecutter_context = {{cookiecutter | pprint }}
print(cookiecutter_context)
```

Related to https://github.com/cookiecutter/cookiecutter/issues/1422

Issue #1672, opened 2022-05-11
How to do proper logging from the hooks?


tdamsma:
I want to add log messages to the post_gen_hook, and ideally I would want to change the log level depending on whether I called cookiecutter with the -v option or not. I can't find any hints in the code or dcos how to achieve that, is it even possible?

I currently use this code in my post_gen_hook to access the cookiecutter context, but this does not contain the applied logging level or other log settings.
```
context = json.loads(
        r"""
        {{ cookiecutter | jsonify }}
        """
    )
```

In my post_gen _script I used `from cookiecutter.hooks import logger` and though this can access a logger and create output (Only for warning and above), but as the post_gen_hook is (I think) called in a subprocess I don't think it is actually the same logger

Issue #1499, opened 2021-02-04
Capture exception or stdout/stderr from hook subprocess


limdauto:
* Cookiecutter version: 1.7.2+
* Template project url: https://github.com/quantumblacklabs/kedro/tree/master/kedro/templates/project/%7B%7B%20cookiecutter.repo_name%20%7D%7D
* Python version: 3.8.5
* Operating System: OSX

### Description:

Currently, cookiecutter executes hooks in a [subprocess](https://github.com/cookiecutter/cookiecutter/blob/master/cookiecutter/hooks.py#L82-L87) but doesn't capture the exception / stdout / stderr of that process. For example, if I put a message inside `sys.exit`, that message is lost. We are building a CLI wrapper around cookiecutter and we would like to test the error message bundled with our CLI. Therefore, we would love to be able to capture the error message from hook execution and validate it against our list of expected error message.

### What I've run:

N/A

simobasso:
Hey @limdauto, thanks for your proposal! As you saw, there isn't the capacity to do it, but if you're still interested and you can provide a proposal or a POC, I'll more than happy to review it.

PR #1466, opened 2020-10-17
Added hooks: 'pre_context_prompt' and 'post_context_prompt'

tomaszkurgan:

PR #1796, opened 2022-11-15
Hook logging and new hook type `pre_context` which can modify the context

RobertKosten:
I've got an issue where I want to modify the context via a hook, in particular *before* anything is rendered (in my case it's pulling data from `git config`, but I believe it would also help with some open tickets, e.g. #1759, etc.; It has some overlap with  #1466, but since that hasn't been touched in two years...). The minimally-invasive way I found to do so was to allow a new type of hook script, `pre_context`, that gets run as early as possible and returns json, which is then overlaid on the existing context structure (it is not replaced wholesale because I assume most of these scripts would *add* completely new keys, copying the entire context would make such scripts needlessly more complex).

In the process I discovered that the scripts have no way to log error messages, they can just fail by exiting non-zero. I've introduced capturing `stderr` and handing it to the logger, that should make things easier. (See #1672, #1499)

If this pull request gets otherwise accepted I'll gladly write some accompanying docs, obviously.


RobertKosten:
I withdraw the PR. The trick schown in https://github.com/samj1912/cookiecutter-advanced-demo is sufficient for my needs.

PR #1803, opened 2022-12-06
feature: Add --dump-input flag to dump user input to file

msoikis:
* Add --dump-input flag to cli.py
* Add _generate_cookiecutter_json_file() with the dump logic to generate.py
* Rename zipfile.py to zipfile_utils.py to fix import collision with zipfile python package
* Correct tests with the above rename


kurtmckee (commented):
@msoikis It looks like this is very similar to an existing feature, [replaying project generation](https://cookiecutter.readthedocs.io/en/stable/advanced/replay.html). The replay feature also allows for re-use of the user input, using the `--replay` CLI option.

@jensens @pydanny @ssbarnea Please close this PR.

PR #1813, opened 2023-01-28
improving test_make_sure_path_exists_correctly_handle_os_error

andrehora:
The test `test_make_sure_path_exists_correctly_handle_os_error` is missing a case when the path is invalid, for example, Path('invalid\\path').

Hope this PR can make this test a bit better.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -70,6 +70,10 @@ def test_make_sure_path_exists_correctly_handle_os_error(mocker):
         utils.make_sure_path_exists(Path('protected_path'))
     assert str(err.value) == "Unable to create directory at protected_path"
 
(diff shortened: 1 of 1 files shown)


kurtmckee (commented):
@audreyfeldroy @jensens @tmeckel Please close this PR.

@andrehora This does not test what you think it's testing.

Above the code you added, `pathlib.Path.mkdir` has been patched to _always_ raise `OSError`. The purpose of this test is to validate that `OSError` is caught and handled correctly, not to exercise possible routes by which `OSError` might be raised in a real context.

If it were testing actual routes for `.mkdir()` to raise an `OSError`, `"invalid\\path"` would succeed because it's not an invalid path.

PR #1814, opened 2023-02-01
fix a typo

LazarusX:


LazarusX:
Close because the typo has already been fixed in the main branch.

PR #1824, opened 2023-03-24
feature: Allow jsonify to be used as function in templating

pvbouwel:
By allowing jsonify to be used as a function in jinja2 templates it is possible to easily render cookiecutter
context using `{{ jsonify(cookiecutter) }}`.

This can just re-use the existing logic.

Before investing in adding tests I want to check if this would be considered to be pulled in. I can use a local extension on my end so I do not depend on it but it seems something that could be useful to others.


kurtmckee (commented):
@audreyfeldroy @jensens @tmeckel Please close this PR.

`jsonify()` is already a filter and the author's use case can already be met by writing `{{ cookiecutter|jsonify }}`.

PR #1829, opened 2023-03-31
feat: Pass errors from pre/post-gen hook(s) to the FailedHookException message

vykozlov:
If cookiecutter functions are called in python code, it is useful to catch possible errors appearing during the execution of cookiecutter pre/post-generate hooks.

modified:   cookiecutter/hooks.py to:

* catch possible errors from the subprocess calling hook scripts
* add these errors in the FailedHookException message

The change is not breaking, all tests run to 100% coverage

Co-authored-by: vykozlov

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -79,11 +79,17 @@ def run_script(script_path, cwd='.'):
     utils.make_executable(script_path)
 
     try:
-        proc = subprocess.Popen(script_command, shell=run_thru_shell, cwd=cwd)  # nosec
-        exit_status = proc.wait()
+        proc = subprocess.Popen(script_command, shell=run_thru_shell, cwd=cwd,  # nosec
+                                stderr=subprocess.PIPE,   # pipe only errors
+                                universal_newlines=True)  # to convert from b' format
+        # catch error during hook execution, if any
+        sp_output, sp_error = proc.communicate()
+        # if hook does not yield an error message, don't attempt to publish
+        error_message = f': {sp_error}' if sp_error else ''
+        exit_status = proc.returncode
         if exit_status != EXIT_SUCCESS:
             raise FailedHookException(
-                f'Hook script failed (exit status: {exit_status})'
+                f'Hook script failed (exit status: {exit_status}){error_message}'
             )
     except OSError as err:
         if err.errno == errno.ENOEXEC:


.pre-commit-config.yaml
```diff
@@ -35,7 +35,7 @@ repos:
       - id: check-xml
       - id: check-yaml
         exclude: "not_rendered.yml|invalid-config.yaml"
-  - repo: https://gitlab.com/pycqa/flake8
+  - repo: https://github.com/pycqa/flake8
```
kurtmckee, line 38: This change has merged recently. Please rebase or merge `main` into your branch to pick up this change.

cookiecutter/hooks.py
```diff
@@ -79,11 +79,21 @@ def run_script(script_path, cwd='.'):
...
-        exit_status = proc.wait()
+        proc = subprocess.Popen(
+            script_command,
+            shell=run_thru_shell,  # nosec
+            cwd=cwd,
+            stderr=subprocess.PIPE,  # pipe only errors
+            universal_newlines=True,  # convert from b' format
+        )
+        # catch error during hook execution, if any
+        sp_output, sp_error = proc.communicate()
```
kurtmckee, line 90: `sp_output` is never used, so I recommend using an underscore as the variable name, which is common when indicating that the variable is unused.

```suggestion
        _, sp_error = proc.communicate()
```

cookiecutter/hooks.py
```diff
@@ -79,11 +79,21 @@ def run_script(script_path, cwd='.'):
...
+        # if hook does not yield an error message, don't attempt to publish
+        error_message = f': {sp_error}' if sp_error else ''
```
kurtmckee, line 92: This looks like a helpful feature for identifying failure causes. Nice!

kurtmckee (changes requested):
@vykozlov I've left feedback. Thanks for your work on this!

Would you add a test to demonstrate this new feature?

ericof (approved):
LGTM

ericof (changes requested):
@vykozlov Could you please add a test?

BorjaEst:
Hi added tests and pass for linux but not for windows:
I tried to raise an error message like this: `f.write("echo 'this err' >&2\n")` -to-> `f.write("echo this err 1>&2\n")` however it seems `f.write("echo this err 1>&2\n")` is not equivalent in windows and the test fails. Some windows supported that would know how ti do that small fix?

BorjaEst:
> Hi added tests and pass for linux but not for windows: I tried to raise an error message like this: `f.write("echo 'this err' >&2\n")` -to-> `f.write("echo this err 1>&2\n")` however it seems `f.write("echo this err 1>&2\n")` is not equivalent in windows and the test fails. Some windows supported that would know how ti do that small fix?

Seems was an issue with the space "err \n", I just replaced `==` by `in` at the assert.

BorjaEst (approved):

BorjaEst (approved):

BorjaEst (approved):

vykozlov:
@ericof , @kurtmckee , after... some time a colleague of mine @BorjaEst added requested tests. One CI/CD test (3.7) is failing but the complain is about vulnerability in Jinja2 version 3.1.4 which we did not change

kurtmckee (approved):
Thanks for adding the test!

Issue #1736, opened 2022-06-14
Cannot change jinja2 delimiters because NonTemplatedInputDirException exception


ericsouza:
* Cookiecutter version: `2.1.1`
* Python version: `3.10.0`
* Operating System: Ubuntu `20.04`

### Description:

Basically I need to use cookiecutter along side with files that use the same sintax from jinja2 for a template engine from golang (is for using with helm-charts). I saw this [PR](https://github.com/cookiecutter/cookiecutter/pull/1419). So I change jinja2 enviroment using the settings below in my `cookiecutter.json`:
```json
"_jinja2_env_vars": {
        "block_start_string": "[%",
        "block_end_string": "%]",
        "variable_start_string": "[[",
        "variable_end_string": "]]"
    }
```
for the content inside the files all went well, but now I'm getting `cookiecutter.exceptions.NonTemplatedInputDirException` because my project's folder doesn't use `{{` and `}}` 

I think the problem is this part of cookiecutter that does not use the jinja2 configuration: https://github.com/cookiecutter/cookiecutter/blob/449f8771650a3daab7c8163b015576eae26150b8/cookiecutter/find.py#L20

xcke:
For now the workaround seems to be use something like this for the project_dir: 

{{[[cookiecutter.project_name]]}}

inside the project directory files and folders can use the changed jinja2 start and end string

liortct:
I fixed this bug in the PR #1767.
I advice to merge the PR and close this issue.

liortct:
Hi @xcke , @ericsouza 

We started working on a new community fork of the Cookiecutter project, named Cookieninja.
The fork is found at: https://github.com/cookieninja-generator/cookieninja

We have re-opened your issue at: https://github.com/cookieninja-generator/cookieninja/issues/35

We're trying to bring the life back to this amazing project.
We are committed to addressing each issue in up to 1 week from the day it's opened.
We'd really like you to join the fork, star it and contribute to it!

Thank you :)

sacha-c:
I am still experiencing this issue in `v2.4.0`. Here is a proposal PR #1963

PR #1837, opened 2023-05-03
Bugfix/jinja2 change delimiters of the template folder

mariovagomarzal:
Addresses https://github.com/cookiecutter/cookiecutter/issues/1736.

Based on the bugfix by @liortct in https://github.com/cookiecutter/cookiecutter/pull/1767 but using the organization name throughout the project.


cookiecutter/find.py
```diff
@@ -3,21 +3,28 @@
...
     """
-    logger.debug('Searching %s for the project template.', repo_dir)
+    logger.debug("Searching %s for the project template.", repo_dir)
 
     for str_path in os.listdir(repo_dir):
-        if 'cookiecutter' in str_path and '{{' in str_path and '}}' in str_path:
+        if (
+            'cookiecutter' in str_path
+            and environment.variable_start_string in str_path
+            and environment.variable_end_string in str_path
```
kurtmckee, line 26: This, and the similar change in `generate.py`, are the crucial elements of this bugfix. Nice!

kurtmckee (approved):
@mariovagomarzal Thanks for re-triggering CI after the linting PR was merged, and for adding a test to demonstrate this is fixed!

@audreyfeldroy @jensens @tmeckel Please merge this PR to close #1736. Thanks!

jensens:
This changes the signature of the functions -> breaking change.

ericof:
As this is a breaking change, we will add it to the 3.0.0 milestone

kurtmckee:
Would it be sufficient to give the `environment` parameter a default of `None` and then conditionally use its settings? For example:

```python
def find(x, environment = None):
    left_delimiter = "{{"
    right_delimiter = "}}"
    if environment is not None:
        left_delimiter = environment.x
        right_delimiter = environment.y
    ...
```

Would that allow this to get merged immediately as a bugfix instead of a breaking change?

jensens:
Was this addressed with merge of #1997 ?

mariovagomarzal:
Yes, I think we can close this pull request.

Sorry for the late response.

PR #47, opened 2013-08-24
Add get_secret_key() function and crypto module for issue #33

pydanny:
- Adds a crypto module with a `get_secret_key()` function.
- In `main.cookiecutter()` the `crypto.get_secret_key()` function is injected into the cookiecutter context.
- Included test coverage.

Issue #33, opened 2013-08-19
{cookiecutter.secret_key}} should automatically inject random string


pydanny:
While the idea of pre-generate hooks (#29) is good, a great utility feature would be the  to generate a 50 character random string and place it automatically in a {cookiecutter.secret_key}} context variable.

PR #1839, opened 2023-05-10
Add secret key extension

hashem-nowruzi:


kurtmckee (commented):
@audreyfeldroy @jensens @tmeckel Please close this PR. It looks like a modern version of PR #47, but as discussed there and in issue #33, this can be implemented independently as a plugin rather than introducing this feature directly to cookiecutter.

PR #1841, opened 2023-05-18
Changes find function, test_find, to use Effect class from pymonad

jamesCodes808:
In this refactored code, the find_template function now returns an Effect instance representing the relative path to the project template. The search_template function performs the actual search logic within the effectful computation.

To run the effect and obtain the result, you can use the run() method as follows:

`
result = find_template("/path/to/repo").run()

if isinstance(result, Exception):
    # Handle the exception if the effect raised one
    print("Exception occurred:", result)
else:
    # Process the successful result
    print("The project template is:", result)

`

The refactored code essentially wraps the original functionality of searching for the project template in an Effect instance. This allows you to perform additional operations, such as logging, alongside the effectful computation. By using the Effect monad, you can encapsulate and manage the effectful computation in a more functional programming style.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ parts/
 sdist/
 var/
 venv/
+.venv/
 *.egg-info/
 .installed.cfg
 *.egg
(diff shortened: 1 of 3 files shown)


kurtmckee (commented):
@audreyfeldroy @jensens @tmeckel Please close this PR. It introduces a dependency on pymonad, which hasn't been updated in 2 years; changes the call semantics of `find_template()`; but does not also fix a bug or add a feature.

PR #1858, opened 2023-06-13
Update base docs, remove tox

### LINKED ISSUE #1857: Update CONTRIBUTING, README and package metadata
### Description:

- [x] Update `CONTRIBUTING.md` to mention the usage of `nox` instead of `tox`
- [x] Reorganize `README.md`
- [x] Update `setup.py` to add links to the project

ericof:
Fixes #1857


jensens (approved):

ericof:
@kurtmckee To be honest, it is not a big difference. And it seems our nice friends at `pypa` like it, so we are not adopting something that is unmaintained.

kurtmckee:
I much more strongly disagree with this now that I've attempted to run the tests on Windows.

nox is finding PyPy 3.9 and believes that it's Python 3.9. It's not able to correctly find 4 of my installed Python versions.

This is happening because PyPy 3.9 and Python 3.11 are the only ones on my PATH. I could modify my PATH, but I don't intend to do that.

nox appears to be an inferior product to tox and I would like to revert this change.

ericof:
@kurtmckee open a PR reverting all nox changes

PR #1864, opened 2023-06-16
Pretty-format JSON files

kurtmckee:
This PR swaps out the existing `check-json` hook for the `pretty-format-json` hook, which will enforce standards on the JSON files, not simply confirm that they can be parsed.

I recommend visiting pre-commit.ci and enabling it for this repo. This will have the following benefits:

* All PRs will be checked by pre-commit
* Changes introduced by the linters will be added as an additional commit to all incoming PRs
* Pre-commit hook versions will be regularly updated

This will reduce maintenance burden.


.pre-commit-config.yaml
```diff
@@ -29,7 +29,12 @@ repos:
       - id: check-symlinks
       - id: check-case-conflict
       - id: check-docstring-first
-      - id: check-json
+      - id: pretty-format-json
+        args:
+          - "--autofix"
+          - "--indent=4"
```
ericof, line 35: Could we use 2 instead of 4?

ericof (changes requested):
I would suggest we use two instead of 4 for the indent.
(pre-commit is configured now)

.pre-commit-config.yaml
kurtmckee: Works for me! I saw examples of both and picked one. I'll update this shortly.

kurtmckee, line 35:
```suggestion
          - "--indent=2"
```

ericof:
@kurtmckee Now we need to fix the files :-)

kurtmckee:
@ericof Haha, I thought that pre-commit.ci would kick in because of the new commit so I took a shortcut. Maybe it skips PRs that are already open?

I'll watch for it on net-new PRs to show up as a CI action like the `codecov/*` CI checks. :+1:

luzfcb:
@kurtmckee I have approved your request to install the pre-commit ci integration for the cookiecutter repository. It is now available for use.

These are the other GitHub Apps we have installed.

image


In case you don't know me, I'm one of the maintainers of cookiecutter-django, but I don't maintain cookiecutter.

Anyway, thank you for the cleanup job you are doing on the issues/pull-requests.

kurtmckee:
Thank you both for the quick PR review, and for setting up pre-commit.ci! I think that'll reduce maintenance burden.

ericof (approved):

PR #1865, opened 2023-06-16
Overcome problems running nox on Windows

kurtmckee:
This PR will help overcome nox's inability to find installed Python versions on Windows, and its inability to distinguish between CPython and PyPy implementations.

Besides this problem, nox is also currently slower than tox. These are the results of timing comparison between nox and tox on my machine:

```
> nox (first run)
Runtime: 8:29

> nox -r (second run; the user must ask nox to reuse venvs)
Runtime: 4:22

> tox (1st run)
Runtime: 7:59

> tox (2nd run)
Runtime: 2:35
```

In addition, tox is able to run test suites in parallel, but nox cannot.

I recommend reverting #1858 and standardizing on tox, but if that is not preferred, this PR will allow the test suite to run on Windows more reliably.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -2,12 +2,52 @@
 
 Nox is Tox tool replacement.
 """
+
+import os
 import shutil
+import sys
 from pathlib import Path
 
 import nox
 
 nox.options.keywords = "not docs"
+
+
+# At the time of writing, nox on Windows cannot reliably find installed Python versions,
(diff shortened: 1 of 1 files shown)


ericof:
Close this PR and focus on reverting the change to nox.

kurtmckee:
I'll be able to tackle this soon. The tox file I used for the timing comparison was a direct transfer of the current nox config and isn't yet production-ready.

PR #1866, opened 2023-06-17
Use tox

kurtmckee:
This PR introduces the following changes:

* CI is modified to skip running linters; this is now handled by pre-commit.ci
* A tox configuration is introduced that supports the same functionality as the nox configuration
* `safety` is automatically run locally only against a single Python version. In CI it will continue to run on all Python versions.
* `safety`'s default telemetry is now disabled.
* tox environment dependencies are configured to allow parallel execution as much as possible.

# Improvement opportunities found during this work

* The linting tests in CI are considered required. However, pre-commit.ci now handles linting. The CI configuration needs to be updated.

* Linting currently fails on Windows; it appears that `.gitattributes` is forcing platform-specific line endings that override developer git settings and cause the doc8 hook to fail all files. This can be addressed in a different PR.

* When building the documentation, myst consistently overwrites `docs/cookiecutter.rst` with slightly different contents than what's stored in git. This can be addressed in a different PR.

* pytest reports a UserWarning during execution; this should be caught by the test that induces it, rather than allowing it to bubble up to pytest. This can be addressed in a different PR.

* The test suite cannot be run fully in parallel due to file access conflicts (at least on Windows). File I/O should not be occurring in the same directory across multiple runs, so it will be beneficial to update the test suite so it uses unique directories, or to virtualize all I/O using pyfakefs. This will be a helpful improvement to allow for parallel execution, rather than serial execution, of the test suite.


ericof (approved):

ericof:
@kurtmckee We need to remove the linting checks, otherwise it is not possible to merge this here

PR #1870, opened 2023-06-19
Standardize newlines for all platforms

kurtmckee:
This resolves pre-commit hook failures that occur on Windows. The `.gitattributes` file forced CRLF newlines on all files via "text=auto", which caused linting to fail.

LF newlines are now enforced for all files except those with "crlf" in their filenames.


jensens (approved):
I think this will do it.

kurtmckee:
> I think this will do it.

Do you mean, "This will free you to merge?" My username doesn't have write access to the repo, so I'll need you to merge if the PR is approved for merging. :smile:

PR #1871, opened 2023-06-20
Add simple_test for custom test extensions

mlasevich:
Currently there is a decorator to create a simple filter Jinja extension. This PR adds a similar decorator to create custom test extensions. Included is the decorator, test updatres, and docs updates

### DIFF
--- a/cookiecutter/utils.py
+++ b/cookiecutter/utils.py
@@ -116,3 +116,15 @@ def simple_filter(filter_function):
 
     SimpleFilterExtension.__name__ = filter_function.__name__
     return SimpleFilterExtension
+
+
+def simple_test(test_function):
+    """Decorate a function to wrap it in a simplified jinja2 extension."""
+
+    class SimpleTestExtension(Extension):
+        def __init__(self, environment):
+            super().__init__(environment)
+            environment.tests[test_function.__name__] = test_function
+
+    SimpleTestExtension.__name__ = test_function.__name__
+    return SimpleTestExtension
--- a/docs/advanced/local_extensions.rst
+++ b/docs/advanced/local_extensions.rst
@@ -54,6 +54,28 @@ It's likely that we'd only want to register a single function as a filter. For t
 
 This snippet will achieve the exact same result as the previous one.
 
+Similarly, to register a single function as a test extension we can use
+the ``simple_test`` decorator:
+
+.. code-block:: json
+
+    {
+        "project_slug": "Foobar",
+        "year": "{% now 'utc', '%Y' %}",
+        "_extensions": ["local_extensions.simpletestextesion"]
+    }
+
+.. code-block:: python
+
+    from cookiecutter.utils import simple_test
+
+
+    @simple_test
+    def simpletestextesion(v):
+        return v in [True, "True", "true]
+
+
+
 For complex use cases, a python module ``local_extensions`` (a folder with an ``__init__.py``) can also be created in the template root.
 Here, for example, a module ``main.py`` would have to export all extensions with ``from .main import FoobarExtension, simplefilterextension`` or ``from .main import *`` in the ``__init__.py``.
 
--- a/tests/test-extensions/local_extension/cookiecutter.json
+++ b/tests/test-extensions/local_extension/cookiecutter.json
@@ -2,8 +2,11 @@
   "project_slug": "Foobar",
   "test_value_class_based": "{{cookiecutter.project_slug | foobar}}",
   "test_value_function_based": "{{cookiecutter.project_slug | simplefilterextension}}",
+  "test_value_function_test_pos_based": "{% if 'XAX' is palindrome%}PASS1{%else%}PASS1{% endif %}",
+  "test_value_function_test_neg_based": "{% if 'XAY' is palindrome%}FAIL2{%else%}PASS2{% endif %}",
   "_extensions": [
     "local_extensions.simplefilterextension",
+    "local_extensions.palindrome",
     "local_extensions.FoobarExtension"
   ]
 }
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -439,6 +439,10 @@ def test_local_extension(tmpdir, cli_runner):
     content = Path(output_dir, 'Foobar', 'HISTORY.rst').read_text()
     assert 'FoobarFoobar' in content
     assert 'FOOBAR' in content
+    assert 'PASS1' in content
+    assert 'FAIL2' not in content
+    assert 'PASS2' in content
+    assert 'FAIL2' not in content
 
 
 def test_local_extension_not_available(tmpdir, cli_runner):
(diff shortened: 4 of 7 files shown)


tests/test-extensions/local_extension/cookiecutter.json
```diff
@@ -2,8 +2,11 @@
   "project_slug": "Foobar",
   "test_value_class_based": "{{cookiecutter.project_slug | foobar}}",
   "test_value_function_based": "{{cookiecutter.project_slug | simplefilterextension}}",
+  "test_value_function_test_pos_based": "{% if 'XAX' is palindrome%}PASS1{%else%}PASS1{% endif %}",
+  "test_value_function_test_neg_based": "{% if 'XAY' is palindrome%}FAIL2{%else%}PASS2{% endif %}",
```
kurtmckee, line 6: These tests are not correct and are not clear (`PASS1` is duplicated, for example).

Please change these lines to something like this, update the test suite to look for `"XX=palindrome"` and `"XY=text"`, and then confirm that the test suite passes:

```suggestion
  "test_value_function_test_pos_based": "XX={% if 'XX' is palindrome%}palindrome{%else%}text{% endif %}",
  "test_value_function_test_neg_based": "XY={% if 'XY' is palindrome%}palindrome{%else%}text{% endif %}",
```

tests/test_cli.py
```diff
@@ -439,6 +439,10 @@ def test_local_extension(tmpdir, cli_runner):
     content = Path(output_dir, 'Foobar', 'HISTORY.rst').read_text()
     assert 'FoobarFoobar' in content
     assert 'FOOBAR' in content
+    assert 'PASS1' in content
+    assert 'FAIL2' not in content
+    assert 'PASS2' in content
+    assert 'FAIL2' not in content
```
kurtmckee, line 445: Apply the changes above and change this to:

```suggestion
    assert 'XX=palindrome' in content
    assert 'XY=text' in content
```

kurtmckee (changes requested):

tests/test-extensions/local_extension/cookiecutter.json
mlasevich: I would really like to have an excuse for this, but looking for it, all I find is embarrassment 😊

mlasevich:
All the corrections have been done, would it be possible to get this reviewed again :-)

Thank you. :-)

cookiecutter/utils.py
```diff
@@ -116,3 +116,15 @@ def __init__(self, environment):
 
     SimpleFilterExtension.__name__ = filter_function.__name__
     return SimpleFilterExtension
+
+
+def simple_test(test_function):
+    """Decorate a function to wrap it in a simplified jinja2 extension."""
```
kurtmckee, line 122: This docstring is identical to the one above it (and neither clearly states what the extension does). Would this be clearer?

```suggestion
    """Register a function for use as a test in Jinja templates."""
```

docs/advanced/local_extensions.rst
```diff
@@ -54,6 +54,28 @@ It's likely that we'd only want to register a single function as a filter. For t
...
 
+Similarly, to register a single function as a test extension we can use
+the ``simple_test`` decorator:
+
+.. code-block:: json
+
+    {
+        "project_slug": "Foobar",
+        "year": "{% now 'utc', '%Y' %}",
+        "_extensions": ["local_extensions.simpletestextesion"]
```
kurtmckee, line 65:
```suggestion
        "_extensions": ["local_extensions.simpletestextension"]
```

docs/advanced/local_extensions.rst
```diff
@@ -54,6 +54,28 @@ It's likely that we'd only want to register a single function as a filter. For t
...
+    }
+
+.. code-block:: python
+
+    from cookiecutter.utils import simple_test
+
+
+    @simple_test
+    def simpletestextesion(v):
```
kurtmckee, line 74:
```suggestion
    def simpletestextension(v):
```

docs/advanced/local_extensions.rst
```diff
@@ -54,6 +54,28 @@ It's likely that we'd only want to register a single function as a filter. For t
...
+        return v in [True, "True", "true]
```
kurtmckee, line 75:
```suggestion
        return v in [True, "True", "true"]
```

mlasevich: *sigh*  I really need to proofread my own stuff :-(

cookiecutter/utils.py
```diff
@@ -116,3 +132,30 @@ def __init__(self, environment):
 
     SimpleFilterExtension.__name__ = filter_function.__name__
     return SimpleFilterExtension
+
+
+def simple_test(test_function):
+    """
+    Register a function for use as a test extension for Jinja templates.
```
kurtmckee, line 139: I think flake8 is upset with this:

```suggestion
    """Register a function for use as a test extension for Jinja templates.
```

cookiecutter/utils.py
```diff
@@ -107,7 +107,23 @@ def prompt_and_delete(path, no_input=False):
 
 
 def simple_filter(filter_function):
-    """Decorate a function to wrap it in a simplified jinja2 extension."""
+    """
+    Register a function for use as a filter extension for Jinja templates.
```
kurtmckee, line 111: I think flake8 is upset with this:

```suggestion
    """Register a function for use as a filter extension for Jinja templates.
```

cookiecutter/utils.py
```diff
@@ -107,7 +107,23 @@ def prompt_and_delete(path, no_input=False):
...
+
+    This decorator allows creation of a filter extension to Jinja as a function
+    instead of writing a full class extending jinja2's Extension. The function
+    name becomes the filter name
```
kurtmckee, line 115:
```suggestion
    name becomes the filter name.
```

cookiecutter/utils.py
```diff
@@ -107,7 +107,23 @@ def prompt_and_delete(path, no_input=False):
...
+
+    Example:
+
+    @simple_test
+    def mirror(value):
```
kurtmckee, line 120: This will be rendered in the API documentation. This should be a more complete example, including RST syntax:

```suggestion
        .. codeblock:: python

            @simple_test
            def mirror(value):
                return value[::-1]
```

This suggestion applies to the code changes lower in this file, too.

cookiecutter/utils.py
```diff
@@ -107,7 +107,23 @@ def prompt_and_delete(path, no_input=False):
...
+
+    Which can then be used in jinja template as
+
+    {{ cookiecutter.param | mirror }}
```
kurtmckee, line 124: This needs to be a codeblock so that it renders in the API documentation. I don't think pygments has a Jinja lexer specifically, so I'm suggesting `text` here to avoid incorrect syntax highlighting:

```suggestion
    .. codeblock:: text

        {{ cookiecutter.param | mirror }}
```

This suggestion applies to the code changes lower in this file, too.

kurtmckee (changes requested):

PR #1879, opened 2023-07-02
Update `.gitignore` and cite where it was copied from

kurtmckee:
The source is clear from the various in-file comments; now it is updated and cited.

I routinely use `.venv/` as my virtual environment directory, which was not included in the old version of the Python-gitignore file. Now it *is* included, and I don't have to mentally ignore that directory when committing changes. 🥳 🎈 📈

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
+# Source:
+# https://raw.githubusercontent.com/github/gitignore/main/Python.gitignore
+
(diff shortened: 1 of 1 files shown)


pydanny (approved):
LGTM

kurtmckee:
Huzzah! Would you be willing to merge in that case?

ericof (approved):

PR #1881, opened 2023-07-04
Added support for providing human-readable prompts to the different variables

vemonet:
Hi @pydanny @jensens @ericof @audreyfeldroy, I added support for providing human-readable questions to the different variables

Many issues mentioned it:

* https://github.com/cookiecutter/cookiecutter/issues/30
* https://github.com/cookiecutter/cookiecutter/issues/249
* https://github.com/cookiecutter/cookiecutter/issues/794
* https://github.com/cookiecutter/cookiecutter/issues/1567
* https://github.com/cookiecutter/cookiecutter/issues/1835

A PR has been attempted but abandonned: https://github.com/cookiecutter/cookiecutter/pull/848

There have been lengthy discussions about wherever the json format should completely be changed, introduce breaking change, etc... I took the simplest approach, to **introduce the least amount of technical debt and keep complete backward compatibility** . So that we can merge this feature, really much needed for a templating library, and discuss about changing the JSON format later!

It was quite straightforward to implement, I just added a `__prompts__` object, which takes the variable names as key and the questions to ask the user as values.

Example of a config file:

```json
{
    "package_name": "my-package",
    "module_name": "{{ cookiecutter.package_name.replace('-', '_') }}",
    "short_description": "A nice python package",
    "__prompts__": {
        "package_name": "Select your package name:",
        "module_name": "Select your module name:"
    }
}
```

The 2 first questions will use the description provided, the 3rd one will use the variable name since there is no description provided

**This format stays easy to read, and easy to write**. Once the dev defined all its variables with their default they can copy them, paste them in the `__prompts__` block, and replace the default by the questions one by one. Also wrap something around  2 underscores is a pythonic way to say "this is special" (`__init__`, `__main__`)

<details><summary>In my opinion when you have many variables it is more readable  than a nested dict where some keys are repeated all other the place</summary>

```json
{
    "package_name": {
        "default": "my-package", 
        "prompts": "Your package name",
    },
    "module_name": {
        "default": "{{ cookiecutter.package_name.replace('-', '_') }}", 
        "prompts": "Your package name",
    },
}
```

</details>

It is **100% backward compatible with the previous `cookiecutter.json` format 🎉** so it will not break existing template, it will only make templates that have defined `__prompts__` better!

- [x] **Not a lot of code have been added**, and the changes introduced don't introduce a lot of complexity in the codebase
- [x] I made sure **all tests pass** with `tox`, and I added a few tests for the new feature, to reach **💯% of code coverage**
- [x] I have updated the example in the README to show how to use those new descriptions 
- [x] A page has been added in the advanced tab in readthedocs documentation 

I hope there is no more excuses to not merge this long-awaited feature 😄 I am already using it for my [templates](https://github.com/MaastrichtU-IDS/cookiecutter-python-package/blob/main/cookiecutter.json) and will continue in the future since those templates are fully compatible!

There are no breaking changes so this feature can be pushed to the next minor release without any concern 🥳

Please let me know if we should use something else instead of `__prompts__`, I can make the changes required in a few minutes (maybe `_prompts_` with only 1 underscore? that might be less prone to user mistakes, but the double underscore seems more pythonic)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,8 @@ parts/
 sdist/
 var/
 venv/
-*.egg-info/
+.venv/
+*.egg-info
 .installed.cfg
 *.egg
 
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ Projects are generated to your current directory or to the target directory if s
   ```py
   {{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}/{{cookiecutter.repo_name}}.py
   ```
-- Simply define your template variables in a `cookiecutter.json` file.
+- Simply define your template variables in a `cookiecutter.json` file. You can also add human-readable questions that will be prompted to the user for each variable using the `__prompts__` key.
   For example:
 
   ```json
@@ -126,7 +126,7 @@ Projects are generated to your current directory or to the target directory if s
     "release_date": "2013-07-10",
     "year": "2013",
     "version": "0.1.1",
-    "__questions__": {
+    "__prompts__": {
       "full_name": "Provide your full name",
       "email": "Provide your email"
     }
--- a/README.md
+++ b/README.md
@@ -125,7 +125,11 @@ Projects are generated to your current directory or to the target directory if s
     "project_short_description": "Refreshingly simple static site generator.",
     "release_date": "2013-07-10",
     "year": "2013",
-    "version": "0.1.1"
+    "version": "0.1.1",
+    "__questions__": {
+      "full_name": "Provide your full name",
+      "email": "Provide your email"
+    }
   }
   ```
 - Pre- and post-generate hooks: Python or shell scripts to run before or after generating a project.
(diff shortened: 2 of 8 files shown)


ericof:
Thanks, @vemonet! As soon as the tests pass, I will merge this!

ericof (approved):
Great addition!

henryiii:
Does this support customizing choice prompts? I'd like to show something more helpful for selecting choices, but don't see how this fits into this structure. If I provided a dict here, how would I customize the top level prompt? Maybe a special key?

```json
{
    "__prompts__": {
        "backend": { 
            "__prompt__": "Select a backend to build your package:",
            "setuptools": "Classic setuptools configuration",
            "setuptools621": "Modern pyproject.toml setuptools configuration"
       }
    }
}
```
?

(This is great, by the way!)

vemonet:
Hi @henryiii , it does not support this, but is an interesting suggestion 

And the solution you proposed is nice, and the easiest to implement I think. I'll take a look into it when I will have time

A complete JSON would look like this:

```json
{
    "package": "my-package",
    "backend": [
         "setuptools", 
         "setuptools621"
    ],
    "__prompts__": {
        "package": "Your package slug",
        "backend": { 
            "__prompt__": "Select a backend to build your package:",
            "setuptools": "Classic setuptools configuration",
            "setuptools621": "Modern pyproject.toml setuptools configuration"
       }
    }
}
```

PR #1890, opened 2023-07-07
docs: add warning for jinja2_time

henryiii:
See https://github.com/cookiecutter/cookiecutter/issues/1888. This is a suggested docs addition.


tranzystorek-io:
I wonder if it would be better to instead make the whole example more generic and obvious, e.g. by specifying a placeholder extension like `foobar.FoobarExtension`.

A warning after a potentially confusing example might just make the whole doc more confusing effectively.

ericof (approved):
Thank you, @henryiii!

PR #1901, opened 2023-07-17
Improve style of prompts using `rich`

vemonet:
Hi, this PR adds [`rich`](https://rich.readthedocs.io/en/stable/introduction.html) to the `cookiecutter` dependencies and improve the style of the prompts without changing the prompt workflow

* Add a counter, e.g.  `[1/10]`, at the start of each prompt so the user knows how many steps are left
* Default values are shown between parenthesis in the cyan color
* This enables cookiecutter template creators to use simple markup to style their questions, e.g. `What is [bold yellow]your name[/]"
* The answers accepted for the different types of prompts are the same as before: 
  * `y/t/true/yes/on` or `n/f/false/no/off` for boolean questions
  * Using integers for multiple choices questions
  * Json prompt supported the same was as before
* Add a 2 spaces indent at the start of each prompts

You can find screenshots of the changes proposed in the issue https://github.com/cookiecutter/cookiecutter/issues/1885 

I used `rich` as it seems the most mature, maintained, and popular package to style CLI apps in python


.github/ISSUE_TEMPLATE/bug.yml
ericof, line 1: Could you remove this file?

ericof (changes requested):
Remove the bug.yml

.github/ISSUE_TEMPLATE/bug.yml
vemonet: Oops, that was not supposed to be pushed in this branch! Well spotted, I removed it @ericof

ericof (approved):

PR #1904, opened 2023-07-23
Fix replay

vemonet:
An error where the context defined in the replay file was not taken into account was introduced with https://github.com/cookiecutter/cookiecutter/pull/1758 (which added prompts for entries missing from the replay file). It seems to impact a few people: https://github.com/cookiecutter/cookiecutter/issues/1900 and https://github.com/cookiecutter/cookiecutter/issues/1902)

This PR:
- Properly uses the context extracted from the replay file
- Fix the tests related to it, removed one useless test (too many mocks, the tests are just testing mocks at some point :p)

Closes https://github.com/cookiecutter/cookiecutter/issues/1900 
Closes https://github.com/cookiecutter/cookiecutter/issues/1902

@ericof @kurtmckee

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,7 @@
 # Adapted from https://github.com/github/gitignore/blob/main/Python.gitignore
 
+tests/tmp/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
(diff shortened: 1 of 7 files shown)


ericof:
@vemonet Could you please provide tests to confirm both issues are really fixed?

vemonet:
Hi @ericof and @kurtmckee 

I added a test that generates a project with a replay file, and check if the replayed values are properly used in the generated file 

It fails if you remove the changes I made to fix the replat (because the replayed values are not passed to the context), and works with my change

ericof (approved):

PR #1913, opened 2023-08-06
dump-input flag

adva10:


kurtmckee:
This appears to duplicate the existing [replay file](https://cookiecutter.readthedocs.io/en/stable/advanced/replay.html) functionality.

I'm closing this PR because it appears to duplicate existing functionality. Is there a better way to advertise the replay file feature?

PR #1918, opened 2023-08-11
Development

zykon2004:
Added a new feature called dump-input which dumps all user input and cookiecutter context into the created project directory.


zykon2004:
Should not go here

PR #1923, opened 2023-08-21
add checkout details to the context (fixes #1759)

### LINKED ISSUE #1759: Add checkout details to context

JonZeolla:
It seems like I can't associate this with #1759 other than referencing it here

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -6,7 +6,6 @@ from pathlib import Path
 
 import pytest
 from click.testing import CliRunner
-
 from cookiecutter import utils
 from cookiecutter.__main__ import main
 from cookiecutter.environment import StrictEnvironment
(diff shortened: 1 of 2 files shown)


tests/test_cli.py
```diff
@@ -432,6 +431,7 @@ def test_echo_undefined_variable_error(output_dir, cli_runner):
             '_template': template_path,
             '_repo_dir': template_path,
             '_output_dir': output_dir,
+            '_checkout': None,
```
JonZeolla, line 434: I’m also willing to write additional use cases if necessary, but because of how these private context keys are managed, they often aren’t available in happy path outlets, so I would just be writing a test case very similar to this one, but with a customized check out being reflected in the context.

Open to other testing approaches. One I considered was creating and checking a replay file

ericof (approved):

PR #1924, opened 2023-08-22
Default values can be passed as a dict

matveyvarg:
I'm using cookiecutter programatically and generating some config values (_replay_dir_ to be specific) dynamically. So, I need ability to pass it from the outside. 
Maybe I'm wrong, but it seems like current version doesn't allow to do that. Creating config file time for every call of the cookiecutter sounds like an overhead to me.
Here's my small suggestion how this functionality can be added without introducing breaking changes into the API.


tests/test_get_user_config.py
```diff
@@ -154,3 +154,13 @@ def test_expand_vars_for_directories_in_config(monkeypatch):
     user_config = config.get_user_config(config_file)
     assert user_config['replay_dir'] == 'Users/bob/cookies/replay-files'
     assert user_config['cookiecutters_dir'] == 'Users/bob/cookies/templates'
+
+
+def test_specify_config_values():
+    """Validate provided custom config values should be respected ."""
```
kurtmckee, line 160:
```suggestion
    """Validate provided custom config values should be respected."""
```

cookiecutter/config.py
```diff
@@ -93,6 +96,10 @@ def get_user_config(config_file=None, default_config=False):
     If the environment variable is not set, try the default config file path
     before falling back to the default config values.
     """
+    # Do NOT load a config. Merge  provided values with defaults and return them instead
```
kurtmckee, line 99:
```suggestion
    # Do NOT load a config. Merge provided values with defaults and return them instead
```

kurtmckee (approved):
This looks good to me. Thanks for including a test case!

ericof (approved):
LGTM

PR #1931, opened 2023-09-03
Update main.py

### LINKED ISSUE #1733: [enhancement] A check for Python version
* Cookiecutter version: 2.1.1
* Template project url:  cookiecutter-adafruit-circuitpython-ansys-cookiecutter.zip
* Python version: Many
* Operating System: Linux 4.15

### Description:

Would be nice, if cookiecutter checks the Python version and exits gracefully in case of unsupported version.
Now, with
Python 3.6, cookiecutter throws ugly "path too long" error
Python 3.10: Exits due to import error.
In both cases a novice user may have trouble to find out that the reason is the Python version.

gouravraghuwanshi:
FIXED #1733

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -201,6 +201,3 @@ class _patch_import_path_for_repo:
 
     def __exit__(self, type, value, traceback):
         sys.path = self._path
-
-    
-


kurtmckee:
I don't think that this type of condition is a best practice.

In the associated issue, I've recommended getting CircuitPython to update their install instructions.

PR #1942, opened 2023-09-22
Gracefully handle files with mixed lined endings

EricHripko:
## Overview
According to [Python docs](https://docs.python.org/3/library/io.html?highlight=newlines#io.TextIOBase.newlines), `newlines` can be one of:
- `None`
- string (if line endings are consistent)
- Tuple of strings (if multiple line endings are used)

Meanwhile, `open` builtin takes `None` or a string representing the new line according to [Python docs](https://docs.python.org/3/library/functions.html#open). Hence, there is a mismatch between the two for the case of multiple line endings. `cookiecutter` will get something like `newlines=("\n", "\r\n")` and try to pass that to `open` function. In turn, this will throw an error that looks like this:
```
TypeError: open() argument 'newline' must be str or None, not tuple
```

This PR proposes that `cookiecutter` takes the first detected new line when it encounters template files with mixed line endings. Whilst this means that original line endings aren't completed preserved, this seems like a better user experience than throwing a `TypeError`. It should also make `cookiecutter` more resilient against (arguably) malformed inputs.

Fixes https://github.com/cookiecutter/cookiecutter/issues/1737.

## Testing
- Added a test `test_generate_file_handles_mixed_line_endings` to reproduce the issue, it only passes with the proposed patch ✅ 
- Existing tests pass locally too ✅ 
- Confirmed that proposed changes are covered via the coverage report ✅ 

## Notes
- New `test_generate_file_handles_mixed_line_endings` test asserts that either CRLF or LF newlines are returned, as Python docs seem to make no guarantees on what order line endings are returned in case of file having mixed newlines.


ericof (approved):
LGTM

EricHripko:
Thank you very much for landing this 🙌

PR #1949, opened 2023-09-28
Update README.md to include installation instructions

david-abn:


README.md
```diff
@@ -23,6 +23,12 @@ A command-line utility that creates projects from **cookiecutters** (project tem
 - PyPI: [https://pypi.org/project/cookiecutter/](https://pypi.org/project/cookiecutter/)
 - Free and open source software: [BSD license](https://github.com/cookiecutter/cookiecutter/blob/main/LICENSE)
 
+## Installation
+
+Install cookiecutter using pip package manager:
+```
+python3 -m pip install --user cookiecutter
```
kurtmckee, line 30: I think that installing via pipx would be a much stronger suggestion here. It's also surprising to still see references to 'python3', but this may be a difference in platforms (e.g. maybe it's still necessary to reference 'python3' on macOS?

I suggest this alternative phrasing:

```suggestion
# pipx is strongly recommended.
pipx install cookiecutter

# If pipx is not an option,
# you can install cookiecutter in your Python user directory.
python -m pip install --user cookiecutter
```

PR #1956, opened 2023-10-04
1928 - allow for customised naming of hooks and sort hooks for execution

david-abn:
__Issue__
https://github.com/cookiecutter/cookiecutter/issues/1928 

__Testing__
Created ```post_gen_project_2.sh``` in new cookiecutter template with
```
#!/bin/bash

touch test_post_gen_project_2
```
Verified that this file is created, and the files are run in sorted order.

__Documentation__
Updated documentation with this new functionality

### DIFF
--- a/cookiecutter/hooks.py
+++ b/cookiecutter/hooks.py
@@ -27,17 +27,19 @@ EXIT_SUCCESS = 0
 def valid_hook(hook_file, hook_name):
     """Determine if a hook file is valid.
 
+    Hook name must be a substring of an element in _HOOKS variable to be valid.
+
     :param hook_file: The hook file to consider for validity
     :param hook_name: The hook to find
     :return: The hook file validity
     """
     filename = os.path.basename(hook_file)
     basename = os.path.splitext(filename)[0]
-    matching_hook = basename == hook_name
-    supported_hook = basename in _HOOKS
+    matching_hook = hook_name in basename
+    supported_hook = [i for i in _HOOKS if i in basename]
     backup_file = filename.endswith('~')
 
-    return matching_hook and supported_hook and not backup_file
+    return matching_hook and bool(supported_hook) and not backup_file
 
 
 def find_hook(hook_name, hooks_dir='hooks'):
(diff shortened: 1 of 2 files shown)


ericof:
@david-abn I like this PR. Could you add a new test to show this is working?

david-abn:
> @david-abn I like this PR. Could you add a new test to show this is working?

Added new commit to update unit. Thank you

cookiecutter/hooks.py
```diff
@@ -27,17 +27,21 @@
...
+
     :param hook_file: The hook file to consider for validity
     :param hook_name: The hook to find
     :return: The hook file validity
     """
     filename = os.path.basename(hook_file)
     basename = os.path.splitext(filename)[0]
-    matching_hook = basename == hook_name
-    supported_hook = basename in _HOOKS
+    matching_hook = hook_name in basename
```
LKajan, line 40: I think it should be a requirement that file names start with the hook identifier to avoid unintended execution of files. For example, without this constraint, a file like `utils_for_pre_gen_project.py` could run as a hook unintentionally.
```suggestion
    matching_hook = basename.startswith(hook_name)
```

cookiecutter/hooks.py
```diff
@@ -27,17 +27,21 @@
...
+    supported_hook = [i for i in _HOOKS if i in basename]
```
LKajan, line 41: The row above already tests that the file is a requested hook type.
It should now be enogh to test if the requested hook type is supported.
```suggestion
    supported_hook = hook_name in _HOOKS
```
Alternatively the `any()` function could be utilized to get a boolean value right away
```suggestion
    supported_hook = any(basename.startswith(supported_hook_name) for supported_hook_name in _HOOKS)
```

tests/test_generate_hooks.py
```diff
@@ -59,14 +59,22 @@ def test_run_python_hooks():
 def test_run_python_hooks_cwd():
     """Verify pre and post generation python hooks executed and result in current dir.
 
+    Check that multiple hooks can be created with python_post and python_post_2.
+    python_post_2 should be created before python_post to verify it is sorted.
```
LKajan, line 63:
```suggestion
    python_post_2 should be created after python_post to verify it is sorted.
```

Issue #1960, opened 2023-10-15
Dictionary variables: example from docs doesn't work any more after 2.2+


sandroden:
Starting from cookiecutter 2.2 and surely on 2.4 the example of dictionary variables from the docs doesn't work anymore.

I should probably say: It doesn't work as I was used to using it... and it seemed perfectly reasonable to me.

It seems that the current implementation doesn't allow anything not written in `cookiecutter.json` while up to 2.1.1 you could just give a template of what you would have added. In the example docs, if you wanted to add different file_types you could just pass them in the context (when working from Python script) and it just worked. 
Now we completely lost any dynamicity, you have to write down all the possibilities in advance, which is not an option in many cases. 

Let me know if I just overlooked the docs and missed how to add dynamicity, if not please consider how many people were relying on that feature or could benefit from that feature. I could provide tons of examples... currently, I was looping on users on docker services, on ports, on files... all this just stopped working.

The docs  say _"Dictionary variables provide a way to define deep structured information when rendering a template."_.
It states *structure* not *data*! The current implementation meddles with *data*, you can't send a file_type if not in `cookiecutter.json`

If I understood correctly the move with 2.2 was to create a nested choice type (please tell me if I'm wrong), but that should not be as the expenses of the possibility already there from 1.5+ to loop on dynamic values.

BTW: If I create an `extra_context` that only has one `file_type`, the other is added anyhow, I would consider this definitely a bug

tied:
Maybe I have the same problem. With cookiecutter 2.1.1 everything works fine.

After that line the try exits with an exception:
```python
    try: 
        context = {{ cookiecutter | jsonify }}
        ...
    except BaseException as e:
        print('Applications create errors! ' + str(e))
        sys.exit(1)
```
`name 'null' is not defined` doesn't really make sense to me because I don't have any var which is called `name`.
Here my `cookiecutter.json`
```json
{
    "customer_name": "",
    "customer_slug": "{{ cookiecutter.customer_name|lower|replace(' ', '-') }}",
    "default_environments": "production,test",
    ...
}
```
Is it possible that everything before the underscore is ignored?

Issue #1954, opened 2023-10-03
disruptive behaviour between 2.1.1 and 2.2 in dictionary variables


sandroden:
I have a template that has this `cookiecutter.json`:

```
 "pg_ports": {"port_list": [] },
```
I'm using cookiecutter from python scripts and send values as `extra_context` as:

```
extra_context = {
    ....
    'pg_ports':` {'port_list': ['5452:5432']},
}
```
It worked as expected in 2.1.1 and stopped working in 2.2+
the error I get is:

```
Traceback (most recent call last):
  File "/home/local/scripts/jmb", line 11, in <module>
    load_entry_point('jmb.tools', 'console_scripts', 'jmb')()
  File "/usr/lib/python3/dist-packages/click/core.py", line 764, in __call__
    return self.main(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/core.py", line 717, in main
    rv = self.invoke(ctx)
  File "/usr/lib/python3/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3/dist-packages/click/core.py", line 1137, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/usr/lib/python3/dist-packages/click/core.py", line 956, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/usr/lib/python3/dist-packages/click/core.py", line 555, in invoke
    return callback(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/click/decorators.py", line 17, in new_func
    return f(get_current_context(), *args, **kwargs)
  File "/home/misc/src/hg/thunder/jmb2/jmb-tools3/jmb/tools/conf/docker.py", line 120, in compose_from_yaml
    cookiecutter(url, **kwargs)
  File "/usr/local/lib/python3.7/dist-packages/cookiecutter/main.py", line 128, in cookiecutter
    extra_context=extra_context,
  File "/usr/local/lib/python3.7/dist-packages/cookiecutter/generate.py", line 130, in generate_context
    apply_overwrites_to_context(obj, extra_context)
  File "/usr/local/lib/python3.7/dist-packages/cookiecutter/generate.py", line 82, in apply_overwrites_to_context
    apply_overwrites_to_context(context_value, overwrite)
[...]

PR #1692, opened 2022-06-06
support partially overwrite keys in nested dict

cksac:
cookiecutter.json
```json
{
    "full_name": "Raphael Pierzina",
    "github_username": "hackebrot",
    "project": {
        "name": "Kivy Project",
        "description": "Kivy Project description",
    }
}
```

user_config.yaml
```yaml
default_context:
  project:
    name: "My Project"
```

Current generated context:
```json
{
    "full_name": "Raphael Pierzina",
    "github_username": "hackebrot",
    "project": {
        "name": "My Project",
    }
}
```
the whole project is replace by user config instead of replace project.name only.

Expected generated context:
```json
{
    "full_name": "Raphael Pierzina",
    "github_username": "hackebrot",
    "project": {
        "name": "My Project",
        "description": "Kivy Project description",
    }
}
```

PR #1961, opened 2023-10-16
Fix recursive context overwrites

padraic-padraic:
In 2.2.0, `generate.apply_overwrites_to_context` was amended with a recursive call to allow merging fields inside dictionary variables (#1692 )

In 2.3.0, it was further amended to allow merging of multichoice variables (#1903). 

However, these two changes interact to mean that cookiecutter began treating any list field in a dictionary variable as a choice (see #1954, #1960 for bugs running into this issue). If all the entries in a list field are hashable, then this may not cause problems, but if e.g. you have complex objects with lists of dictionary variables then the call to set introduce in #1903 would also cause a crash.

This PR fixes the merge behaviour by adding a simple recursive flag when applying the overwrites to dictionary variables. Cookiecutter itself will only treat "top-level" lists as choice variables, so this is consistent with the interaction model and the docs which state "The dictionary values can, themselves, be other dictionaries and lists - the data structure can be as deep as you need."

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -45,7 +45,8 @@ def is_copy_only_path(path, context):
     return False
 
 
-def apply_overwrites_to_context(context, overwrite_context):
+def apply_overwrites_to_context(context, overwrite_context, *,
+                                in_dictionary_variable=False):
     """Modify the given context in place based on the overwrite_context."""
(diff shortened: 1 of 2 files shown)


ericof (approved):
Thanks for the fix

PR #1963, opened 2023-10-18
fix: modifying start and end variable strings

sacha-c:
**Description**

Cookiecutter supports using _jinja2_env_vars to modify the jinja environment. However when modifying `variable_start_string` or `variable_end_string` it will not work because we are hard-coding the start and end strings for the main directory creation (see [here](https://github.com/cookiecutter/cookiecutter/blob/2.4.0/cookiecutter/find.py#L20)). In addition, the jinja context is not always populated with the _jinja2_env_vars, for example when getting the [placeholder prompts here](https://github.com/cookiecutter/cookiecutter/blob/2.4.0/cookiecutter/prompt.py#L240)

**Technical details**

This change does 3 things (which are in separate commits for readability):
1. Creates a common utility function `create_env_with_context` so the jinja environment is created similarly and with context each time.
   - Note that I also considered creating a single jinja env object & passing it around, however it requires much more refactoring (especially of tests), and we need to create it at least twice anyway (once to create the cookiecutter prompt, and once after the prompts are passed to the context), so I decided to go for the easier solution for now.
2. Leverages the _jinja2_env_vars to search for the initial cookiecutter directory, to allow this customization of variable_start_string and variable_end_string.
3. Small refactor to move `prompt_and_delete` from `utils` to `prompt`. It was causing circular import issues when having `prompt` import `utils`, because `utils` was importing `prompt`. And in any case I would say it makes more sense for it to live in `prompt` because it is a prompt as well after all.

I also removed the `ensure_dir_is_templated` check because the same logic is used to find the dir, and is therefore redundant.

**Use-case**
In react-native projects, it's extremely common to use `{{` and `}}` to pass objects to components. For example:
```tsx
<MyComponent
    some-prop={{ key: 'value' }}
</MyComponent>
```

Since `{{` is the default jinja variable string prefix, jinja goes crazy with react-native templates 🤪 .  
A solution to this is to set `_jinja2_env_vars.variable_start_string == '{%{'` (or something else). But cookiecutter currently has issues when overriding this setting.

**Testing**

Tested locally on a react-native project with `_jinja2_env_vars.variable_start_string == '{%{'` and `_jinja2_env_vars.variable_end_string == '}%}'`


sacha-c:
After further testing, I found that the placeholders used to generate the `context` are rendered _without_ using the `_jinja2_env_vars`.

This means that currently a file with:
```json
{
    "name": "Easy Project",
    "slug": "{%{ cookiecutter.name.lower().replace(' ', '-') }%}",
    "repo_path": "elements/elements-react-native/{%{ cookiecutter.slug }%}",
    "author_name": "John Doe",
    "author_email": "john.doe@elements.nl",
    "description": "The greatest project ever created.",
    "use_renovate": [
        "yes",
        "no"
    ],
    "_jinja2_env_vars": {
        "variable_start_string": "{%{",
        "variable_end_string": "}%}"
    }
}
```

will not work, because cookiecutter renders `{%{ cookiecutter.name.lower().replace(' ', '-') }%}` using `{{` instead of the provided override `{%{`

The issue is that [https://github.com/cookiecutter/cookiecutter/blob/2.4.0/cookiecutter/prompt.py#L233](prompt_for_config) is being called without passing the `envvars` to the jinja context...

I am setting this PR to draft and will try to make the improvements for this to work.

kurtmckee:
@sacha-c Thank you for your work on this! However, this appears to be a duplicate of #1837.

I'm closing this as a duplicate, but please comment back if there's functionality that your PR addresses that isn't addressed in #1837. Thank you!

sacha-c:
@kurtmckee thanks for taking a look! I really appreciate it 👍 

And thanks for linking that PR, I had totally missed it when investigating this issue.

However reading through it it looks like it is missing an additional fix
Basically there are two issues:
1. `variable_start_string` & `variable_end_string` are not taken into account when looking for the cookiecutter root directory
   - This is included in the PR you linked, and in mine
2. `_jinja2_env_vars` not taken into account when generating the updated cookiecutter context
   - This is missing from the PR you linked, I only discovered it while further testing my changes on a project. [see my comment above](https://github.com/cookiecutter/cookiecutter/pull/1963#issuecomment-1777330179)

I did like the use of `Environment.variable_start_string` rather than hardcoding default values as I was doing, so I've updated my PR with that improvement

I would suggest closing the other PR in favor of this one.

kurtmckee:
Oh, nice! Thanks for responding back!

...and I just discovered that I only have repo permissions to close PRs but not to re-open them. I assumed that I could re-open PRs. 😞

@ericof and @jensens, please re-open this PR and review in light of the additional bug fix that this offers over #1837 (interpreting template openers/closers in variables in `cookiecutter.json`).

sacha-c:
Hello!
I see this issue is still closed, so just want to make sure it is still on the radar, if it can be re-opened and reviewed
Thanks! @kurtmckee @ericof @jensens

sacha-c:
Hello,
What is the state of this PR, would you like me to create a new PR in order reopen it? 
Thanks in advance
@kurtmckee @ericof @jensens

kurtmckee:
I cannot re-open PRs; @ericof it will be helpful if you can reopen this. Thanks!

kurtmckee:
@sacha-c If this doesn't get re-opened, please open a new PR. My apologies for these difficulties!

ericof:
Unfortunately I could not reopen (at least not in mobile app)

PR #1991, opened 2023-12-02
[1518] Directory Name Render and Create Fix

DanielZhangD:
This pull request addresses the issue described in https://github.com/cookiecutter/cookiecutter/issues/1518. Written by @rafyak and @DanielZhangD

The solution was to add a check to render_and_create_dir to make sure the dirname was not empty.

Testing was completed via adding a new unit tests for render_and_create_dir and making sure coverage stayed at 100%


jensens (dismissed):

jensens (approved):

flying-sheep:
Wait. I want the exact opposite, i.e. the ability to skip directories when rendering.

It works for files by using an empty filename, so it should work for directories as well.

Issue #1694, opened 2022-06-06
Remove hard-codded extended template folder


insspb:
```
env.loader = FileSystemLoader(['.', '../templates'])
```
Should be configurable, documentation should be updated. ``../templates`` can leaved as default.

@simobasso I will merge #1485 without this fix, as you waited very long and may be not interested in fixing this simple case. But if you have time, please make a fix PR for this issue, so I can review other PRs. Thank you.

Related to #1485

PR #1992, opened 2023-12-02
[1694] Remove hardcoded template folder 

KareemElMaaddawy:
This PR address the issue #1694.

The proposed solution here is to have a default template path of `['.', '../templates']` and if the user would like to use the `FileSystemLoader` function to load other templates, then they are able to do so by adding paths to `additional_template_paths = []`.

I have also modified `test_templates.py ` to account for the changes I have made.

Ensured that `tox` coverage is 100%.


felipemontinoLGC (approved):

cookiecutter/generate.py
```diff
@@ -344,7 +344,14 @@ def generate_files(
...
     with work_in(template_dir):
-        env.loader = FileSystemLoader(['.', '../templates'])
+        default_template_path = ['.', '../templates']
+        # Option for additional tenplate paths (configurable, can be an empty list)
+        # Usage:
+        # additional_template_paths = ['/path/to/more/templates']
+        additional_template_paths = []
+        all_template_paths = default_template_path + additional_template_paths
+
+        env.loader = FileSystemLoader(all_template_paths)
```
jensens, line 354: Sorry, I do not get how this changes anything?

jensens:
There was no reply on this for a month.
I do not see any real change in this RP. See comment above.
I close this request for now. If you plan to work further on this please reopen, answer and go on.

PR #2012, opened 2024-02-04
Use 'Ruff' for linting and formatting

danieleades:
Adds a minimal Ruff configuration. I've chosen this configuration to match as closely as possible the behaviour of the existing linting and formatting, and have the least possible impact on existing code.

Ruff is a couple of orders of magnitude faster than flake8, which is particularly handy when using it as a pre-commit hook. It also has a massive range of additional lints that can be applied, and is able to autofix a large proportion of them.

If adding additional linting or more aggressive formatting is desirable that can be added (and discussed) in separate PRs. I've left the other lint groups in the config file, commented out. Suggest that these can be switched on one by one in other PRs (or removed) as appropriate.


danieleades:
@jensens anything i can do to move this along?

jensens:
Looks overall good to me, just one thing, maybe personal flavor of mine: Why a ruff.toml if the very same configuration could be placed in pyproject.toml ?

danieleades:
> Looks overall good to me, just one thing, maybe personal flavor of mine: Why a ruff.toml if the very same configuration could be placed in pyproject.toml ?

I'll shift it over. Originally I was trying to minimise the impact on the original config, but then found it made it fully redundant

jensens (approved):

jensens:
Thanks!

PR #2013, opened 2024-02-04
Prefer double quotes for strings

danieleades:
This PR is stacked on #2012, so should be rebased when (if) that gets merged.

This repo currently uses a mixture of single and double quoted strings. This PR standardises all quotes to the ruff/black default (double quotes).

If single quotes are preferred that's an easy config change. I'm in favour of double, but I think either is better than a mixture of the two

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -60,6 +60,5 @@ repos:
     rev: v0.2.0
     hooks:
       - id: ruff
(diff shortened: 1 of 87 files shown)


danieleades:
i've just noticed the string quoting conventions in the contributing guide. Seems a bit complex to me but at least it's documented. I'll leave this PR here for discussion, but I can see that this change will be in violation of the current style guide.

PR #2014, opened 2024-02-04
Add 'pyflakes' lints

danieleades:
this PR is stacked on #2012 so should be rebased when (if) that's merged.

Adds the 'pyflakes' lint group to the ruff config


danieleades:
@jensens this is ready for review

jensens (approved):

PR #2020, opened 2024-02-04
enable lint groups for string formatting

danieleades:
stacked on #2012


danieleades:
@jensens this is ready for review

jensens (approved):

PR #2049, opened 2024-03-28
remove some unused args

danieleades:
lints for and removes unused function arguments


danieleades:
@jensens this is ready to go!

PR #2051, opened 2024-03-30
Add type hints for cookiecutter.cli and check it with mypy

danieleades:


danieleades:
@jensens any idea why `typing_extensions` is not present in the docs test environment? I tried adding it explicitly to the `test_requirements` but that doesn't seem to have done the trick

jensens (approved):

PR #2052, opened 2024-03-30
Drop support for Python 3.7

danieleades:
drops support for python 3.7
python 3.7 is end of life


danieleades:
> I think this is fine. @cookiecutter/cookie-monsters What do you think?

great group name

ericof:
I'm 👍🏼 but I would like to test what happens with pipx in 3.7 trying to get the latest cookiecutter

ericof (approved):

jensens:
According to Semantic Versioning this is a major change, right? I would prefer to wait for a 3.0 release to introduce this.

danieleades:
i don't think there's any rush to merge this. It's dead easy to keep rebasing.

It really only becomes important once you want to use newer features, or you want to update the dependencies and those dependencies have dropped support for 3.7.

I personally drop EOF python versions like a hot brick, but it would be perfectly reasonable to hold onto it until it's actually causing an issue and then pull the cord on this PR

kurtmckee (approved):
@ericof / @jensens CI no longer passes because requests has known security bugs that are not getting released for Python 3.7. Let's get this merged and released.

@ericof The pipx question is easy: requiring Python 3.8 means that pipx on Python 3.7 will not find newer versions for installation.

@jensens This isn't a breaking change -- nobody on Python 3.7 could install this after release, so they won't see anything break. No need to wait for a v3 release.

I recommend merging this asap and releasing it.

danieleades:
I'll get this cleaned up and rebased

jensens (approved):

PR #2053, opened 2024-03-30
shrink mypy whitelist for cookiecutter.utils

danieleades:


danieleades:
@jensens this is ready to go!

PR #2057, opened 2024-04-06
use pathlib for paths

danieleades:


danieleades:
mypy coverage still not good enough to do this effectively

danieleades:
having another go now mypy coverage is higher

danieleades:
CI failures due to #2063

danieleades:
@jensens this is getting pretty close to ready.

I will need some help to go back through this PR and ensure that I haven't made breaking changes to the API.
For example, I've changed some signatures from `str` to `Path`, which is fine for internal APIs but a breaking change for public APIs. There are at least some instances of this to be addressed. Can you help me spot them?

My intention is to change them to `Path | str`, and add a deprecation warning for uses of `str`, so that they can be changed to `Path` in a future major release

jensens:
@danieleades Great! But ... Ok, this is kind of massive. I need to take some time to review all this. I'll come back in the next days, no time today or tomorrow.

jensens (changes requested):
I would not change the signature of functions. At least not in the 2.x series. If we want to merge and release this one in the 2.x series (what I prefer) allow both style and output a DeprecationWarning for the 3.x release, like so:
```python
import warnings
if isinstance(path, str):
    warnings.warn(
        "String paths are deprecated for parameter aaa in function bbb and will be removed in release 3.0", 
        DeprecationWarning)
        path = Path(path)
````

PR #2059, opened 2024-04-06
run mypy on tests directory

danieleades:


danieleades:
remaining errors are due to this monstrosity:

```python
def run_pre_prompt_hook(repo_dir: Path | str) -> Path | str:
    """Run pre_prompt hook from repo directory.

    :param repo_dir: Project template input directory.
    """
    # Check if we have a valid pre_prompt script
    with work_in(repo_dir):
        scripts = find_hook('pre_prompt')
        if not scripts:
            return repo_dir

    # Create a temporary directory
    repo_dir = create_tmp_repo_dir(repo_dir)
    with work_in(repo_dir):
        scripts = find_hook('pre_prompt') or []
        for script in scripts:
            try:
                run_script(script, str(repo_dir))
            except FailedHookException as e:  # noqa: PERF203
                raise FailedHookException('Pre-Prompt Hook script failed') from e
    return repo_dir
```

which sometimes returns a `str` and sometimes a `Path`.

there are places in the codebase where it fails if it doesn't return a `str`, and other places where it fails if it doesn't return a `Path`, so it's not just a matter of refactoring it to return only one or the other.

tests/replay/test_dump.py
```diff
@@ -35,13 +35,13 @@ def fin_remove_replay_file() -> None:
...
     with pytest.raises(TypeError):
-        replay.dump(replay_test_dir, None, context)
+        replay.dump(replay_test_dir, None, context)  # type: ignore[arg-type]
 
 
 def test_type_error_if_not_dict_context(replay_test_dir, template_name) -> None:
     """Test that replay.dump raises if the context is not of type dict."""
     with pytest.raises(TypeError):
-        replay.dump(replay_test_dir, template_name, 'not_a_dict')
+        replay.dump(replay_test_dir, template_name, 'not_a_dict')  # type: ignore[arg-type]
```
danieleades, line 44: i personally feel type safety makes this test (and ones like it) redundant. @jensens?

jensens: I am fine removing these tests in favor of type checks.

jensens:
@danieleades well, somethings wrong after rebase/conflict solving? Any idea?

danieleades:
> @danieleades well, somethings wrong after rebase/conflict solving? Any idea?

I'll take a look this evening!

danieleades:
@jensens should be good now

PR #2061, opened 2024-04-07
add more ruff lints

danieleades:


jensens:
This one needs manual conflict resolution.

danieleades:
> This one needs manual conflict resolution.

that should do it!

failing CI jobs due to #2063

cookiecutter/generate.py
```diff
@@ -250,8 +253,7 @@ def generate_file(
 
     logger.debug('Writing contents to file %s', outfile)
 
-    with open(outfile, 'w', encoding='utf-8', newline=newline) as fh:
-        fh.write(rendered_file)
+    Path(outfile).write_text(rendered_file, encoding='utf-8', newline=newline)
```
kurtmckee, line 256: Wow, ruff really boned this line. CI fails on Pythons 3.9 and lower.

This will need to be addressed.

kurtmckee (changes requested):

cookiecutter/generate.py
danieleades: done

cookiecutter/prompt.py
```diff
@@ -391,7 +391,8 @@ def choose_nested_template(
 
     template = Path(template) if template else None
     if not (template and not template.is_absolute()):
-        raise ValueError("Illegal template path")
+        msg = "Illegal template path"
+        raise ValueError(msg)
```
kurtmckee, line 395: I don't prefer ruff; changes like this, and others here, reinforce that view.

Nevertheless, enforcing style can be valuable, and I'm generally okay with the additional linting.

jensens (approved):

cookiecutter/prompt.py
jensens: Actually I have no idea why ruff thinks this is a good idea, but OTOH it does not hurt too.

danieleades: this lint is because python prints both the error message and the context. When the context also contains the message this is redundant.

ie. this would print something like (paraphrasing)

```
error: Illegal template path
`raise ValueError("Illegal template path")`
```

but if the msg is defined outside the context you get

```
error: Illegal template path
`raise ValueError(msg)`
```

in the trace

cookiecutter/generate.py
```diff
@@ -250,7 +253,7 @@ def generate_file(
 
     logger.debug('Writing contents to file %s', outfile)
 
-    with open(outfile, 'w', encoding='utf-8', newline=newline) as fh:
+    with open(outfile, 'w', encoding='utf-8', newline=newline) as fh:  # noqa: FURB103 (false positive for python < 3.10)
```
danieleades, line 256: see https://github.com/astral-sh/ruff/issues/12222

cookiecutter/generate.py
danieleades: this is now fixed upstream in ruff - https://github.com/astral-sh/ruff/issues/12222

that shouldn't stop this PR from getting merged, but the error suppression will need to be removed next time ruff is updated

PR #2063, opened 2024-04-24
don't test old interpreters on macOS and windows

danieleades:
the `macos-latest` github action runners no longer support python interpreters older than 3.10. This PR retains testing of older interpreters on linux, but only tests 'newish' ones for macos and windows


jensens (approved):

jensens:
@danieleades Do you have an idea why the six combinations are still in the list of checks here in the PR?

danieleades:
> @danieleades Do you have an idea why the six combinations are still in the list of checks here in the PR?

yes, they're configured as 'required checks' in the repo settings

PR #2071, opened 2024-05-24
refactor: make _get_version work in python binary

simkimsia:
I want to use cookiecutter as part of a python binary.

However, I faced two obstacles.

One obstacle was because of the use of `Path` in `cookiecutter/__init__.py`.

This PR is to replace the use of `Path` with `importlib.resources` which works just as well, but with the added bonus of it working within a python binary.
 
See https://github.com/GreenDeploy-io/cookiecutter-for-binary/wiki/Why%3F for details


jensens (changes requested):
`read_text` is deprecated since Python 3.11, see https://docs.python.org/3/library/importlib.resources.html#importlib.resources.read_text

kurtmckee (commented):
@simkimsia There is far too much happening in this PR.

I'm on board for using `importlib.resources`, but the GitHub workflow that only works with your `use-importlib-resources` branch, the `executable_script.py` file, the `cookiecutter.spec` file, and commenting out code in `hooks.py` goes far beyond what's needed for this change.

@jensens I recommend closing this PR. A new PR with a single commit that only touches `cookiecutter/__init__.py` can be submitted if needed.

simkimsia:
Understood @kurtmckee 

I'll close this and do a new or that only touches __init__

PR #2077, opened 2024-06-10
Update repository.py / Fix handling of file:// URLs in is_repo_url fn

edgarefraindp:
This change ensures that file-based repository URLs are correctly identified as valid.
Sry... I don't see CICD-Test for Ubuntu


kurtmckee (changes requested):
I'm not clear what this is fixing. This appears to shuffle some logic, but I don't yet see anything that fixes handling of `file://` URLs.

Could you give me an example of a `file://` URL that wasn't working previously but works now?

PR #2091, opened 2024-08-22
Nested configuration files recurse templates.

MatthewJM96:
Nested configuration files are great, but right now the only configuration for which a `templates` directory is made accessible to jinja2 is the leaf configuration. Combining these two ideas by providing an ordered list of the ascendant `templates` directories allows shared assets between configurations.

In order for this to work I also had to change how the repo directory is set up whenever a copy is made in running the pre_prompt hook. Instead of just copying the so-far-descended-to layer of any nested configurations, I copy the whole structure, tracking the depth of recursion into cookiecutter calls to achieve this and establish the FilesystemLoader correctly.

I don't think my solution is the most elegant, but it didn't require too much change to the code to achieve, and ensures any side-effects the directory copies were made to deal with will continue to behave as expected.


caerulescens:
I would really love if this could get merged; this is exactly what I need right now. Thank you @MatthewJM96  for contributing this!

Issue #2028, opened 2024-03-05
Feature-Request: Use inheritance with nested configuration


donnyyung:
* Cookiecutter version: 2.6.0
* Template project url:
* Python version: 3.11
* Operating System: Macos

### Description:

I'm trying to use nested [configuration files ](https://cookiecutter.readthedocs.io/en/stable/advanced/nested_config_files.html#) since we have one repo with a multiple templates for users to create an app whether it's python or go or some other language.  Some of the files like our README, our circle pipelines, etc. will be the same templates for all the services, so I was also trying to use [inheritance](https://cookiecutter.readthedocs.io/en/stable/advanced/templates.html) to have a set of shared files at the top of the repo that I can use in both templates, the file structure looks like: 

main-directory/
    ├── project-1
    │   ├── cookiecutter.json
    │   ├── {{cookiecutter.project_slug}}
    |       │   ├── ...
    ├── package
    │   ├── cookiecutter.json
    │   ├── {{cookiecutter.project_slug}}
    |       │   ├── ...
    ├── cookiecutter.json
    ├── templates/
     └── README_tpl.md
    
 And then I have a file in project-1/cookiecutter.project_slug/README.md that contains: 
 ```
 {% include "README_tpl.md" %}
 ```   
### What I've run:
And when I try to create an app using cookiecutter from the root dir, I get this error: 
```
Traceback (most recent call last):
  File "/usr/local/bin/cookiecutter", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "/usr/local/Cellar/cookiecutter/2.6.0/libexec/lib/python3.12/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/cookiecutter/2.6.0/libexec/lib/python3.12/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
         ^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/cookiecutter/2.6.0/libexec/lib/python3.12/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/Cellar/cookiecutter/2.6.0/libexec/lib/python3.12/site-packages/click/core.py", line 783, in invoke
[...]

PR #2103, opened 2024-09-20
Implement ability to override inherited templates directory in user configuration

RyanWhite25:
This change reconciles the usage of nested configurations with a shared inheritance directory by allowing a user to customise the relative path to the inherited templates.

See #2028 for a related feature request.

The initial approach to this PR passed this value on the CLI, but required a lot of boilerplate changes. It feels cleaner from a code & version control perspective to have this as a value stored in config instead.


butterlyn (approved):

jessebrizzi (approved):

alexcochran:
Could have one more approving review from a dev w/ write access, please? This merge would greatly benefit a few of my projects

jensens (changes requested):
Thanks for this PR. The feature is useful but there's a **security issue that must be fixed before this can be merged**.

## Critical Security Issue: Path Validation

The current implementation has no path validation whatsoever. This allows templates to reference arbitrary filesystem locations:

```python
# Current code - UNSAFE
inheritance_template_dir = context['cookiecutter'].get(
    '_inheritance_template_dir', '../templates'
)
env.loader = FileSystemLoader(['.', inheritance_template_dir])
```

A malicious or poorly-written template can set `_inheritance_template_dir` to:
- `/etc/` - read system files
- `../../../../home/user/.ssh/` - read SSH keys
- Any absolute path on the system

Since Jinja2's `FileSystemLoader` will happily load from these locations and templates execute arbitrary code via hooks anyway, this expands the attack surface significantly.

**Required fix**: Add strict path validation:
1. Reject absolute paths
2. Resolve the path and verify it stays within acceptable boundaries
3. Handle errors gracefully when paths don't exist

This isn't optional - templates already execute arbitrary code, but we shouldn't make it easier to access arbitrary files on the filesystem.

## Naming

`_inheritance_template_dir` is quite verbose. Consider shorter alternatives:
- `_templates_dir` - clear and concise
- `_includes_dir` - matches Jinja2 include terminology

## Testing

Your test only covers the happy path. You need tests for:
- Invalid paths (non-existent directories)
- Absolute paths (should be rejected)
- Directory traversal attempts (`../../../../etc`)
- How the system behaves when the path is inaccessible

Also: fix the typo "inheritence" → "inheritance" in the test docstring.

## Documentation

The documentation needs to explicitly state:
- Paths are relative to the project directory (clarify exactly what that means)
- Security warning: only use templates from trusted sources
- What happens if the path is invalid

## Summary

This feature can't be merged without proper path validation. The current implementation creates a security vulnerability. Once that's fixed with appropriate tests, this should be good to go.

PR #2105, opened 2024-10-04
fix post gen project hook code example

mschoettle:
os.rmtree fails if the directory is not empty.

Switch to shutil.rmtree which can delete a non-empty directory.

### DIFF
--- a/docs/advanced/hooks.rst
+++ b/docs/advanced/hooks.rst
@@ -108,11 +108,12 @@ A ``pre_gen_project`` hook can validate template variables. The following script
 
 **Conditional File/Directory Removal:**
 
-A ``post_gen_project`` hook can conditionally control files and directories. The example below removes unnecessary files based on the selected packaging option.
+A ``post_gen_project`` hook can conditionally control files and directories. The example below removes unnecessary files and directories based on the selected packaging option.
 
 .. code-block:: python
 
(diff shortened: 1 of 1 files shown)


kurtmckee (approved):
This seems okay to me. However, the context is that the two paths that could be removed should always be files, so I would therefore prefer that this not account for directories at all and simply error if `requirements.txt` or `poetry.lock` is a directory.

docs/advanced/hooks.rst
```diff
@@ -122,4 +123,4 @@ A ``post_gen_project`` hook can conditionally control files and directories. The
     for path in REMOVE_PATHS:
         path = path.strip()
         if path and os.path.exists(path):
-            os.unlink(path) if os.path.isfile(path) else os.rmdir(path)
+            os.unlink(path) if os.path.isfile(path) else shutil.rmtree(path, ignore_errors=True)
```
kurtmckee, line 126: `path` is exclusively `requirements.txt` or `pyproject.toml`. In context, these are understood to _always_ be Python packaging files, and will _never_ be directories.

In other words, there should be no call to `os.rmdir()`, nor to `shutil.rmtree()`. This code should instead simply be:

```python
if path and os.path.isfile(path):
    os.unlink(path)
```

Issue #2114, opened 2024-11-19
Attribute `_path` Defined Outside `__init__`


joero888:
* Cookiecutter version: 2.6.0
* Template project url: NA
* Python version: 3.12.6
* Operating System: NA

### Description:

The attribute `_path` is being used in the class without being properly initialized in the `__init__` method. This can lead to potential issues such as `AttributeError` when trying to access `_path` in other methods. The suggestion is to move the line 'self._path = copy(sys.path)'  into the `__init__` method to ensure _path is properly initialized when an instance of the class is created.

### Original Code Snippet: main.py

```
class _patch_import_path_for_repo:  # noqa: N801
    def __init__(self, repo_dir: Path | str) -> None:
        self._repo_dir = f"{repo_dir}" if isinstance(repo_dir, Path) else repo_dir

    def __enter__(self) -> None:
        self._path = copy(sys.path)
        sys.path.append(self._repo_dir)

    def __exit__(self, _type, _value, _traceback):  # type: ignore[no-untyped-def]
        sys.path = self._path

```

malek72:
Hi, my name is Malek, and I'm a student at the University of Toronto. For one of my courses, my task is to fix an issue in an open-source GitHub repository and submit a PR) for it. Would it be okay if I tried to resolve that issue for my assignment?

joero888:
Hello Malek,

I am also a student and I was hoping to do a PR for this issue, since I am taking a class on Open source too.

Rosemary
________________________________

joero888:
Hello Malek,

I am also a student and I was hoping to do a PR for this issue, since I am taking a class on Open source too.

Rosemary

________________________________

malek72:
Hey Rosemary,

I think I resolved the issue [🤣](https://emojipedia.org/rolling-on-the-floor-laughing) I didn't know that you were willing to fix that issue too.

kurtmckee:
Hello @joero888! While it's generally worthwhile to define instance attributes in `.__init__()`, I'm closing this issue because this is a context manager, and realistic usage of the class precludes the possibility that `._path` won't be set when it needs to be.

Regarding you and @malek72 doing this for classes, I'd like to reach out to your professors to provide feedback about the approach of the class. Could you @-mention the professors' GitHub usernames here, or perhaps email me their contact info?

You can find my email address on my GitHub profile page while you're logged into GitHub. Thanks!

PR #2115, opened 2024-11-19
Fixing Issue #2114 Attribute _path Defined Outside __init__

malek72:
Pull Request addresses the issue #[2114](https://github.com/cookiecutter/cookiecutter/issues/2114#issue-2670485270).

Ensured _path properly initialized in _patch_import_path_for_repo Class.
Previous behavior: Allowed _path to be initialized only when _enter_ function was called.
Could lead to potential issues in the case _exit_ was invoked without prior _enter_ which could cause an AttributeError. (As mentioned in the #[2114](https://github.com/cookiecutter/cookiecutter/issues/2114#issue-2670485270) issue)
PR updates to initialize _path in __init__ function and include an if statement for it in the __exit__ function.

New unit test test_patch_import_path_for_repo was added in the test_main.py file and made sure all test passed at 100%.

### DIFF
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -203,12 +203,12 @@ def cookiecutter(
 class _patch_import_path_for_repo:  # noqa: N801
     def __init__(self, repo_dir: Path | str) -> None:
         self._repo_dir = f"{repo_dir}" if isinstance(repo_dir, Path) else repo_dir
-        self._path = None  # Intialize _path to None
+        self._path: list[str] | None = None  # Intialize _path to None
 
     def __enter__(self) -> None:
         self._path = copy(sys.path)
         sys.path.append(self._repo_dir)
 
-    def __exit__(self, _type, _value, _traceback):  # type: ignore[no-untyped-def]
+    def __exit__(self, _type, _value, _traceback) -> None:  # type: ignore[no-untyped-def]
         if self._path is not None:  # Check if _path is initialized
             sys.path = self._path
--- a/cookiecutter/main.py
+++ b/cookiecutter/main.py
@@ -203,10 +203,12 @@ def cookiecutter(
 class _patch_import_path_for_repo:  # noqa: N801
     def __init__(self, repo_dir: Path | str) -> None:
         self._repo_dir = f"{repo_dir}" if isinstance(repo_dir, Path) else repo_dir
+        self._path = None # Intialize _path to None
 
     def __enter__(self) -> None:
         self._path = copy(sys.path)
         sys.path.append(self._repo_dir)
 
     def __exit__(self, _type, _value, _traceback):  # type: ignore[no-untyped-def]
-        sys.path = self._path
+        if self._path is not None: # Check if _path is initialized
+            sys.path = self._path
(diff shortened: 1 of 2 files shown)


cookiecutter/main.py
```diff
@@ -203,10 +203,12 @@ def cookiecutter(
 class _patch_import_path_for_repo:  # noqa: N801
     def __init__(self, repo_dir: Path | str) -> None:
         self._repo_dir = f"{repo_dir}" if isinstance(repo_dir, Path) else repo_dir
+        self._path: list[str] | None = None  # Intialize _path to None
```
kurtmckee, line 206: Comments like this are generally undesirable; the code is too obvious to need a comment, so the comment ends up adding noise instead of clarity.

cookiecutter/main.py
```diff
@@ -203,10 +203,12 @@ def cookiecutter(
...
 
     def __enter__(self) -> None:
         self._path = copy(sys.path)
         sys.path.append(self._repo_dir)
 
-    def __exit__(self, _type, _value, _traceback):  # type: ignore[no-untyped-def]
-        sys.path = self._path
+    def __exit__(self, _type, _value, _traceback) -> None:  # type: ignore[no-untyped-def]
+        if self._path is not None:  # Check if _path is initialized
+            sys.path = self._path
```
kurtmckee, line 214: This check is not needed because `.__enter__()` and `.__exit__()` are part of the context manager contract. No calling code will ever call `.__exit__()` and find that `._path` is None.

This situation can be synthetically created, as you demonstrated in the test you added, but it's not a realistic issue to address.

tests/test_main.py
```diff
@@ -114,3 +116,19 @@ def test_custom_replay_file(monkeypatch, mocker, user_config_file) -> None:
...
+def test_patch_import_path_for_repo():
+    """Test the _patch_import_path_for_repo context manager."""
+    original_sys_path = sys.path[:]
+    repo_dir = '/fake/repo/path'
+
+    patch = _patch_import_path_for_repo(repo_dir)
+    patch.__enter__()
+    assert sys.path[-1] == repo_dir
+    patch.__exit__(None, None, None)
+    assert sys.path == original_sys_path
```
kurtmckee, line 130: `.__enter__()` and `.__exit__()` are part of Python's context manager protocol; using `_patch_import_path_for_repo()` as a context manager would make this simpler:

```suggestion
    with _patch_import_path_for_repo(repo_dir):
        assert sys.path[-1] == repo_dir
    assert sys.path == original_sys_path
```

tests/test_main.py
```diff
@@ -114,3 +116,19 @@ def test_custom_replay_file(monkeypatch, mocker, user_config_file) -> None:
...
+
+    patch = _patch_import_path_for_repo(repo_dir)
+    patch.__exit__(None, None, None)
```
kurtmckee, line 133: This is not a realistic situation.

kurtmckee (commented):
Hi @malek72! I saw that you posted that you're a student in one of the cookiecutter issues so I've left feedback.

Because of its usage in practice, there's no need to guard for `._path` to be set in `.__init__()`. However, I'll also point out that the patcher class could be replaced with a simpler context manager function, like this:

```python
import contextlib


@contextlib.contextmanager
def _patch_import_path_for_repo(repo_dir: Path | str):
    _path = copy(sys.path)
    sys.path.append(repo_dir)
    try:
        yield
    finally:
        sys.path = _path
```

PR #2129, opened 2025-02-13
Preserve boolean types from cookiecutter.json in render_variables()

### LINKED ISSUE #2126: Can't define a private variable as boolean type
It seems like it's not possible to assign private variables as booleans. E.g. in cookiecutter.json if I have:

```
{
  "my_var": true,
  "__my_var_private": "{{ cookiecutter.my_var }}",
}
```

Then `__my_var_private` is a string rather than a boolean and I can't see any other way of doing it. Am I missing something? Is there a way to do this?

Thanks!

mnothman:
Pull request fixes issue where boolean values in created 'cookiecutter.json' are converted into strings. Now the 'render_variable()' function preserves boolean values instead of turning everything from the .json into strings

### Issue Addressed
Fixes #2126 

-Resolves issue where: 
boolean 'true' or 'false' in 'cookiecutter.json' would be rendered as string instead of boolean. 

Now *booleans remain booleans** after rendering in render_variable(), string value are unaffected

### Testing
Added test in tests/test_generate_file.py titled 'def test_boolean_variable_preserved'

All other tests pass except test 'test_should_invoke_main' which failed BEFORE and AFTER my changes. Appears to be issue with missing dependency ('binaryornot') didn't look into it much

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -200,7 +200,7 @@ def render_variable(
     env: Environment,
     raw: _Raw,
     cookiecutter_dict: dict[str, Any],
-) -> str:
+) -> Any:
     """Render the next variable to be displayed in the user prompt.
 
     Inside the prompting taken from the cookiecutter.json file, this renders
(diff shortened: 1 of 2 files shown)


jensens:
Ok, this solves a problem, but has some issues:
- Breaking Change - String values "true"/"false" will also be converted to booleans, which could break existing templates - this would result in a MAJOR version bump!
- Incomplete Tests - Only tests True, missing tests for False, edge cases, and strings
- Missing Type Safety - No isinstance() check before calling .lower()
- Return Type - Uses -> `Any` instead of the more precise -> `_Raw` type

TODO:
- Add `isinstance(rendered_value, str)` check for type safety
- Change return type from -> Any to -> _Raw
- Add comprehensive test coverage (False, strings, edge cases)
- You may want to move/add tests to` test_prompt.py`

PR #2167, opened 2025-09-15
Improved CLI, GUI and TUI interface

e3rd:
My proposal is to implement miniterface, a versatile CLI and dialog toolkit. This gives us various advantages.

# Automatic GUI and colourful helps

If left empty, there is an automatic wizzard. Note the file/dir-picker buttons.

```bash
$ cookiecutter
```

image

See nicer output and colour when invoking help.

```bash
$ cookiecutter  --help
cookiecutter --help
usage: cookiecutter [-h] [OPTIONS] [PATH] [[STR [STR ...]]]

╭─ positional arguments ─────────────────────────────────────────────────────╮
│ [PATH]                  Template to use (default: None)                    │
│ [[STR [STR ...]]]       Extra context parameters as KEY=VALUE pairs        │
│                         (default: )                                        │
╰────────────────────────────────────────────────────────────────────────────╯
╭─ options ──────────────────────────────────────────────────────────────────╮
│ -h, --help              show this help message and exit                    │
│ --version               show program's version number (Cookiecutter 2.6.0  │
│                         from                                               │
│                         /media/ram/project2/lib/python3.12/site-packages   │
│                         (Python 3.12.3 (main, Aug 14 2025, 17:47:21) [GCC  │
│                         13.3.0])) and exit                                 │
│ --no-input              Do not prompt for parameters and only use          │
│                         cookiecutter.json file content.                    │
│                         Defaults to deleting any cached resources and      │
│                         redownloading them.                                │
│                         Cannot be combined with the --replay flag.         │
│                         (default: False)                                   │
│ -c STR, --checkout STR  Branch, tag or commit to checkout after git clone  │
│                         (default: None)                                    │

...
```

# Automatic TUI

On machines without display (ex. through SSH), there is an automatic fallback for a mouse-clickable text interface.

```bash
$ cookiecutter 
```

image

# Transparent parser

Instead of the `click`, I've refactored the configuration to dataclasses. This improves readibility, allows the type to be defined only there on one place etc.

The original structure:

```python
...
@click.option(
    '--replay',
    is_flag=True,
    help='Do not prompt for parameters and only use information entered previously. '
    'Cannot be combined with the --no-input flag or with extra configuration passed.',
)
@click.option(
    '--replay-file',
    type=click.Path(),
    default=None,
    help='Use this file for replay instead of the default.',
)
...
```

The dataclass mentioned:

```python
@dataclass
class Env:
    ...
    replay: bool = False
    """Do not prompt for parameters and only use information entered previously.
    Cannot be combined with the --no-input flag or with extra configuration passed.
    """

    replay_file: Optional[File] = None
    """Use this file for replay instead of the default"""
   ...
```

# IDE benefits

Automatic type control, IDE suggestions and IDE is now able to show hints. See the value (`Literal`) and the programmer's comment (`Accept pre/post hooks`) which would be impossible with click.

image

As this is a dialog toolkit, we need no click or rich dialogs which will mean a shorter code.

image

# Dialogs

If you say yes, I'll shrink prompt.py to use mininiterface instead of rich for dialogs. This would mean much shorter code and versatility – dialogs will be GUI or mouse-clickable TUI, according to the user's endpoint.

image

TUI fallback:

image

# Conclusion

Are you interested in this PR? What do you think? Are there any changes I should catch up?

If you agree, I rewrite `prompt.py`.


jensens:
I like the idea of a GUI for cookiecutter, but this would better be an own project (like cookiecutter/cookiecutter-gui) which uses cookicutter (CLI) in order to have a clean CLI tool with a GUI on top. i think this is a common practice I have seen in other projects too. If it needs glues code here or changes to enable a GUI tool, I am open for contributions.

PR #2171, opened 2025-09-30
fix issue where empty list causes a crash due to index out of range

meganlkm:
Fixes the issue where sending an empty list in the options will raise an `IndexError` exception.

### DIFF
--- a/cookiecutter/prompt.py
+++ b/cookiecutter/prompt.py
@@ -275,6 +275,8 @@ def prompt_choice_for_config(
     """
     rendered_options = [render_variable(env, raw, cookiecutter_dict) for raw in options]
     if no_input:
+        if len(rendered_options) == 0:
+            return ""
         return rendered_options[0]
     return read_user_choice(key, rendered_options, prompts, prefix)
 
--- a/tests/test_prompt.py
+++ b/tests/test_prompt.py
@@ -520,6 +520,18 @@ class TestPromptChoiceForConfig:
         read_user_choice.assert_called_once_with('orientation', choices, None, '')
         assert expected_choice == actual_choice
 
+    def test_empty_list_returns_empty_string(self) -> None:
+        """Verify empty list returns empty string."""
+        context = {'project': 'foobar'}
+        actual_choice = prompt.prompt_choice_for_config(
+            cookiecutter_dict=context,
+            env=environment.StrictEnvironment(),
+            key='orientation',
+            options=[],
+            no_input=True,  # Suppress user input
+        )
+        assert actual_choice == ""
+
 
 class TestReadUserYesNo:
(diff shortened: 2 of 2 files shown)


kurtmckee:
@meganlkm I continue to think that an empty options list should manifest as a `ValueError` that explicitly says "the list of options is empty". This is the feedback that I previously provided in #2066, which was closed for lack of response and lack of tests.

@pydanny and @audreyfeldroy Would you take a gander at this behavioral change? Here's the summary:

When cookiecutter is told NOT TO ASK the user for input, but the list of choices to select from is empty:

* cookiecutter currently raises an `IndexError` as a side effect of accessing the empty list's first element.
* This PR modifies the behavior to return an empty string.
* My suggestion is that an empty list should raise a `ValueError` to indicate that the list of choices is empty. Something like this:

  ```python
      if not rendered_options:
          raise ValueError("The list of choices is empty")
      if no_input:
          return rendered_options[0]
      return read_user_choice(key, rendered_options, prompts, prefix)
  ```

The current behavior is incorrect, but I don't think the solution is to return a hard-coded empty string.

If returning an empty string seems like correct, new behavior to you, this PR is good for merging.

jensens:
sorry, i was about to suggest and well ... not enough coffee here

tests/test_prompt.py
```diff
@@ -520,6 +520,18 @@ def test_should_read_user_choice(self, mocker, choices, context) -> None:
...
+        """Verify empty list returns empty string."""
+        context = {'project': 'foobar'}
+        actual_choice = prompt.prompt_choice_for_config(
+            cookiecutter_dict=context,
+            env=environment.StrictEnvironment(),
+            key='orientation',
+            options=[],
+            no_input=True,  # Suppress user input
+        )
+        assert actual_choice == ""
```
jensens:
```suggestion
        with pytest.raises(ValueError):
            actual_choice = prompt.prompt_choice_for_config(
                cookiecutter_dict=context,
                env=environment.StrictEnvironment(),
                key='orientation',
                options=[],
                no_input=True,  # Suppress user input
            )
```

jensens (changes requested):

cookiecutter/prompt.py
```diff
@@ -275,6 +275,8 @@ def prompt_choice_for_config(
     """
     rendered_options = [render_variable(env, raw, cookiecutter_dict) for raw in options]
     if no_input:
+        if not rendered_options:
+            raise ValueError("The list of choices is empty")
```
jensens:
```suggestion
            msg = "The list of choices is empty"
            raise ValueError(msg)
```

jensens (changes requested):

PR #2172, opened 2025-10-05
The contributing guide was overly strict about the use of classes.

krishivs789:
While functions are still preferred, this change clarifies that classes are
  acceptable in specific cases, such as for custom Jinja2 extensions that
  require inheritance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,7 +72,7 @@ from cookiecutter.main import cookiecutter
 cookiecutter('cookiecutter-pypackage/')
 
 # Create project from the cookiecutter-pypackage.git repo template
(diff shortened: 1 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -136,7 +136,7 @@ Before you submit a pull request, check that it meets these guidelines:
 ### Coding Standards
 
 - PEP8
-- Functions over classes except in tests
+- Default to functions. Classes should only be used when there is a compelling reason, for example, when inheritance is required for custom Jinja2 extensions. Tests are another exception.
```
jensens:
```suggestion
- Functions over classes, except in tests or when classes provide a compelling advantage (e.g., inheritance for custom Jinja2 extensions)
```

PR #2176, opened 2025-11-13
fix: drop the end year in copyright

CodeWithEmad:
This value was hardcoded before.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -15,6 +15,7 @@
 
 import os
 import sys
+from datetime import datetime
 
 # If extensions (or modules to document with autodoc) are in another directory,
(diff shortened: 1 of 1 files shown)


CodeWithEmad:
Thanks for the feedback @jensens. 
I understand the point about copyright year updates not being legally required. However, my change wasn’t about legal necessity but about improving consistency and maintainability in the project. It could have been discussed rather than closing the PR outright.
I’m always happy to adjust the approach if there’s a preferred standard.

jensens:
the correct approach would be to keep the first release year and drop the end year.

CodeWithEmad:
Got it. Here you go https://github.com/cookiecutter/cookiecutter/pull/2179

PR #2179, opened 2025-11-16
fix: update copyright year in documentation

CodeWithEmad:
This will only keep the start year and drop the end year


audreyfeldroy:
This conflicts with main. The epub fields in docs/conf.py were updated in the 2.7.0 cycle to use the current author name and year range:

```python
epub_author = 'Audrey M. Roy Greenfeld'
epub_publisher = 'Audrey M. Roy Greenfeld'
epub_copyright = '2013-2026, Audrey M. Roy Greenfeld'
```

Merging this PR would revert those changes. Thanks for the contribution!

PR #2194, opened 2026-03-02
Update VERSION.txt

bollwyvl:
Thanks for `cookiecutter`, and congratulations on `2.7.0` :cookie: :heart: 

This PR updates the hard-coded `VERSION.txt` to the current version.

<details><summary>Another approach altogether...</summary> 

...would be to remove `VERSION.txt` altogether, and rely on `importlib.metadata.version` 

At the very simplest, `_get_version` could be replaced with something like the slightly awkward, but one-line:

```python
__version__ = f"{__import__('importlib.metadata').metadata.version('cookiecutter')}"
```

This would prevent this kind of PRs in the future, but would entail:
- an extra import
- reading a bigger file
- breaking workflows that try to use `cookiecutter` from a source checkout without an `install`

With some more jumping jacks, the first two downsides could be avoided by making it a lazily loaded module property, but that's maybe weirder.
</details>

References:
- noted on https://github.com/conda-forge/cookiecutter-feedstock/pull/35


bollwyvl:
Great, glad it worked out!

PR #2209, opened 2026-03-30
feat: add _force_render option to bypass binary detection for text files

### LINKED ISSUE #2205: [Bug]: Files starting with PACKAGE_NAME := ... misclassified as binary and skipped
## Summary
Cookiecutter incorrectly classifies certain plain text template files as binary when they start with:

```
PACKAGE_NAME := {{ cookiecutter.project_slug }}
```

As a result, the file is copied without Jinja rendering, leaving template variables unresolved.

---

## Expected Behavior
The file should be treated as text and rendered:

```
PACKAGE_NAME := example-package
```

---

## Actual Behavior
The file is treated as binary and copied as-is:

```
PACKAGE_NAME := {{ cookiecutter.project_slug }}
```

No warning or error is emitted.

---

## Root Cause
Cookiecutter uses the `binaryornot` library to determine whether a file is binary.

The file starts with:

```
PACKAGE_NAME := ...
```

The first bytes are:

```
PACK
```

`binaryornot` includes `b"PACK"` in its binary signature list, so:

```
chunk.startswith(b"PACK") → True
```

This results in a false positive binary classification, causing Cookiecutter to skip rendering entirely.

---

## Debug Evidence
```python
from binaryornot.check import is_binary
print(is_binary("foo_file"))  # True
```

```python
from binaryornot.helpers import _BINARY_SIGNATURES
chunk = b'PACKAGE_NAME := {{ cookiecutter.project_slug }}\n'
for sig in _BINARY_SIGNATURES:
    if chunk.startswith(sig):
        print(sig)
```

**Output:**
```
b'PACK'
```

## Workarounds
Any of the following avoids the issue:

- Add a blank line
```
PACKAGE_NAME := {{ cookiecutter.project_slug }}
```

- Add a comment
```
# Generated by cookiecutter
PACKAGE_NAME := {{ cookiecutter.project_slug }}
```

- Rename variable
```
PROJCT_NAME := {{ cookiecutter.project_slug }}
```
## 💡 Proposed Enhancement
Since binary detection is heuristic and can produce false positives, it would be useful to provide a way to override it.

### Proposal
Add a configuration option similar to `_copy_without_render`, for example:

```json
{
  "_force_render": [
    "Makefile",
    "foo_file",
    "*.mk"
  ]
}
```

### Suggested Behavior
1. If path matches `_copy_without_render` → copy without rendering  
2. If path matches `_force_render` → always render as text  
3. Otherwise → use binary detection

---

## Benefits
- Avoids fragile reliance on heuristics  
- Gives template authors explicit control  
- Solves real-world edge cases like this one  
- Maintains backward compatibility

---

## Impact
- Silent rendering failures  
- Hard to debug without `--verbose`  
- Affects common files (e.g., Makefile, config files)  
- Any file starting with `PACK...` may be impacted

---

## Conclusion
This is a false positive in binary detection leading to skipped rendering. A configurable override would make Cookiecutter more robust and predictable.

### Steps to reproduce

1. Create a minimal template:

```json
cookiecutter.json
{
  "project_slug": "example-package"
}
```

```
{{cookiecutter.project_slug}}/foo_file
PACKAGE_NAME := {{ cookiecutter.project_slug }}
```

2. Run:
```
cookiecutter --no-input /path/to/template
```

3. Inspect output:
```
cat example-package/foo_file
```

### Template

cookiecutter.json
```
{
  "project_slug": "example-package"
}
```

### Output / traceback

```shell

```

### Cookiecutter version

2.7.1

### Python version

3.12

### Operating system

Linux

### How did you install Cookiecutter?

uv

### Additional context

_No response_

juliosuas:
closes #2205.

hit this with a `Makefile` that starts with `PACKAGE_NAME := ...`. the first 4 bytes spell `PACK` which is a git packfile signature, so binaryornot calls it binary and cookiecutter copies the file verbatim — none of the jinja2 vars get rendered.

the easiest out is to let the template author say "render these as text no matter what binaryornot thinks". added `_force_render` in `cookiecutter.json`, same shape as the existing `_copy_without_render` (list of glob patterns):

```json
{
  "_force_render": ["Makefile", "*.mk", "GNUmakefile"]
}
```

files matching any of the patterns skip the `is_binary` check in `generate_file()` and go straight into jinja2 rendering. non-matching files behave exactly like before — no change to the default path.

implementation is a new `is_force_render_path()` helper (mirrors `is_copy_only_path`) plus one branch in `generate_file`. tests cover match/no-match and precedence over `_copy_without_render`.

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -56,6 +56,35 @@ def is_copy_only_path(path: str, context: dict[str, Any]) -> bool:
     return False
 
 
+def is_force_render_path(path: str, context: dict[str, Any]) -> bool:
+    """Check whether the given `path` should always be rendered as text.
+
+    Some tools (e.g. ``make``) produce files whose first bytes happen to
+    match a binary-file signature used by ``binaryornot``.  For example, a
+    ``Makefile`` that starts with ``PACKAGE_NAME := ...`` begins with the
+    bytes ``b'PACK'``, which ``binaryornot`` treats as a binary signature.
+
+    ``_force_render`` is a list of glob patterns in the cookiecutter context
+    that overrides the binary detection heuristic and forces those files to
+    be rendered as Jinja2 templates.
+
+    Returns True if `path` matches a pattern in the ``_force_render`` list,
+    otherwise False.
+
+    :param path: A file-system path referring to a file that should be
+        checked.
+    :param context: cookiecutter context.
+    """
+    try:
+        for force_render in context['cookiecutter']['_force_render']:
+            if fnmatch.fnmatch(path, force_render):
+                return True
+    except KeyError:
+        return False
+
+    return False
+
+
 def apply_overwrites_to_context(
     context: dict[str, Any],
     overwrite_context: dict[str, Any],
@@ -217,8 +246,12 @@ def generate_file(
     logger.debug('Created file at %s', outfile)
 
     # Just copy over binary files. Don't render.
+    # A file listed in ``_force_render`` bypasses the binary-detection
+    # heuristic — useful when a text file's first bytes match a binary
+    # signature (e.g. a Makefile starting with ``PACKAGE_NAME := ...``
+    # whose first bytes are ``b'PACK'``).
     logger.debug("Check %s to see if it's a binary", infile)
-    if is_binary(infile):
+    if not is_force_render_path(infile, context) and is_binary(infile):
         logger.debug('Copying binary %s to %s without rendering', infile, outfile)
         shutil.copyfile(infile, outfile)
         shutil.copymode(infile, outfile)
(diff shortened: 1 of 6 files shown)


juliosuas:
One design note: I chose to keep `_force_render` as a list of glob patterns (matching the shape of `_copy_without_render`) rather than a boolean flag, since it gives more fine-grained control. Users can target specific files (`["Makefile"]`) or use patterns (`["*.mk", "GNUmakefile"]`) rather than disabling binary detection globally. The helper function `is_force_render_path()` is a direct mirror of `is_copy_only_path()` for consistency.

juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/cookiecutter/cookiecutter/pull/2267

PR #2221, opened 2026-04-27
Add license file and valid SPDX identifier to project settings

ecederstrand:
Fixes license specifiers according to https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files

This allows automatic license inspection tools like pip-licenses to correctly categorize this package.


ecederstrand:
Ping @audreyfeldroy ?

PR #2237, opened 2026-06-14
Shell Completion for Templates and Template Variables

### LINKED ISSUE #1887: [Feature Request] tab completion for installed cookiecutters
A project can be created from an installed cookiecutter template by typing `$ cookiecutter <template name>`. It is not currently possible to type `$ cookiecutter <TAB><TAB>` and get a list of installed cookiecutters. That is, tab completion is not enabled for installed cookiecutters. I think this would be convenient for quick installation of new projects.

I took a quick look at the source code for cookiecutter and see that it uses click. Click has [documentation](https://click.palletsprojects.com/en/8.1.x/shell-completion/) for shell completion. I haven't used click before, but would be happy to have a go at implementing this.

neilramsay:
This adds shell completion for:
* `cookiecutter [template]`
* `cookiecutter [template] [extra_context]`

It uses the native [click shell completion](https://click.palletsprojects.com/en/stable/shell-completion/) with custom completers defined for the above.

> [!IMPORTANT]
> Please note that the click library has temporarily broken fish completion (see https://github.com/pallets/click/pull/3504), which will be fixed in their next release.

---

The default click shell completion shows this for options (for Fish shell):
image

---

The template completion shows templates installed in the `~/.cookiecutters` folder (by default, but overridable with the [`cookiecutters_dir`](https://cookiecutter.readthedocs.io/en/stable/advanced/user_config.html) user configuration)
image

---

The template "extra_context" (i.e. `cookiecutter.json` fields) completion shows the `__prompts__` [human readable prompts](https://cookiecutter.readthedocs.io/en/stable/advanced/human_readable_prompts.html) and field defaults.

For example, the `cookiecutter-pypackage` has these fields with defaults (but no human readable prompts)
image

The "extra_context" completion can be repeated, and removes previously provided values.
image

Another example template includes both the `__prompts__` fields, and a default value (Python Version = 3.14)
image

Fixes #1887

### DIFF
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -88,7 +88,7 @@ def list_installed_templates(
 
 
 def complete_templates(
-    ctx: Context, param: Argument, incomplete: str
+    ctx: Context, _param: Argument, _incomplete: str
 ) -> list[CompletionItem]:
     """Provide Completions for installed (locally cloned) templates."""
 
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -158,7 +158,7 @@ def complete_extra_context(
 
         completions.append(CompletionItem(value=f"{key}=", help=help_text))
 
-    return completions
+    return sorted(completions, key=lambda item: item.value)
 
 
 @click.command(context_settings={"help_option_names": ['-h', '--help']})
--- a/cookiecutter/cli.py
+++ b/cookiecutter/cli.py
@@ -6,6 +6,7 @@ import json
 import os
 import sys
 from collections import OrderedDict
+from pathlib import Path
 from typing import TYPE_CHECKING, Any
 
 from click.shell_completion import CompletionItem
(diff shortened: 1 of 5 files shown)


neilramsay:
Currently kept in draft as missing pytests.

I attempted to create some unit tests using the `cli_runner` but was unable to trigger the click completion process.
Their documentation notes:
> Completion is only available if a script is installed and invoked through an entry point, not through the python command. See [Packaging Entry Points](https://click.palletsprojects.com/en/stable/entry-points/). Once the executable is installed, calling it with a special environment variable will put Click in completion mode.

neilramsay:
> Currently kept in draft as missing pytests.
> 
> I attempted to create some unit tests using the `cli_runner` but was unable to trigger the click completion process. Their documentation notes:
> 
> > Completion is only available if a script is installed and invoked through an entry point, not through the python command. See [Packaging Entry Points](https://click.palletsprojects.com/en/stable/entry-points/). Once the executable is installed, calling it with a special environment variable will put Click in completion mode.

I've managed to resolve testing by using:
1. `monkeypatch` to update the test environment,
2. Calling the Click `shell_complete` function directly,
3. Collecting the standard out using `capsys`

neilramsay:
Hi @audreyfeldroy could you please have a look at this PR?

PR #2240, opened 2026-06-22
fix: collect all overwrite errors instead of stopping at first invalid entry

### LINKED ISSUE #2219: apply_overwrites_to_context silently drops remaining overrides after first invalid entry
When `default_context` from the user config (e.g. `~/.cookiecutterrc` or `$COOKIECUTTER_CONFIG`) contains a single override that fails type conversion, every subsequent override in the same `default_context` dict is silently discarded. The generated project ends up with a mix of user-set values and `cookiecutter.json` defaults, without any indication which overrides were lost.

## Affected overrides

- Boolean variables: override that cannot be parsed by `YesNoPrompt`.
- Choice variables: override that is not in the declared list.
- Multichoice variables: override not a subset of the declared list.

## Where to look

- `cookiecutter/generate.py` — `apply_overwrites_to_context`: the first invalid entry raises `ValueError` and exits the `for variable, overwrite in overwrite_context.items()` loop, so all remaining entries are never processed.
- `cookiecutter/generate.py` — `generate_context`: the `ValueError` is caught and downgraded to `warnings.warn(f"Invalid default received: {error}")`. The warning names only the failing variable, not the abandoned ones.

## Impact

Silent, partial config merge. The warning is easy to miss in CI/Docker logs. Downstream templates produce incomplete output files and there is no way for a template author to detect this from a pre-gen hook — by then the context has already been partially merged.

## Not affected

`extra_context` goes through the same function but without the surrounding `try/except`, so invalid overrides there raise hard. The silent path is `default_context` only.

## Expected behavior

Per-entry isolation: collect errors, continue processing the remaining overrides, then report all offending entries together (either as a hard error or a single consolidated warning that lists every dropped variable).

kaXianc2-gom:
## Summary

Fixes #2219: `apply_overwrites_to_context` now collects validation errors from **all** entries instead of raising on the first invalid one. Valid entries after an invalid one are no longer silently dropped.

## Problem

When `default_context` (from `~/.cookiecutterrc`) contains multiple overrides and one fails validation, every subsequent entry is silently discarded. For example:

```yaml
default_context:
  full_name: "valid"
  orientation: "invalid"    # ← raises, loop exits
  project_slug: "valid"     # ← silently dropped!
```

The generated project ends up with a mix of user-set and default values, with no indication which overrides were lost.

## Fix

- `apply_overwrites_to_context` accumulates errors in a list instead of raising immediately
- After the loop completes, all errors are raised as a single combined `ValueError`
- For nested dict overrides, errors are prefixed with the parent variable name
- `generate_context` still catches the ValueError and issues a single `UserWarning`

## AI Assistance Disclosure

This contribution was developed with AI assistance (Claude / Anthropic Claude Code).

## Verification Process

1. Forked → cloned to local sandbox (`~/sandbox/cookiecutter/`)
2. `uv sync --group dev` installed all dependencies
3. Ran existing test suite: 377 passed, 6 skipped (×3 for flaky detection)
4. Applied fix (29 lines changed in `generate.py`)
5. Added 5 targeted tests covering:
   - Valid entry after invalid entry is still applied
   - All errors are collected in a single message
   - Nested dict errors are prefixed with parent variable
   - No exception when all entries are valid
   - `generate_context` reports all invalid defaults
6. Full test suite: 382 passed, 6 skipped, 0 failures
7. `ruff format` + `ruff check`: all clean

## Cross-Validation

| Scenario | Before | After |
|----------|--------|-------|
| All entries valid | ✅ | ✅ |
| Single invalid entry | ⚠️ raises immediately | ✅ collected, reported |
| Invalid entry before valid entry | ❌ valid dropped | ✅ valid applied |
| Multiple invalid entries | ❌ only first reported | ✅ all reported |
| Nested dict errors | ❌ first only | ✅ all, with parent prefix |
| `extra_context` (hard failure) | ✅ | ✅ (all errors raised) |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cookiecutter/generate.py
+++ b/cookiecutter/generate.py
@@ -62,7 +62,14 @@ def apply_overwrites_to_context(
     *,
     in_dictionary_variable: bool = False,
 ) -> None:
-    """Modify the given context in place based on the overwrite_context."""
+    """Modify the given context in place based on the overwrite_context.
+
+    Collects validation errors from all entries instead of stopping at the
+    first invalid one. This ensures that later entries in the overwrite
+    dictionary are still processed even when an earlier one fails.
+    """
+    errors: list[str] = []
+
     for variable, overwrite in overwrite_context.items():
         if variable not in context:
             if not in_dictionary_variable:
@@ -82,11 +89,10 @@ def apply_overwrites_to_context(
                 if set(overwrite).issubset(set(context_value)):
                     context[variable] = overwrite
                 else:
-                    msg = (
+                    errors.append(
                         f"{overwrite} provided for multi-choice variable "
                         f"{variable}, but valid choices are {context_value}"
                     )
-                    raise ValueError(msg)
             else:
                 # We are dealing with a choice variable
                 if overwrite in context_value:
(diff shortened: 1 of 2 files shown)


kaXianc2-gom:
Closing - shifting focus. Thanks!

PR #2246, opened 2026-07-16
Fix pytest parametrize iterator deprecation warnings

felipe-f-rocha:
Fixes deprecation warnings from pytest regarding non-Collection iterables passed to parametrize.

Changes:
- Convert `context_data()` generator to list in `test_generate_context`
- Convert `enumerate()` iterator to list in `test_read_user_choice`

This prevents `PytestRemovedIn10Warning` with newer pytest versions.

### DIFF
--- a/tests/test_generate_context.py
+++ b/tests/test_generate_context.py
@@ -15,45 +15,47 @@ from cookiecutter.exceptions import ContextDecodingException
 from cookiecutter.prompt import YesNoPrompt
 
 
-def context_data() -> Iterator[tuple[dict[str, Any], dict[str, Any]]]:
+def context_data() -> list[tuple[dict[str, Any], dict[str, Any]]]:
     """Generate pytest parametrization variables for test.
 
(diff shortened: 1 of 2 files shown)


felipe-f-rocha:
@audreyfeldroy

audreyfeldroy:
Thanks @felipe-f-rocha! I'm thinking maybe we just make `context_data()` return a list and pass it directly to parametrize, any reason not to?

felipe-f-rocha:
> Thanks @felipe-f-rocha! I'm thinking maybe we just make `context_data()` return a list and pass it directly to parametrize, any reason not to?

That makes sense to me. I don't see a reason to keep context_data() as a generator, so I changed it to return a list directly and kept the parametrize call simple.

I also converted the enumerate() iterator in test_click_invocation to a list to address the same pytest deprecation warning.

Thanks for the suggestion!

felipe-f-rocha:
@audreyfeldroy

